PR #106, opened 2026-04-19
feat: add sb stop terminal command

eminwux:
## Summary
- New `sb stop terminal <name>` command (plus `--all`, `--force`, `--timeout`, `--kill-after`, `--ignore-not-found`) to shut down live terminals previously started with `sbsh -d`.
- Adds a `TerminalController.Stop` RPC (api + server + client + controller) that triggers `Close` asynchronously so the reply can flush before teardown; falls back to SIGTERM on the recorded PID if the control socket is unreachable.
- Closes gap #3 in docs/gap-analysis.md. Metadata is left in place for `sb prune terminals`.

## Test plan
- [ ] `go test ./cmd/... ./pkg/...`
- [ ] `sbsh -d -p <profile>` then `sb stop terminal <name>` — terminal exits gracefully, metadata visible in `sb get terminals --all` as Exited.
- [ ] `sb stop terminal <name> --force` — SIGKILL path.
- [ ] `sb stop terminal <name> --kill-after --timeout 2s` against a process that ignores SIGTERM — escalates to SIGKILL.
- [ ] `sb stop terminal <name> --ignore-not-found` — exits 0 when name missing.
- [ ] `sb stop terminal --all` — stops every active terminal, summary line shown.

Note: `internal/terminal/controller_test.go:Test_HandleEvent_EvCmdExited` hangs when run with `-count>1` or after clearing the test cache; reproduces on `main` with this branch stashed, so unrelated to this change.

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -118,6 +118,16 @@ var (
 	SB_DETACH_ID = DefineKV("SB_DETACH_ID", "sb/detach/id")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SB_DETACH_SOCKET = DefineKV("SB_DETACH_SOCKET", "sb/detach/socket")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_STOP_ALL = DefineKV("SB_STOP_ALL", "sb/stop/all")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_STOP_FORCE = DefineKV("SB_STOP_FORCE", "sb/stop/force")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_STOP_TIMEOUT = DefineKV("SB_STOP_TIMEOUT", "sb/stop/timeout")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_STOP_KILL_AFTER = DefineKV("SB_STOP_KILL_AFTER", "sb/stop/killAfter")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_STOP_IGNORE_NOT_FOUND = DefineKV("SB_STOP_IGNORE_NOT_FOUND", "sb/stop/ignoreNotFound")
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_CONFIG_FILE = DefineKV("SBSH_CONFIG_FILE", "sbsh/configFile")
--- a/cmd/sb/sb.go
+++ b/cmd/sb/sb.go
@@ -31,6 +31,7 @@ import (
 	"github.com/eminwux/sbsh/cmd/sb/detach"
 	"github.com/eminwux/sbsh/cmd/sb/get"
 	"github.com/eminwux/sbsh/cmd/sb/prune"
+	"github.com/eminwux/sbsh/cmd/sb/stop"
 	"github.com/eminwux/sbsh/cmd/sb/version"
 	"github.com/eminwux/sbsh/cmd/types"
 	"github.com/eminwux/sbsh/internal/errdefs"
@@ -117,6 +118,7 @@ func setupRootCmd(rootCmd *cobra.Command) error {
 	rootCmd.AddCommand(version.NewVersionCmd())
 	rootCmd.AddCommand(autocomplete.NewAutoCompleteCmd(rootCmd))
 	rootCmd.AddCommand(prune.NewPruneCmd())
+	rootCmd.AddCommand(stop.NewStopCmd())
 	rootCmd.AddCommand(detach.NewDetachCmd())
 	rootCmd.AddCommand(attach.NewAttachCmd())
 	rootCmd.AddCommand(get.NewGetCmd())
--- a/cmd/sb/stop/stop.go
+++ b/cmd/sb/stop/stop.go
@@ -14,22 +14,29 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
-package terminal
+package stop
 
 import (
-	"context"
-	"net"
-
-	"github.com/eminwux/sbsh/pkg/api"
+	"github.com/spf13/cobra"
 )
 
-type Client interface {
-	Ping(ctx context.Context, ping *api.PingMessage, pong *api.PingMessage) error
-	Resize(ctx context.Context, args *api.ResizeArgs) error
-	Detach(ctx context.Context, id *api.ID) error
-	Attach(ctx context.Context, clientID *api.ID, response any) (net.Conn, error)
-	Close() error
-	Metadata(ctx context.Context, metadata *api.TerminalDoc) error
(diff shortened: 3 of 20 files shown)


eminwux:
**PR #106 Review — `feat: add sb stop terminal command`**

Solid addition — the RPC + SIGTERM/SIGKILL fallback layering is the right shape, async `Close` from the RPC handler correctly lets the reply flush, and idempotency (`ESRCH` swallow, `--ignore-not-found`, pre-signal liveness recheck) is well handled. No security concerns: signalling is bounded by the existing per-user permission model on the control socket.

A few items:

- **Correctness — PID-reuse race** (`cmd/sb/stop/terminals.go:340-356`, `:291`): `processAlive` and `sendSignal` operate purely on the metadata-recorded PID. Once the original process exits, the kernel can recycle that PID; the `--kill-after` SIGKILL (and `--force` SIGKILL) then targets whatever unrelated process now owns it. Consider capturing process start time (`/proc/<pid>/stat` field 22) at metadata write and re-validating before signalling, or using `pidfd_open` (race-free on Linux). **Blocking** — this is a real foot-gun on long-lived hosts.
- **Correctness — effective timeout is up to 2×** (`cmd/sb/stop/terminals.go:279-296`): with `--kill-after`, after the SIGTERM wait expires the SIGKILL path waits another full `opts.timeout`. `--timeout 30s --kill-after` ⇒ ~60s wall time. Either share one deadline or use a short fixed post-SIGKILL grace (1–2s). **Blocking** (small fix).
- **Dead code** (`internal/errdefs/errdefs.go:80`): `ErrStopTimeout` is declared but never returned — wire it into the timeout failure paths (`terminals.go:262/285/296`) or drop it. Non-blocking nit since it's an unexported-style sentinel pattern, but per project convention worth cleaning.
- **Style** (`cmd/sb/stop/terminals.go:114,141-143`): `resolveStopOpts(cmd, args)` doesn't use `cmd` (literal `_ = cmd`). Drop the parameter.
- **Style** (`cmd/sb/stop/terminals.go:313-316`): `stopViaRPC` sets both `context.WithTimeout(ctx, stopRPCDialTimeout)` and `WithDialTimeout(stopRPCDialTimeout)` — redundant, harmless. Pick one.
- **Tests** (`cmd/sb/stop/terminals_test.go`): coverage is shallow — only flag resolution and `processAlive(0/-1)`. No coverage for the RPC graceful path, SIGTERM fallback, `--all` aggregation, or `--kill-after` escalation. The `ControllerTest.StopFunc` you already added in `controller_fake.go` makes the RPC path easy to exercise. Non-blocking but worth a follow-up. The acknowledged unrelated `Test_HandleEvent_EvCmdExited` hang deserves its own issue.
- **UX nit** (`cmd/sb/stop/terminals.go:60`): `Use: "terminal [name]"` while `--all` consumes zero args is mildly misleading; `terminal [name|--all]` reads cleaner.

LGTM once the PID-reuse race and double-timeout are addressed — the rest are non-blocking.

eminwux:
Thanks for the review — pushed [`8500b11`](https://github.com/eminwux/sbsh/pull/106/commits/8500b11ae8d08249448b813be46a1dc520bc1323) addressing everything:

**Blocking**
- **PID-reuse race** — added `internal/pidutil` reading `/proc/<pid>/stat` field 22 on Linux (no-op fallback on Darwin/FreeBSD, since the repo already build-tags those). The token is captured into `TerminalStatus.PidStart` / `ClientStatus.PidStart` at metadata write, then `pidutil.Match` gates `processIsOurs`, `sendSignal`, and `waitForExit` — a recycled PID is treated as "already exited" and never signalled. I went with the `/proc` token over `pidfd_open` because it's portable to existing on-disk metadata and a uint64 round-trips cleanly through the JSON doc; happy to revisit with `pidfd_open` as a follow-up if you'd prefer the kernel-enforced variant.
- **Effective timeout up to 2×** — fixed by capping the post-SIGKILL wait at a small fixed grace (`sigkillGrace = 2s`) instead of reusing `--timeout`. The graceful path keeps the full `--timeout`; total wall time is now `--timeout + 2s` worst case for `--kill-after`.

**Non-blocking**
- `errdefs.ErrStopTimeout` — wired into all three timeout failure messages in `cmd/sb/stop/terminals.go`.
- Dropped the unused `cmd` parameter from `resolveStopOpts`; updated the caller and test.
- Removed the redundant `context.WithTimeout` in `stopViaRPC`; only `WithDialTimeout` remains.
- Cobra `Use:` is now `terminal [name|--all]`.

**Tests**
- Added `internal/pidutil` unit tests (self-stable, non-positive PID, zero/non-zero token match).
- Added `processIsOurs` tests covering the live-but-mismatched-token rejection and the zero-token liveness fallback.
- Fixed an unrelated build break in `internal/client/clientrunner/io_test.go` — `mockTerminalClient` was missing the `Stop` method that landed when this PR introduced the Stop RPC. The deeper RPC-graceful-path / SIGTERM-fallback / `--all` aggregation / `--kill-after` escalation tests you suggested are still worth adding; I'll open a follow-up issue rather than balloon this PR. Same for the unrelated `Test_HandleEvent_EvCmdExited` hang.

`go build ./...`, `go vet ./...`, and `go test ./... -skip Test_HandleEvent_EvCmdExited` are green.

eminwux:
**LGTM** — reviewed 8500b11 against the prior feedback:

**Blocking items — resolved**
- **PID-reuse race** — `pidutil.StartTime` is captured at `CreateMetadata` for both the terminal runner (`internal/terminal/terminalrunner/metadata.go:44`) and the client runner (`internal/client/clientrunner/metadata.go:44`), persisted as `PidStart` on `TerminalStatus` / `ClientStatus`, and checked in `processIsOurs`, `sendSignal`, and `waitForExit`. The `/proc/<pid>/stat` parser in `internal/pidutil/pidutil_linux.go` correctly splits after the *last* `)` so a `comm` containing spaces or parens doesn't shift the field index. The non-Linux fallback `(0, nil)` paired with the `expected == 0 → true` branch in `Match` gives a sensible "trust the PID" fallback for pre-upgrade metadata and unsupported platforms.
- **Effective timeout up to 2×** — `sigkillGrace = 2 * time.Second` is used in both `--force` and `--kill-after` SIGKILL waits; total wall time is now `--timeout + 2s` worst case.

**Non-blocking — resolved**
- `errdefs.ErrStopTimeout` wired into all three timeout failure messages (`cmd/sb/stop/terminals.go:264, 288, 299`).
- `resolveStopOpts` no longer takes `cmd`; caller and tests updated.
- `stopViaRPC` now uses only `WithDialTimeout`.
- `Use: "terminal [name|--all]"`.
- Unit coverage added for `internal/pidutil` (self-stable / non-positive / zero-expected / self-match / different-token) and for `processIsOurs` (live-but-mismatched-token rejection, zero-token liveness fallback).

**Verification**
- `go build ./...` ✅
- `go vet ./...` ✅
- `go test ./... -skip Test_HandleEvent_EvCmdExited` ✅

**Follow-ups tracked** (not blocking this PR):
- #107 — deeper stop-command tests (RPC graceful / SIGTERM fallback / `--all` / `--kill-after` / PID-reuse rejection)
- #108 — `Test_HandleEvent_EvCmdExited` hang
- #109 — `pidfd_open` kernel-enforced variant as a future alternative to the `/proc` start-time token

Ship it.

Issue #25, opened 2025-11-10
Lifecycle - Add field to TerminalDoc-TerminalSpec-ExecStep to send raw bytes to terminal (alternative to script)


eminwux:
Extend ExecStep so callers can push raw byte payloads instead of only scripts, wiring the new field through lifecycle execution; expect updates in internal/terminal/terminalrunner/lifecycle.go, related spec structs, and command parsing.

PR #119, opened 2026-04-20
feat: add sb write/read and Write/Subscribe RPCs for programmatic terminal I/O

### LINKED ISSUE #117: API - Programmatic terminal I/O (`sb write` / `sb read` + Go API)
## Motivation

Today the only way to interact with a live terminal is `sb attach` (bidirectional, human-driven TTY). Capture is always-on, but read-only and file-based. Agents and scripts have no way to inject input or observe output programmatically. This blocks the primary use case of this project.

## New surface

| Command | Behavior |
|---|---|
| `sb write <name> "text"` | Write bytes to the terminal's PTY input. Caret notation interpreted (`^C`, `^D`, `^Z`, `^[`). |
| `sb write <name> --raw "bytes"` | Disable interpretation; send literal bytes. |
| `sb write <name> -` | Read payload from stdin (for piping). |
| `sb read <name>` | Print the full capture log and exit. |
| `sb read -f <name>` | Print the full capture log, then follow live output (like `tail -f`). |

No auto-appended newline on `sb write` — users submit with explicit `\n` or `^M`.

## Implementation sketch

- Two new RPC methods on the terminal control socket:
  - `Write(ctx, []byte) error` — pushes bytes down the same PTY input path attach uses.
  - `Subscribe(ctx) (stream, error)` — registers a new `io.Writer` on the runner's existing fan-out multiwriter; server streams bytes back to the client.
- `sb read -f` handoff order: **subscribe first, then replay the capture file, then switch to the live stream**. Subscribing before replay avoids gaps between the end of the file and the first live byte.
- Go API under `pkg/rpcclient/terminal` (or a thin public wrapper):

```go
Write(ctx context.Context, data []byte) error
Read(ctx context.Context) (io.ReadCloser, error) // replay + follow
```

## Design questions to pin down in the PR

1. **Caret notation scope.** `^A`..`^Z`, `^[`, `^\`, `^]`, `^^`, `^_`, `^?` are the unambiguous set (stty convention). Multi-byte escape sequences (arrow keys = `\e[A`) don't fit caret cleanly — either extend with `\xHH` hex escapes alongside, or require `--raw` for those. Recommend: caret + `\xHH` + literal characters; no shell-style `\n`/`\t` to avoid ambiguity with `^M`/`^I`.
2. **Slow-reader backpressure.** The subscriber RPC stream needs a bounded buffer; if full, either block the writer (risks stalling other attachers) or drop oldest and mark the stream lossy. Recommend: per-subscriber buffer, drop + close on overflow, caller gets a clear "lagged" error.
3. **Write concurrency.** Multiple writers (human attach + `sb write`) already interleave on the PTY; document as existing behavior, no ordering guarantee between clients.

## Scope / non-goals

- No `expect`/wait-for-pattern helpers. Users layer those on `Read`.
- No ANSI stripping or line framing. Raw bytes only.
- No permission/ACL model — if you can reach the socket, you can write.

## Supersedes

- #68 (`Client - sb attach --read-only or -r`) — `sb read` replaces it.

## Related

- #25 (raw bytes at lifecycle time) — different surface (declarative, in profile), same underlying PTY write path. Should share an internal helper.
- #49 (public API surface) — this fits naturally as part of that epic.

eminwux:
Closes #117.

## Summary
- New `Write` and `Subscribe` RPCs on the terminal control socket. `Write` pushes bytes into the PTY; `Subscribe` registers an output-only consumer on the existing `DynamicMultiWriter` fan-out and returns its end of a socketpair via SCM_RIGHTS (same pattern as `Attach`).
- New `sb write <name> [text|-]` CLI. Supports caret notation (`^@`..`^_`, `^?`), `\xHH` hex escapes, `--raw` for verbatim bytes, and `-` to read from stdin. Shell-style `\n`/`\t` are intentionally rejected (ambiguous with `^M`/`^I`).
- New `sb read <name>` CLI. Without `-f` it reads the capture file directly from disk (works even if the terminal has exited); with `-f` it subscribes, replays the capture, then streams live output. Subscribing before replay closes the cutover gap.
- Per-subscriber 1 MiB bounded ring: a slow/abandoned subscriber is disconnected with a `lagged` sentinel rather than stalling the PTY reader and every other attacher.
- Go API: `Client.Write(ctx, *WriteRequest)` and `Client.Subscribe(ctx, *SubscribeRequest, resp) (net.Conn, error)` added to `pkg/rpcclient/terminal`.

## Test plan
- [x] `go build ./...` + `go vet ./...` clean
- [x] `go test ./... -count=1` green (incl. unit tests for the caret parser and subscriber overflow)
- [x] `go test -tags=integration ./cmd/sb/get/...` green
- [x] `E2E_BIN_DIR=\$(pwd) go test ./e2e -count=1` green (new round-trip test: boot sbsh, `sb write`, `sb read`, plus `--raw` and caret-parse-error cases)
- [ ] Manual smoke: `sb write mybox -` < script.sh; `sb read -f mybox | tee out.log`

## Design notes
The three design questions called out in #117 are resolved in the code as follows:
1. **Caret scope**: stty-style caret + `\xHH`; no shell-style escapes (they collide with `^M`/`^I`).
2. **Slow-reader backpressure**: per-subscriber bounded ring; on overflow the subscriber is detached and the stream closes with a sentinel so the client can surface a "lagged" error.
3. **Write concurrency**: multiple writers (human attach + `sb write`) interleave on the PTY; documented as pre-existing behavior, no new locking beyond what `*os.File.Write` provides.

Issue #25 (profile-level raw PTY writes at lifecycle time) should later be migrated to call the shared `WritePTY` helper introduced here — left as a follow-up PR, not part of this one.

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -128,6 +128,10 @@ var (
 	SB_STOP_KILL_AFTER = DefineKV("SB_STOP_KILL_AFTER", "sb/stop/killAfter")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SB_STOP_IGNORE_NOT_FOUND = DefineKV("SB_STOP_IGNORE_NOT_FOUND", "sb/stop/ignoreNotFound")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_WRITE_RAW = DefineKV("SB_WRITE_RAW", "sb/write/raw")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_READ_FOLLOW = DefineKV("SB_READ_FOLLOW", "sb/read/follow")
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_CONFIG_FILE = DefineKV("SBSH_CONFIG_FILE", "sbsh/configFile")
--- a/cmd/sb/read/read.go
+++ b/cmd/sb/read/read.go
@@ -126,7 +126,7 @@ func runRead(
 
 	// Subscribe BEFORE reading the capture file so bytes written during the
 	// replay are captured in the live stream rather than lost at the cutover.
-	conn, err := rc.Subscribe(ctx, &api.SubscribeRequest{ClientID: api.ID(naming.RandomID())}, nil)
+	conn, err := rc.Subscribe(ctx, &api.SubscribeRequest{ClientID: api.ID(naming.RandomID())}, &api.Empty{})
 	if err != nil {
 		return fmt.Errorf("subscribe: %w", err)
 	}
@@ -162,19 +162,39 @@ func dumpCapture(path string, stdout io.Writer) error {
 	return nil
 }
 
-// laggedNotice mirrors the sentinel the server writes before disconnecting
-// a slow subscriber. Detecting it lets us surface a clear error code.
+// laggedNotice mirrors the bracketed core of the server sentinel (see
+// internal/terminal/terminalrunner/subscriber.go). The server wraps this
+// in \r\n framing; the needle here is the bare bracketed form so framing
+// can evolve on the server without breaking client detection.
 var laggedNotice = []byte("[sbsh: subscriber lagged, disconnecting]")
 
 func streamLive(conn net.Conn, stdout, stderr io.Writer) error {
 	buf := make([]byte, 4096)
+	// A single server-side Write delivers the sentinel atomically, but
+	// TCP/socket reads can still split it across chunk boundaries. Keep
+	// a rolling tail of (len(laggedNotice)-1) bytes from the previous
+	// chunk and scan tail+chunk so a split sentinel is still detected.
+	tailCap := len(laggedNotice) - 1
+	var tail []byte
 	laggedSeen := false
 	for {
 		n, err := conn.Read(buf)
 		if n > 0 {
 			chunk := buf[:n]
-			if !laggedSeen && bytes.Contains(chunk, laggedNotice) {
-				laggedSeen = true
+			if !laggedSeen {
+				scan := chunk
+				if len(tail) > 0 {
+					scan = append(append(make([]byte, 0, len(tail)+n), tail...), chunk...)
+				}
+				if bytes.Contains(scan, laggedNotice) {
+					laggedSeen = true
+				} else if tailCap > 0 {
+					if len(scan) > tailCap {
+						tail = append(tail[:0], scan[len(scan)-tailCap:]...)
+					} else {
+						tail = append(tail[:0], scan...)
+					}
+				}
 			}
 			if _, werr := stdout.Write(chunk); werr != nil {
 				return fmt.Errorf("write stdout: %w", werr)
--- a/cmd/sb/read/read.go
+++ b/cmd/sb/read/read.go
@@ -1,0 +1,194 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package read
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"net"
+	"os"
+	"os/signal"
+	"syscall"
+
+	"github.com/eminwux/sbsh/cmd/config"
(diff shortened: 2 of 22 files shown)


eminwux:
## Review

Scoped, well-tested PR. One correctness issue worth fixing before merge; the rest is minor.

### Correctness

**Lagged-notice delivery is effectively dead code** — `internal/terminal/terminalrunner/subscriber.go:77-83`.

On overflow, `Write` sets `closed`/`lagged`, broadcasts, then calls `s.conn.Close()`. When `Run` wakes up:

- If the buffer still has bytes → `conn.Write(chunk)` fails on the just-closed conn → error branch returns without writing `laggedNotice`.
- If the buffer is empty → takes the "buf empty + closed" branch and attempts `conn.Write(laggedNotice)`, but the conn is already closed, so the notice never reaches the peer.

Net effect: `sb read -f` sees a clean EOF on lag and `ErrSubscriberLagged` is never surfaced. `TestSubscriber_LaggedReaderIsDropped` passes because it only asserts the detach callback, not sentinel delivery.

Suggested fix: drop the `_ = s.conn.Close()` from the overflow path in `Write` and let `Run` write the sentinel then close. A write deadline in `Run` would keep a hung peer from stalling the drain goroutine. Worth a follow-up test that reads the sentinel from the client side.

### Minor

- **Cross-chunk sentinel detection** — `cmd/sb/read/read.go:196-199`. `bytes.Contains(chunk, laggedNotice)` misses a notice split across two `Read` returns. Low probability given a single server-side `conn.Write`, but scanning a small rolling tail would be safer.
- **Nil response on Subscribe** — `cmd/sb/read/read.go:150` passes `nil` for the response. The jsonrpc codec tolerates nil (`ReadResponseBody` no-ops), so this works, but it's load-bearing on a stdlib detail. Passing an explicit `&struct{ OK bool }{}` would make the contract visible.
- **Duplicate `laggedNotice` literals** — defined in `subscriber.go:33` and `read.go:188` with different whitespace wrapping. One is the server sentinel, the other is the client needle; a comment noting the asymmetry would help future readers.

### Strengths

- Caret parser is tight and well-covered (lowercase `^c`, `\\`, negative cases for `\n`/`\t`, dangling escapes).
- Fakes updated for both `ControllerTest` and `Test` runner — consistent with existing patterns.
- Subscribe-before-replay cutover is the right trade-off and documented clearly in the command help.
- Security surface is consistent with existing Attach; no new exposure.
- E2E test covers round-trip, `--raw`, and caret parse-error paths.

### Recommendation

Approve after fixing the lagged-notice delivery path and extending the subscriber test to assert the sentinel reaches the client.

eminwux:
Thanks for the review — all four points addressed in d4f693d.

**Lagged-notice delivery.** Dropped `s.conn.Close()` from the overflow path in `Write`; `Run` now owns conn lifecycle (`defer conn.Close()`) and emits `laggedNotice` before returning. Added `SetWriteDeadline(subscriberWriteTimeout)` before each `conn.Write` in `Run` so a hung peer cannot stall drain now that `Write` no longer force-closes. Timeout is a `var` (default 5s) so tests can shorten it.

**Sentinel-delivery test.** New `TestSubscriber_LaggedNoticeReachesClient` — forces an oversized chunk through `Write`, has a real reader draining `clientSide`, asserts `bytes.Contains(got, laggedNotice)` before EOF. Also updated `TestSubscriber_LaggedReaderIsDropped` with a 100ms write timeout so the hung-peer detach path still completes quickly; had to drop `t.Parallel` from both since they share the package-level timeout var.

**Cross-chunk detection.** `streamLive` now keeps a `len(laggedNotice)-1`-byte rolling tail from the prior chunk and scans `tail+chunk`, so a sentinel split across two `Read` returns is still caught.

**Nil Subscribe response.** Swapped `nil` for `&api.Empty{}` at the call site.

**Framing asymmetry.** Added comments in both `subscriber.go` (server side, `\r\n` framing for terminal rendering) and `read.go` (client needle is the bare bracketed form so framing can evolve without breaking detection).

Verification: `go build`, `go vet`, and the full unit suite green including the new test. E2E couldn't exec in my current sandbox (it produces `ar` archives instead of ELF binaries), so the existing `E2E_BIN_DIR=$(pwd) go test ./e2e` step from the PR test plan still needs a local run before merge.

eminwux:
Confirmed on d4f693d:

- `Write` no longer closes the conn; `Run` owns lifecycle via `defer conn.Close()`. `SetWriteDeadline(subscriberWriteTimeout)` bounds drain on both the chunk write and the sentinel write. Clean ownership split.
- `TestSubscriber_LaggedNoticeReachesClient` forces overflow with a real draining reader and asserts the sentinel arrives before EOF — would have caught the original bug.
- Rolling-tail scan in `streamLive` covers the split-across-Read case.
- `&api.Empty{}` at the Subscribe call site makes the response contract explicit.
- Framing-asymmetry comments in both `subscriber.go` and `read.go` close the loop on the two-literal question.

Verification in my env: `go build`, `go vet`, and the full unit suite green, including all three subscriber tests. The e2e suite needs a local run (my sandbox can't produce executable ELF binaries — same limitation you hit).

LGTM to merge once e2e passes locally.

eminwux:
Correction on my last comment — the e2e failure wasn't a sandbox issue, it was me mis-building the binary. `cmd/sb` and `cmd/sbsh` are library packages; the actual main is `./cmd` with `sb` as a hardlink to `sbsh`. After `go build -o sbsh ./cmd && ln -f sbsh sb`, `E2E_BIN_DIR=$(pwd) go test ./e2e -count=1` passes green, including `TestSb_WriteRead_Roundtrip` and `TestSb_Write_CaretParsingErrors`. PR is good to merge.

Issue #118, opened 2026-04-20
API - Consume sbsh as a Go library (driven by kukeon integration)


eminwux:
## Motivation

`kukeon` (`github.com/eminwux/kukeon`) needs to spawn and drive `sbsh` `Terminal` and `Client` processes from Go — not by shelling out to `sb`/`sbsh`. Today `pkg/api` and `pkg/rpcclient/{client,terminal}` expose types and RPC interfaces, but there is no documented path for an external module to spawn either process, connect to one that already exists, drive its full RPC surface, or redirect its on-disk state away from `~/.sbsh`.

This issue is the umbrella that captures consumer-side requirements; it depends on #49 (`pkg/spawn`) and #117 (write/read I/O) and surfaces the remaining gaps.

## Target consumer

`kukeon` will use `sbsh` to attach interactive sessions to managed containers. It needs to:

- `import "github.com/eminwux/sbsh/pkg/..."` from a separate module without touching `internal/`.
- Build a `TerminalSpec` / `ClientSpec` from a profile name or inline fields (no YAML round-trip).
- Spawn a `Terminal` in detached mode and get a handle (PID, run path, socket).
- Spawn a `Client` that attaches to an existing `Terminal` by ID / name.
- Reconnect from a different process and drive the full `TerminalController` / `ClientController` surface via `pkg/rpcclient`.
- Run the whole stack under a caller-supplied state root (e.g. `/opt/kukeon/sbsh/`) with **no** implicit fallback to `~/.sbsh`.
- Get typed errors (stable sentinels in `pkg/api` or re-exported from `internal/errdefs`).

## Scope

| Need | Current state | Gap |
|---|---|---|
| Spawn `Terminal` process | `cmd/sbsh/terminal` orchestrates it | `pkg/spawn.NewTerminal` (#49) |
| Drive `Terminal` via RPC | `pkg/api.TerminalController`, `pkg/rpcclient/terminal.Client` exist | Document full surface (`Ping`/`Resize`/`Attach`/`Detach`/`Metadata`/`State`/`Stop` + `Write`/`Read` from #117) as public contract |
| Spawn `Client` process | `internal/client` only | `pkg/spawn.NewClient` (#49) |
| Drive `Client` via RPC | `pkg/rpcclient/client.Client` has only `Detach`/`Close` | Audit `ClientController` needs (status, state observation, …) and grow the RPC surface |
[...]

PR #122, opened 2026-04-20
feat(discovery): promote terminal/client/profile discovery to pkg/discovery

eminwux:
## Summary

- Promote `ScanTerminals`, `FindTerminalByID`, `FindTerminalByName`, `ScanClients`, `FindClientByName`, `LoadProfilesFromPath`, `LoadProfilesFromReaderWithContext`, and `FindProfileByName` into a new `pkg/discovery` package so external modules (e.g. `kukeon`) can enumerate terminals, clients, and profiles under a caller-supplied state root without importing `internal/`.
- Export canonical ID/Name accessors (`TerminalID`, `TerminalName`, `ClientID`, `ClientName`) from `pkg/discovery` and have `internal/discovery` call those instead of keeping duplicated helpers.
- Keep `internal/discovery` as a thin proxy for the scan/find/load names so in-tree callers (`cmd/sb/*`, `cmd/config/autocomplete.go`, `internal/client/controller.go`, `internal/profile/profile.go`, e2e tests) compile unchanged.
- Printer, prune, and reconcile helpers (`ScanAndPrintTerminals`, `ScanAndPruneClients`, `ReconcileTerminals`, `PrintHuman`, `NoTerminalsString`, etc.) stay internal — only the scan/find/load surface moves.
- Use `slices.SortStableFunc` (Go 1.21+) in `ScanTerminals`/`ScanClients` so the doc-comment guarantee of stable output holds when `(Spec.ID, Name)` tuples collide.
- Add unit tests in `pkg/discovery` covering empty/missing/invalid-JSON scans, `FindByID`/`ByName` found and not-found paths, sort order, and multi-doc/empty-doc/not-found behavior for profiles.

Refs #118 (PR-C). Does not close the umbrella — PR-G will.

## Test plan

- [x] \`go build ./...\`
- [x] \`go vet ./...\`
- [x] \`go test -count=1 ./pkg/... ./cmd/sb/... ./internal/discovery/...\`
- [x] \`go test -count=1 -tags=integration ./cmd/sb/get/...\`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/discovery/clients.go
+++ b/internal/discovery/clients.go
@@ -90,8 +90,8 @@ func ScanAndPruneClients(ctx context.Context, logger *slog.Logger, runPath strin
 		"client",
 		func(s api.ClientDoc) bool { return s.Status.State == api.ClientExited },
 		PruneClient,
-		clientID,
-		clientName,
+		pkgdiscovery.ClientID,
+		pkgdiscovery.ClientName,
 	)
 	return err
 }
@@ -122,15 +122,6 @@ func PruneClient(logger *slog.Logger, metadata *api.ClientDoc) error {
 		logger.InfoContext(context.Background(), "PruneClient: client folder removed", "path", metadata.Status.ClientRunPath)
 	}
 	return err
-}
-
-func clientID(s api.ClientDoc) string {
-	// If your type uses Id instead of ID, change to: return s.Id
-	return string(s.Spec.ID)
-}
-
-func clientName(s api.ClientDoc) string {
-	return s.Metadata.Name
 }
 
 func clientLabels(s api.ClientDoc) map[string]string {
@@ -171,14 +162,14 @@ func printClients(w io.Writer, clients []api.ClientDoc, printAll, wide bool) err
 		}
 		if wide {
 			fmt.Fprintf(tw, "%s\t%s\t%s\t%s\n",
-				clientID(s),
-				clientName(s),
+				pkgdiscovery.ClientID(s),
+				pkgdiscovery.ClientName(s),
 				s.Status.State.String(),
 				joinLabels(clientLabels(s)),
 			)
 		} else {
 			fmt.Fprintf(tw, "%s\t%s\n",
-				clientName(s),
+				pkgdiscovery.ClientName(s),
 				s.Status.State.String(),
 			)
 		}
@@ -203,11 +194,11 @@ func FindAndPrintClientMetadata(
 	logger *slog.Logger,
 	runPath string,
 	w io.Writer,
-	terminalName string,
+	clientName string,
 	format string,
 ) error {
 	logger.DebugContext(ctx, "FindAndPrintClientMetadata: scanning clients", "runPath", runPath)
-	clients, err := FindClientByName(ctx, logger, runPath, terminalName)
+	clients, err := FindClientByName(ctx, logger, runPath, clientName)
 	if err != nil {
 		logger.ErrorContext(ctx, "FindAndPrintClientMetadata: failed to scan clients", "error", err)
 		return err
--- a/internal/discovery/clients.go
+++ b/internal/discovery/clients.go
@@ -22,11 +22,10 @@ import (
 	"io"
 	"log/slog"
 	"os"
-	"sort"
 	"text/tabwriter"
 
-	"github.com/eminwux/sbsh/internal/defaults"
 	"github.com/eminwux/sbsh/pkg/api"
+	pkgdiscovery "github.com/eminwux/sbsh/pkg/discovery"
 )
 
 // ScanAndPrintClients finds all metadata.json under runPath/clients/*,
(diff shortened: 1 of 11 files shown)


eminwux:
## PR #122 Review — feat(discovery): promote to pkg/discovery

### Overview
Pure refactor/move: promotes the *read* discovery surface (`ScanTerminals`, `FindTerminalByID/ByName`, `ScanClients`, `FindClientByName`, `LoadProfilesFromPath`, `LoadProfilesFromReaderWithContext`, `FindProfileByName`) from `internal/discovery` to a new `pkg/discovery` package so external modules (e.g. kukeon) can consume it without reaching into `internal/`. Existing in-tree callers keep working because `internal/discovery` retains thin pass-through wrappers with the same signatures. Printers, prune, and reconcile helpers stay internal as intended. Net: +432/−229, single commit. Build is green.

### Strengths
- Clean split: read/enumerate surface is public; side-effecting/presentational helpers (`PrintHuman`, `ScanAndPrune*`, `Reconcile*`) stay internal. This matches the stated contract.
- BC-friendly: `internal/discovery` wrappers preserve in-tree imports — no churn in `cmd/sb/*`, `cmd/config/autocomplete.go`, `internal/client/controller.go`, etc.
- Good package doc on `pkg/discovery/scan.go` explaining the caller-supplies-`runPath` design.
- The restricted surface (no Print/Prune/Reconcile) keeps the external API small and avoids leaking log-writer/tabwriter concerns.

### Concerns

- **No tests for the new public surface.** `pkg/discovery` ships with `[no test files]`. Since this is now the supported external contract, at minimum add table-driven tests for:
  - `ScanTerminals`/`ScanClients` on an empty `runPath`, a missing `runPath`, and a populated one with invalid JSON (to lock the current error-on-any-bad-file behavior).
  - `FindTerminalByID`/`ByName` not-found and found paths (sort stability when two docs share the same ID).
  - `LoadProfilesFromPath` / `FindProfileByName` with multi-doc YAML, empty docs (should be skipped with warn), and the not-found error message.
  The existing `e2e` and `cmd/sb/get` tests exercise the internal wrappers, but once kukeon is the real consumer, unit tests at the public boundary are what prevent silent breakage.

- **Duplicated helpers.** `terminalID`/`terminalName` and `clientID`/`clientName` now exist in both `pkg/discovery/{terminals,clients}.go` and `internal/discovery/{terminals,clients}.go` (the internal copies are still used by the printer/prune code). Low risk today — they just return `Spec.ID`/`Spec.Name`/`Metadata.Name` — but if the shape of `api.TerminalDoc`/`ClientDoc` changes, both must be updated in lockstep. Consider exporting them from `pkg/discovery` (`TerminalID(doc)`, etc.) and having the internal printers call those, or moving them to a shared helper file. Not blocking.

- **PR description is missing one moved function.** `LoadProfilesFromReaderWithContext` also moves (and is part of the public API surface), but the summary only names `LoadProfilesFromPath`. Worth updating so the audit trail is accurate.

- **`sort.Slice` is not stable.** Consecutive-scan stability is guaranteed only when `(Spec.ID, Name)` tuples are unique. That's almost always true here, but `slices.SortStableFunc` (Go 1.21+) is cheap insurance and matches the doc comment's "stable output" wording.

- **Minor naming nit.** In `internal/discovery/clients.go` the parameter is named `terminalName` in the `FindAndPrintClientMetadata` chain (pre-existing, but copy-pasted here). If you're touching the file anyway, rename to `clientName` for clarity.

### Risk / Correctness
- No behavior change in the moved functions — code was transplanted verbatim (error messages, log fields, sort comparator, YAML skip-empty-doc rule all preserved). Low regression risk for in-tree callers.
- No security or perf implications: same file I/O, same glob pattern, same JSON/YAML decoders.
- `pkg/discovery` imports `internal/defaults` for `TerminalsRunPath`/`ClientsRunPath`. That's fine for in-module use, but worth noting: if kukeon ever needs to *override* those subdirs it will have to thread them in as args — the current signature only takes `runPath` and hardcodes the subdir. Probably out of scope for this PR, but a likely PR-G follow-up.

### Verdict
Ship-worthy as a mechanical split, but I'd ask for unit tests in `pkg/discovery` before relying on this as an external contract, and bump `sort.Slice` → `slices.SortStableFunc` to match the comment. The helper duplication and PR-description omission are nits.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

eminwux:
**PR #122 Review — `feat(discovery): promote terminal/client/profile discovery to pkg/discovery`**

Clean, mechanical promotion of the scan/find/load surface into `pkg/discovery` with the printer/prune/reconcile helpers correctly held back in `internal/discovery`. Apache/SPDX headers on every new file, no secret-leak surface, thin-wrapper shim keeps all in-tree callers compiling. Unit tests in `pkg/discovery` cover the happy paths plus empty/missing/invalid-JSON scans, not-found on find-by-{ID,name}, sort order, and multi-doc/empty-doc profile loading. `go build`, `go vet`, `go test ./pkg/... ./cmd/sb/... ./internal/discovery/...`, and `go test -tags=integration ./cmd/sb/get/...` all green locally at `546dd52`.

- **Nit — duplicated comparator** (`pkg/discovery/scan.go:106`): `cmpString` re-implements the standard `cmp.Compare[string]` (`import \"cmp\"`, Go 1.21+). Dropping it would shave 10 lines and match the style of `slices.SortStableFunc` calls in the same package.
- **Nit — package doc placement** (`pkg/discovery/scan.go:15`): the package-level doc comment is attached to `scan.go`. Not wrong, but a dedicated `doc.go` is the more conventional home and makes the public-surface intent easier to find when browsing godoc.
- **Nit — not-found errors still string-only** (`pkg/discovery/scan.go:116`): `findMetadataBy` returns `errors.New(msg)`, inherited from the old internal helper. Now that this is a public API, a typed `ErrNotFound` (in `internal/errdefs/` or a package-local sentinel) would let external consumers like kukeon use `errors.Is` instead of substring matching. Fine to defer — flag it if/when the umbrella (#118) wants a consistent error vocabulary.

LGTM to merge as-is — all three items are optional polish, not blockers.

PR #121, opened 2026-04-20
feat(profile): require explicit RunPath in BuildTerminalSpec

eminwux:
## Summary

PR-A of the umbrella #118 (consume sbsh as a Go library, driven by `kukeon` integration). This PR closes the *parametrize state root* gap from the scope table by making `internal/profile.BuildTerminalSpec` refuse to silently default `RunPath` to `~/.sbsh`.

- `internal/profile.BuildTerminalSpec` now returns `errdefs.ErrRunPathRequired` when `RunPath` is empty, instead of falling back to `config.DefaultRunPath()`. Library consumers (e.g. `kukeon`) can now run the stack under an arbitrary `StateRoot` (e.g. `/opt/kukeon/sbsh`) without any risk of silent writes into the current user's home directory.
- CLI behavior is unchanged: `cmd/sbsh/sbsh.go:LoadConfig` and `cmd/sbsh/terminal/terminal.go:PreRunE` already pre-fill `RunPath` via `viper.SetDefault(config.DefaultRunPath())`, so `sb` / `sbsh` continue to default to `~/.sbsh` when the user doesn't pass `--run-path`.
- Also reorders the `ProfilesFile` derivation so it runs *after* the new `RunPath` check — previously it was constructed from an empty `RunPath` before the fallback ran (latent bug, now unreachable).

Refs #118. Does **not** `Closes #118` — the umbrella stays open until PR-G lands.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/profile/...` — new `TestBuildTerminalSpec_EmptyRunPath_ReturnsErrRunPathRequired` passes
- [x] `go test ./...` (non-e2e) — no regressions
- [x] `make sbsh-sb && make test-e2e` — full e2e suite passes (binaries exercise the CLI-side default path, confirming no regression)

## Out-of-scope (flagged for later sub-PRs)

- `cmd/sbsh/terminal/terminal.go:88` reads `SB_ROOT_RUN_PATH` rather than `SBSH_ROOT_RUN_PATH`. May or may not be intentional; untouched here.
- The duplicated `LoadConfig` defaulting pattern across `cmd/sb/sb.go`, `cmd/sbsh/sbsh.go`, and `cmd/sbsh/terminal/terminal.go` would benefit from a shared `ResolveRunPath` helper — CLI-layer refactor, out of scope for PR-A.

PR #123, opened 2026-04-20
feat(spawn): add pkg/spawn for library-driven Terminal/Client lifecycle

### LINKED ISSUE #49: API - Add process spawning API in pkg/spawn
Create a pkg/spawn package that exposes NewSupervisor and NewTerminal helpers encapsulating process setup, wiring through existing runners and tests.

### LINKED ISSUE #118: API - Consume sbsh as a Go library (driven by kukeon integration)

eminwux:
## Summary

PR-B of the umbrella #118 (consume sbsh as a Go library, driven by `kukeon` integration). Closes #49. Depends on PR-A (#121, merged).

- `pkg/spawn.NewTerminal` spawns `<BinaryPath> terminal -` with the `TerminalSpec` encoded as JSON on stdin and the child detached via `Setsid`, so library consumers (e.g. kukeon) can own a terminal process from Go without shelling out.
- `pkg/spawn.NewClient` spawns `<BinaryPath> attach` against an existing Terminal identified by ID or Name. `RunNewTerminal` mode is intentionally out of scope — compose `NewTerminal` + `NewClient`.
- `TerminalHandle` / `ClientHandle` expose the full lifecycle surface requested by the umbrella: `PID()`, `SocketPath()`, `WaitReady(ctx)`, `Close(ctx)`, `WaitClose(ctx)`. Terminal readiness is signalled via a `Ping` over the control socket; Client readiness via socket-file presence (until PR-E lands `ClientController.Ping`).
- `Close` performs a graceful three-step escalation: pre-signal hook (`Stop` RPC for Terminal, `Detach` RPC for Client) → SIGTERM → SIGKILL, with a configurable grace period.
- Inputs validate against stable sentinels: `ErrBinaryPathRequired`, `ErrSpecRequired`, `ErrDocRequired`, `ErrSocketPathRequired`, `ErrAttachTargetRequired`, `ErrUnsupportedClientMode`, `ErrProcessExited`, `ErrReadyTimeout`. spawn explicitly does not consult `$PATH` so callers can't cross a state-root boundary by accident.

Refs #118. Does **not** ` Closes #118` — the umbrella stays open until PR-G lands.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./pkg/spawn/...` — 14 subtests pass (validation, argv construction, process-exits-before-ready, SIGTERM/SIGKILL escalation)
- [x] `go test ./...` — no regressions (e2e suite passes after rebuilding `sb`/`sbsh`)
- [x] `golangci-lint run ./pkg/spawn/...` — 0 issues

## Out of scope (flagged for later sub-PRs)

- No CLI refactor to consume `pkg/spawn` internally — CLI flows (`cmd/sbsh/terminal`, `cmd/sb/attach`, `internal/client/clientrunner`) keep their current subprocess invocation. A follow-up could unify them, but PR-B stays additive.
- No `ClientController.Ping`/`Stop` yet — `ClientHandle.WaitReady` falls back to socket-file presence and `Close` uses `Detach` + signals. PR-E will grow the RPC surface; `ClientHandle` will pick up the new calls then.
- No `pkg/builder` helpers (PR-D) — callers still build `TerminalSpec` / `ClientDoc` directly. Once PR-D lands, the example in `docs/examples/library-consumer/` (PR-G) will use it in combination with `pkg/spawn`.

### DIFF
--- a/pkg/spawn/client.go
+++ b/pkg/spawn/client.go
@@ -1,0 +1,268 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package spawn
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"os/exec"
+	"time"
+
+	"github.com/eminwux/sbsh/pkg/api"
+	clientrpc "github.com/eminwux/sbsh/pkg/rpcclient/client"
+)
+
+// ClientOptions configures NewClient.
+type ClientOptions struct {
+	// BinaryPath is the absolute path to an sb binary that supports the
+	// "attach" subcommand. Required — spawn does not consult $PATH so
+	// callers can't accidentally cross a state-root boundary.
+	BinaryPath string
+
+	// ExtraArgs, if non-empty, are inserted before "attach" (useful
+	// for e.g. "--verbose" or overriding root-level flags).
+	ExtraArgs []string
+
+	// Env is the environment for the spawned process. nil = inherit
+	// the parent's os.Environ(). An explicit empty slice means "start
+	// with no environment variables."
+	Env []string
+
+	// Stdin/Stdout/Stderr are wired to the child. Nil streams are
+	// redirected to /dev/null. For interactive use the caller
+	// typically hands in a PTY pair; for headless RPC-driven use,
+	// leaving them nil is fine.
+	Stdin  io.Reader
+	Stdout io.Writer
+	Stderr io.Writer
+
+	// Logger is used for spawn-side diagnostic logging. nil = quiet.
+	Logger *slog.Logger
+
+	// ReadyTimeout caps how long ClientHandle.WaitReady waits for the
+	// control socket to appear. 0 (the default) defers to the ctx
+	// passed to WaitReady.
+	ReadyTimeout time.Duration
+
+	// ReadyPollInterval sets how frequently WaitReady checks for the
+	// control socket. 0 means use a sensible default.
+	ReadyPollInterval time.Duration
+
+	// StopGracePeriod caps each step of the Close escalation
+	// (Detach RPC, SIGTERM, SIGKILL). 0 uses the default.
+	StopGracePeriod time.Duration
+}
+
+// ClientHandle is a library-side handle on a spawned Client
+// subprocess. Consumers can observe lifecycle via WaitReady/WaitClose,
+// request shutdown via Close, and connect to SocketPath() for the
+// subset of ClientController RPCs currently exposed (Detach today;
+// Ping/State/Stop land in PR-E).
+type ClientHandle struct {
+	doc        *api.ClientDoc
+	socketPath string
+
+	opts ClientOptions
+	proc *process
+}
+
+// NewClient spawns `<opts.BinaryPath> attach` with flags derived from
+// doc, backed by a detached (Setsid) process. Only
+// ClientMode=AttachToTerminal is supported in this release; compose
+// NewTerminal + NewClient for the run-new-terminal flow.
+//
+// doc.Spec.SockerCtrl must be set to the desired control-socket path
+// (spawn does not invent one); doc.Spec.TerminalSpec must carry either
+// an ID or a Name identifying the terminal to attach to.
+func NewClient(ctx context.Context, doc *api.ClientDoc, opts ClientOptions) (*ClientHandle, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+	if err := validateClientInputs(doc, opts); err != nil {
+		return nil, err
+	}
+
+	args := buildClientAttachArgs(doc, opts)
+
+	// Deliberately using exec.Command (not CommandContext) — the
+	// child is detached (Setsid) and must outlive the caller's ctx.
+	//nolint:gosec,noctx // Caller controls BinaryPath and ExtraArgs; spawn
+	// explicitly does not consult $PATH.
+	cmd := exec.Command(opts.BinaryPath, args...)
+	if opts.Env != nil {
+		cmd.Env = opts.Env
+	} else {
+		cmd.Env = os.Environ()
+	}
+
+	opened, err := resolveStdio(cmd, opts.Stdin, opts.Stdout, opts.Stderr)
+	if err != nil {
+		return nil, fmt.Errorf("wire stdio: %w", err)
+	}
+	defer closeFiles(opened)
+
+	proc, err := startProcess(cmd, opts.Logger)
+	if err != nil {
+		return nil, fmt.Errorf("start client process: %w", err)
+	}
+
+	return &ClientHandle{
+		doc:        doc,
+		socketPath: doc.Spec.SockerCtrl,
+		opts:       opts,
+		proc:       proc,
+	}, nil
+}
+
+func validateClientInputs(doc *api.ClientDoc, opts ClientOptions) error {
+	if opts.BinaryPath == "" {
+		return ErrBinaryPathRequired
+	}
+	if doc == nil {
+		return ErrDocRequired
+	}
+	if doc.Spec.ClientMode != api.AttachToTerminal {
+		return fmt.Errorf("%w: only AttachToTerminal is supported", ErrUnsupportedClientMode)
+	}
+	if doc.Spec.SockerCtrl == "" {
+		return ErrSocketPathRequired
+	}
+	if doc.Spec.TerminalSpec == nil ||
+		(doc.Spec.TerminalSpec.ID == "" && doc.Spec.TerminalSpec.Name == "") {
+		return ErrAttachTargetRequired
+	}
+	return nil
+}
+
+// buildClientAttachArgs translates a ClientDoc into the CLI argv for
+// `sb [--run-path X] attach [--socket S] [--id I | <name>] [--disable-detach]`.
+// It is exported-for-test via exportedForTestBuildClientAttachArgs to
+// keep the API surface small.
+func buildClientAttachArgs(doc *api.ClientDoc, opts ClientOptions) []string {
+	const maxAttachArgs = 8 // --run-path X attach --socket S --id I --disable-detach
+	args := make([]string, 0, maxAttachArgs+len(opts.ExtraArgs))
+
+	if doc.Spec.RunPath != "" {
+		args = append(args, "--run-path", doc.Spec.RunPath)
+	}
+	args = append(args, opts.ExtraArgs...)
+	args = append(args, "attach")
+
+	if doc.Spec.SockerCtrl != "" {
+		args = append(args, "--socket", doc.Spec.SockerCtrl)
+	}
+	if !doc.Spec.DetachKeystroke {
+		args = append(args, "--disable-detach")
+	}
+
+	term := doc.Spec.TerminalSpec
+	switch {
+	case term.ID != "":
+		args = append(args, "--id", string(term.ID))
+	case term.Name != "":
+		args = append(args, term.Name)
+	}
+
+	return args
+}
+
+// PID returns the OS PID of the spawned client process.
+func (h *ClientHandle) PID() int { return h.proc.pid() }
+
+// SocketPath returns the client's control-socket path.
+func (h *ClientHandle) SocketPath() string { return h.socketPath }
+
+// Doc returns the ClientDoc used to spawn this handle. The returned
+// pointer is shared — callers must not mutate it.
+func (h *ClientHandle) Doc() *api.ClientDoc { return h.doc }
+
+// WaitReady blocks until the client's control socket file is visible
+// on disk. Until PR-E adds a Ping RPC on ClientController, socket
+// presence is the best readiness signal spawn can offer without
+// peeking at internal state.
+func (h *ClientHandle) WaitReady(ctx context.Context) error {
+	if h.proc.exited() {
+		return fmt.Errorf("%w: %w", ErrProcessExited, h.proc.takeExitErr())
+	}
+
+	readyCtx, cancel := h.readyContext(ctx)
+	defer cancel()
+
+	poll := h.opts.ReadyPollInterval
+	if poll <= 0 {
+		poll = defaultReadyPollInterval
+	}
+
+	for {
+		if h.proc.exited() {
+			return fmt.Errorf("%w: %w", ErrProcessExited, h.proc.takeExitErr())
+		}
+		if _, err := os.Stat(h.socketPath); err == nil {
+			return nil
+		}
+
+		select {
+		case <-readyCtx.Done():
+			if errors.Is(readyCtx.Err(), context.DeadlineExceeded) && ctx.Err() == nil {
+				return ErrReadyTimeout
+			}
+			return ctx.Err()
+		case <-h.proc.exitCh:
+			return fmt.Errorf("%w: %w", ErrProcessExited, h.proc.takeExitErr())
+		case <-time.After(poll):
+		}
+	}
+}
+
+func (h *ClientHandle) readyContext(ctx context.Context) (context.Context, context.CancelFunc) {
+	if h.opts.ReadyTimeout <= 0 {
+		return context.WithCancel(ctx)
+	}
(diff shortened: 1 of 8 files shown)


eminwux:
## Code Review

### Overview
Adds `pkg/spawn` (8 files, ~1316 LoC, zero deletions). Provides `NewTerminal` / `NewClient` plus `TerminalHandle` / `ClientHandle` with `PID/SocketPath/WaitReady/Close/WaitClose`, layered on a shared `process` reaper that handles `Setsid` detachment and a 3-step graceful shutdown (preSignal RPC → SIGTERM → SIGKILL). Cleanly scoped — additive, no CLI/internal refactor, sentinel-based error contract.

### Strengths
- **Tight surface, clear contracts.** Stable sentinel errors, scope explicitly bounded to `AttachToTerminal` for clients with a typed `ErrUnsupportedClientMode`.
- **`process` abstraction is well factored.** The exit-once reaper + `exitCh`/`exitMu` split, `waitWithTimeout`'s tri-state return contract, and the documented `nil`/`ctx.Err()`/`DeadlineExceeded` distinction are clean.
- **Refusing `\$PATH` lookup is the right call** for state-root isolation, and the rationale is documented inline and on `ErrBinaryPathRequired`.
- **Stdio defaults to `/dev/null`**, with files closed after `Start()` — avoids the "child inherits TTY" footgun and the "leaked /dev/null fd" footgun simultaneously.
- License headers, doc comments, `t.Parallel()`, table-driven validation tests — all consistent with the rest of the repo.

### Issues / Suggestions

**1. `fmt.Errorf(\"%w: %w\", ErrProcessExited, h.proc.takeExitErr())` formats badly when exit was clean.**
`client.go:209,222,235` and `terminal.go:1137,1152,1175`. If the child exits 0 (e.g. `/bin/true`), `cmd.Wait()` returns `nil`, so the message renders as `\"spawned process exited before ready: %!w(<nil>)\"`. `errors.Is(err, ErrProcessExited)` still works, but the user-facing string is ugly. Consider:
\`\`\`go
if exitErr := h.proc.takeExitErr(); exitErr != nil {
    return fmt.Errorf(\"%w: %w\", ErrProcessExited, exitErr)
}
return ErrProcessExited
\`\`\`

**2. Missing `//go:build unix` (or `linux`) build tag.**
The package hard-depends on `syscall.SysProcAttr.Setsid`, `/bin/true`, `/bin/sleep`, `/dev/null`. Building on Windows will fail with cryptic errors. A `//go:build unix` constraint on `process.go` (and matching tests) would surface intent and stop accidental Windows builds. The umbrella issue should clarify whether non-Unix targets are ever in scope.

**3. `--socket` flag semantics on `sb attach` are conflated upstream — verify this PR's mapping is correct.**
`cmd/sb/attach/attach.go:161` registers `--socket` with help text \"Optional socket file for the **terminal**\", but the runtime at `cmd/sb/attach/attach.go:210,222-224,252` uses the same value as both `ClientSpec.SockerCtrl` and `TerminalSpec.SocketFile` — clearly buggy in the CLI. `pkg/spawn/client.go:175-177` passes `doc.Spec.SockerCtrl` as `--socket`, which matches the *first* usage but means the CLI will then mis-set `TerminalSpec.SocketFile` to the client socket (since `socketFileFlag != \"\"` short-circuits the terminal-socket derivation). End result: `sb attach` likely won't connect to the actual terminal when invoked via spawn unless `--socket` is left blank and the terminal is found by ID/name through the default `<runPath>/terminals/<id>/socket` derivation. This is a pre-existing CLI bug, not introduced here, but PR-B silently inherits it. Either: (a) fix the CLI in this PR / a sibling, (b) explicitly *don't* emit `--socket` from spawn and rely on the default derivation, or (c) document the constraint loudly.

**4. `if doc.Spec.SockerCtrl != \"\"` check is dead code.**
`client.go:175`. `validateClientInputs` already enforces non-empty `SockerCtrl` via `ErrSocketPathRequired`. Drop the `if` (and the equivalent dead-code comment in `maxAttachArgs`).

**5. ExtraArgs ordering vs `--run-path` collision.**
`client.go:169-172` emits `--run-path` then appends `ExtraArgs` then `\"attach\"`. If a caller puts `--run-path` in `ExtraArgs`, cobra will error on duplicate flag. Either document that explicitly or normalize. Low priority — caller-controlled.

**6. `WaitReady` for `ClientHandle` is socket-presence only.**
The PR description acknowledges this defers to PR-E for `ClientController.Ping`. Until then, \"ready\" can fire before the client RPC server has bound — any caller that immediately tries `Detach()` may race. Worth a doc note on `WaitReady` itself, not just on the package overview.

**7. Test coverage gap: argv when both `ID` and `Name` are set.**
`buildClientAttachArgs`'s `switch` prefers ID. Existing tests cover ID-only and Name-only; an explicit test asserting ID wins (matching the CLI's mutual-exclusion behavior) would lock down the contract.

**8. `defaultGracePeriod = 2 * time.Second` total budget is short.**
A heavy terminal (PTY + supervisor + multiple subscribers) may need longer for a clean Stop-RPC unwind. `2s` × 3 escalation steps = 6s wall-clock max. Probably fine for most use; flagging because it's caller-tunable but the default is buried as an unexported const, not documented on `StopGracePeriod`.

### Risk Assessment
- **Correctness:** Mostly sound. Issue #3 is the only one with a real \"wrong-behavior-in-prod\" risk; the rest are polish.
- **Performance:** Polling at 100ms is fine for spawn workloads; no hot paths.
- **Security:** Argv construction is safe (no shell, no `\$PATH`). `gosec`/`noctx` suppressions are justified and commented. Good.
- **Test coverage:** Good for argv + validation + escalation. Real terminal/client integration is rightly out of scope (e2e suite).

### Recommendation
Approve with #1, #3, #4 addressed (or #3 explicitly tracked as a follow-up). #2, #5–#8 are nits.

eminwux:
Pushed `eea4c1d` addressing the review.

**Fixed**
- **#1** nil-wrap: added `wrapProcessExited()` so a clean exit renders as bare `ErrProcessExited` instead of `"%!w(<nil>)"`. Applied to both `TerminalHandle.WaitReady` and `ClientHandle.WaitReady` (all three exit-check sites each). New test: `TestWrapProcessExited_NilExitError`.
- **#2** build tag: added `//go:build unix` to all eight `pkg/spawn` files. On non-Unix the package is simply absent, which surfaces intent without leaking cryptic `undefined: startProcess` errors from partial compilation.
- **#3** `--socket` conflation: added a block comment on `buildClientAttachArgs` explaining (a) that `sb attach`'s `--socket` flag is documented as "for the terminal" but runtime-wires it as the client socket, (b) that spawn relies on this actual behavior so `ClientHandle.SocketPath()` stays predictable, and (c) that the stale `TerminalSpec.SocketFile` set by the CLI is harmlessly overridden from metadata.json in `internal/client.createAttachTerminal`. CLI cleanup (rename/split the flag, fix the help text) left as a pre-existing bug outside PR-B scope.
- **#4** dead code: dropped the `if doc.Spec.SockerCtrl != ""` guard; `--socket` is now always emitted, since `validateClientInputs` already enforces the invariant via `ErrSocketPathRequired`.
- **#6** `ClientHandle.WaitReady` race: added a per-method caveat noting that socket presence fires before the client has finished wiring the attach, and that callers should retry or back off around any immediate `Detach()` until PR-E adds `ClientController.Ping`.
- **#7** argv coverage gap: added `TestBuildClientAttachArgs_IDWinsOverName` locking in the ID-over-Name tie-break with a reference to the CLI's mutual-exclusion check at `cmd/sb/attach/attach.go:72-83`.
- **#8** `defaultGracePeriod` visibility: documented the 2s-per-step, ~6s-wall-clock budget on `TerminalOptions.StopGracePeriod` and `ClientOptions.StopGracePeriod`, with a pointer to raise it for heavy terminals.

**Skipped (by design)**
- **#5** `ExtraArgs` vs `--run-path` collision: caller-controlled; duplicate-flag errors from cobra surface clearly. Not worth normalizing spawn-side.

**Verified** `go build ./...`, `go vet ./...`, `go test ./pkg/spawn/...` (now 16 subtests), `golangci-lint run ./pkg/spawn/...` — all clean.

eminwux:
**PR #123 Review — `feat(spawn): add pkg/spawn for library-driven Terminal/Client lifecycle`**

Solid, well-scoped PR. The API surface is clean, the three-step Close escalation is implemented correctly, and the `wrapProcessExited` helper avoids the `%!w(<nil>)` trap nicely. Build/vet/`go test ./pkg/spawn/...` all clean; all eight sentinels are referenced; SPDX + Apache headers and `//go:build unix` consistently applied. No sensitive-info leaks in error messages. The pre-signal → SIGTERM → SIGKILL ladder in `gracefulShutdown` correctly distinguishes `DeadlineExceeded` (local step) from `ctx.Err()` (caller canceled), and the reaper goroutine sets `exitErr` under the mutex before closing `exitCh`, so the `exited()` → `takeExitErr()` sequence is race-safe.

- **Comment precision** (`pkg/spawn/client.go:174-175`): "If --id and --name are both set the CLI rejects the combination" — actually `cmd/sb/attach/attach.go:72-83` only rejects *positional + flag*, not both flags together (the `run()` path at lines 215-219 happily sets both `TerminalSpec.ID` and `TerminalSpec.Name`). The deterministic ID-wins rule in `buildClientAttachArgs` is still correct because spawn emits `--id X` *or* positional name, never both — but the "CLI rejects the combination" premise is slightly off. Consider softening to "avoids the ambiguous input the CLI doesn't handle cleanly".
- **Redundant stdin reassignment** (`pkg/spawn/terminal.go:1147,1156`): `cmd.Stdin = bytes.NewReader(specBytes)` is set, then passed as the stdin arg to `resolveStdio`, which reassigns the same pointer back to `cmd.Stdin`. Works, but reads oddly. Either skip the stdin arg in the call (pass nil and let resolveStdio leave it alone — needs a small tweak there) or add a one-line comment that the round-trip is intentional.
- **Validation-order UX** (`pkg/spawn/client.go:144-162`): because `api.RunNewTerminal` is iota=0, a zero-valued `ClientSpec` trips `ErrUnsupportedClientMode` before any of the more specific "missing socket / attach target" errors. Not wrong, just a slightly less useful first error for an accidentally-empty doc — consider checking `SockerCtrl` / `TerminalSpec` first if you want tighter diagnostics.
- **Test-plan nit**: the unrelated e2e failure on `main` (`sb write` missing from the built binary — from commit 931746c) is noise, not a regression from this PR. You're clearly aware (the test plan says "after rebuilding sb/sbsh"), just noting it for the record.

LGTM to merge as-is — the items above are all nits, not blockers. Nice foundation for PR-D/PR-E to build on.

PR #125, opened 2026-04-21
fix: expand env vars in terminal cwd

eminwux:
## Summary
- `TerminalSpec.Cwd` (from `shell.cwd` in a profile) previously only expanded a leading `~`/`~/`, so a path like `$HOME/projects` was passed verbatim to `exec.Cmd.Dir`, the directory lookup failed, and the terminal never started in the configured working directory.
- Apply `os.ExpandEnv` before the tilde check so common forms (`$HOME/foo`, `${HOME}/foo`, custom inherited env vars) resolve against the parent process environment.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/terminal/terminalrunner/...`
- [x] `go test ./...`
- [x] `make sbsh-sb` and `file ./sbsh` is ELF executable
- [x] Manual repro: profile with `cwd: "$HOME/cwd-target-dir"` now starts the shell in the expanded directory

Closes #81

### DIFF
--- a/internal/terminal/terminalrunner/terminal.go
+++ b/internal/terminal/terminalrunner/terminal.go
@@ -104,9 +104,12 @@ func (sr *Exec) prepareTerminalCommand() error {
 	return nil
 }
 
-// expandCwd resolves a leading "~" or "~/" against the user's home directory.
-// Other shell expansions (env vars, globs) are not performed.
+// expandCwd resolves environment variables (e.g. "$HOME", "${HOME}") and a
+// leading "~" or "~/" against the user's home directory. Glob patterns are
+// not expanded. Env var lookup uses the parent process environment, so vars
+// defined only in the profile's shell.env are not visible here.
 func expandCwd(p string) (string, error) {
+	p = os.ExpandEnv(p)
 	if p == "~" || strings.HasPrefix(p, "~/") {
 		home, err := os.UserHomeDir()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #125 Review — `fix: expand env vars in terminal cwd`**

Clean, surgical fix: one `os.ExpandEnv` call in the right place, updated doc comment that accurately describes the new semantics (including the parent-env limitation), and a thorough table-driven test that pins down the happy paths plus the unset-var behavior. No sensitive data in new log/error strings, no dead symbols, license header matches the repo convention, and `Closes #81` is appropriate for the reported issue.

- **Ordering note** (`terminal.go:109`): running `os.ExpandEnv` before the tilde check means a profile like `cwd: "$MY_PATH"` where `MY_PATH=~/projects` will expand env first → `~/projects`, then tilde → `$HOME/projects`. Probably intended, but worth a one-line test case (`env-then-tilde`) to lock in the composition so a future refactor doesn't flip the order.
- **Silent empty expansion** (`terminal_test.go:44`): the `env-unset` case documents that `$TYPO/sub` quietly becomes `/sub`. This matches shell semantics and is fine, but if `exec.Cmd.Dir` later fails with `no such file or directory`, the user won't immediately see it was a typo'd env var. Not something to fix here; a future "cwd expansion ended up empty at the leading component → log a warning" is a reasonable follow-up if this ever bites.

LGTM — merging.

PR #130, opened 2026-04-22
fix: surface terminal exit errors in client

### LINKED ISSUE #43: Client - If `sbsh terminal` fails client must print and log error
Ensure client surfaces terminal startup errors to stdout/stderr and logs, adjusting internal/client logging paths and CLI error handling.

eminwux:
## Summary
- Client previously swallowed errors when the `sbsh terminal` subprocess failed to start or exited abnormally — user saw nothing and exit code was 0.
- Controller now distinguishes a clean subprocess exit (exit status 0) from a failure, so normal shell exits stay silent while real failures propagate.
- `runClient` returns the wrapped error instead of `nil`, letting cobra print it to stderr and set a non-zero exit code.

## Test plan
- [ ] `make test` (includes e2e)
- [ ] `./sbsh --terminal-command /bin/nonexistent-xyz` prints error and exits 1
- [ ] Normal interactive session (exit 0 from shell) stays silent and exits 0

Closes #43

### DIFF
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -422,11 +422,7 @@ func runClient(
 				err = fmt.Errorf("%w: %w: %w", err, errdefs.ErrWaitOnClose, errC)
 			}
 			logger.ErrorContext(ctx, "client controller exited with error", "error", err)
-			logger.DebugContext(ctx, "controller exited after error")
-
-			// return err
-			// return nothing to avoid polluting the terminal with errors
-			return nil
+			return err
 		}
 	}
 	return nil
--- a/cmd/sbsh/sbsh_test.go
+++ b/cmd/sbsh/sbsh_test.go
@@ -292,10 +292,14 @@ func TestRunTerminal_ErrChildExit(t *testing.T) {
 	// Wait for Run() to complete and error to be handled
 	select {
 	case err := <-done:
-		// runClient returns nil when errCh receives error (line 412)
-		// The error is logged but not returned to avoid polluting terminal
-		if err != nil {
-			t.Fatalf("expected nil (error logged but not returned); got: '%v'", err)
+		if err == nil {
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #130 Review — `fix: surface terminal exit errors in client`**

Small, precise fix. Three coordinated changes correctly flip the client from "silently swallow all terminal failures" to "silent on clean exit, propagate on failure": `lifecycle.go` now only builds an `eventErr` when `cmd.Wait()` returned an error (previously an unconditional `fmt.Errorf` made `eventErr` always non-nil, which is the root of the bug); `controller.go` guards the `closeReqCh` receive so a nil close reason returns nil instead of wrapping nil with `ErrCloseReq`; `runClient` returns the wrapped error instead of swallowing it. `%w` is preserved throughout, so `errors.Is(err, errdefs.ErrChildExit)` etc. still hold end-to-end. No sensitive data in the new error message — only the internal terminal ID. Test updates correctly assert the full wrap chain including `ErrWaitOnClose`, and the existing `Test_EventCmdExited` already exercises the clean-exit → nil path through `closeReqCh`.

A couple of small notes, non-blocking:

- **Test plan** (PR description): all three checkboxes are unchecked. CI covers `make test`, and the new unit tests cover the error-propagation logic, but the manual smoke tests (`/bin/nonexistent-xyz` → exit 1, normal shell exit → silent exit 0) are the scenarios from the bug report — worth actually running once and checking the boxes so the fix is confirmed end-to-end through cobra's exit-code path, not just through `runClient`'s return value.
- **Nit** (`cmd/sbsh/sbsh.go:425`): the stale `// return err` / `// return nothing to avoid polluting the terminal with errors` comments got removed along with the behavior change — good. The `logger.DebugContext(ctx, "controller exited after error")` removal right above is fine too; the `ErrorContext` line just before already logs the same thing with the error.

LGTM.

Issue #72, opened 2025-11-10
API - Client Control API (start/stop/list/attach/tail)


eminwux:
Define a Client Control API exposing start/stop/list/attach/tail operations, touching pkg/api (ClientController) and rpcclient implementations.

PR #136, opened 2026-04-23
feat: grow ClientController with Ping/Metadata/State/Stop RPCs

### LINKED ISSUE #132: API - Grow ClientController RPC surface (Ping/Metadata/State/Stop)
## Context

Sub-PR **E** of umbrella #118 (consume sbsh as a Go library, driven by `kukeon` integration). No dependency on other sub-PRs.

`pkg/rpcclient/client.Client` currently only exposes `Detach` / `Close`. External consumers need parity with `TerminalController` so a library caller can observe and control a live `Client` process the same way they can a `Terminal`.

PR-B's `ClientHandle.WaitReady` also falls back to socket-file presence because `ClientController.Ping` does not exist yet — this closes that gap.

## Proposal

Grow `ClientController` (server + API contract) with:

- `Ping()` — readiness probe.
- `Metadata() (*api.ClientDoc, error)` — full doc snapshot.
- `State() (*api.ClientStatusMode, error)` — current lifecycle state.
- `Stop(opts StopOptions) error` — graceful shutdown mirroring `TerminalController.Stop` semantics.

Mirror each on `pkg/rpcclient/client.Client`. Update `pkg/spawn.ClientHandle.WaitReady` to prefer `Ping` over socket-file presence.

## Acceptance criteria

- [ ] `pkg/api.ClientController` interface extended with `Ping` / `Metadata` / `State` / `Stop`.
- [ ] Server-side handlers implemented in `internal/client`.
- [ ] `pkg/rpcclient/client.Client` exposes matching methods.
- [ ] `pkg/spawn.ClientHandle.WaitReady` uses `Ping` (remove socket-presence fallback once available).
- [ ] Unit + integration tests covering each RPC against a live `Client` process.
- [ ] `go build ./...` + `go vet ./...` + `go test ./...` clean.
- [ ] PR body references #118 as PR-E but does not `Closes #118`.

## Notes

Refs umbrella: #118. Related: #72 (broader Client Control API).

eminwux:
## Summary
- Brings `ClientController` to parity with `TerminalController` so library consumers can observe and control a live `Client` process the same way they can a `Terminal` (umbrella #118, sub-PR E).
- `pkg/spawn.ClientHandle.WaitReady` now uses the new `Ping` RPC instead of polling for socket-file presence; `Close` uses the new `Stop` RPC instead of `Detach`.
- Mirrors the existing terminal-side patterns end-to-end: `pkg/api.ClientController` interface + method-name constants, server wrapper in `internal/client/clientrpc`, controller methods in `internal/client`, and matching client methods in `pkg/rpcclient/client`.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/client/... ./pkg/rpcclient/... ./pkg/api/... ./pkg/spawn/...`
- [x] `make sbsh-sb` and `file ./sbsh` confirms ELF binary

## Notes
Pre-existing failure on `origin/main`: `Test_HandleEvent_EvCmdExited` in `internal/terminal` hangs at `controller.go:264` (chan send blocked) and times out. Reproduced on a clean worktree of `origin/main` — unrelated to this change, which touches only `internal/client`, `pkg/api/client.go`, `pkg/rpcclient/client`, and `pkg/spawn/client.go`.

Refs umbrella: #118 (PR-E). Related: #72.

Closes #132

### DIFF
--- a/internal/client/clientrpc/rpc_server.go
+++ b/internal/client/clientrpc/rpc_server.go
@@ -30,3 +30,34 @@ func (s *ClientControllerRPC) WaitReady(_ *api.Empty, _ *api.Empty) error {
 func (s *ClientControllerRPC) Detach(_ *api.Empty, _ *api.Empty) error {
 	return s.Core.Detach()
 }
+
+func (s *ClientControllerRPC) Ping(in *api.PingMessage, out *api.PingMessage) error {
+	pong, err := s.Core.Ping(in)
+	if err != nil {
+		return err
+	}
+	*out = *pong
+	return nil
+}
+
+func (s *ClientControllerRPC) Metadata(_ api.Empty, doc *api.ClientDoc) error {
+	md, err := s.Core.Metadata()
+	if err != nil {
+		return err
+	}
+	*doc = *md
+	return nil
+}
+
+func (s *ClientControllerRPC) State(_ api.Empty, state *api.ClientStatusMode) error {
+	st, err := s.Core.State()
+	if err != nil {
+		return err
+	}
+	*state = *st
+	return nil
+}
+
+func (s *ClientControllerRPC) Stop(args *api.StopArgs, _ *api.Empty) error {
+	return s.Core.Stop(args)
+}
--- a/internal/client/clientrpc/rpc_server_test.go
+++ b/internal/client/clientrpc/rpc_server_test.go
@@ -1,0 +1,121 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package clientrpc_test
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/eminwux/sbsh/internal/client"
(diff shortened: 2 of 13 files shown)


eminwux:
**PR #136 Review — `feat: grow ClientController with Ping/Metadata/State/Stop RPCs`**

Clean parity pass with the terminal-side controller surface. The wiring (`pkg/api` → `internal/client/clientrpc` → `internal/client` → `pkg/rpcclient/client`) mirrors the existing terminal pattern, `ClientHandle.WaitReady` now has an actual readiness signal via `Ping`, and `Close` flipping from `Detach` to `Stop` is a clear semantic win. Unit coverage plus the real `net/rpc`-over-Unix-socket roundtrip tests in `pkg/rpcclient/client/rpc_client_test.go` is solid. No credential/PII leaks — the only user-controlled string echoed back is the `PingMessage.Message` in the unexpected-ping error, which is fine.

- **Doc overclaims shallow copy as deep** (`internal/client/clientrunner/metadata.go:71-79`): the comment says *"The returned pointer is a deep copy — callers may freely mutate it without racing the runner's own metadata writes."* but `doc := sr.metadata` is a value copy only — `ClientMetadata.Labels`, `ClientMetadata.Annotations`, and `ClientSpec.TerminalSpec` (pointer) remain shared with the runner's copy. In practice the labels/annotations maps are only written during setup so the race is narrow, and this matches `terminalrunner/metadata.go:123-129`'s existing shallow-copy behavior — but please soften the wording to match the terminal-side's honest *"Return a copy to avoid race conditions"* so callers don't rely on a deep-copy guarantee we're not actually providing.
- **Nil-out branches in `pkg/rpcclient/client/rpc_client.go`** (`Metadata`/`State`/`Ping` lines 72-89): when the caller passes a nil output pointer, the code issues the RPC against a throwaway local and discards the reply. That's a silent no-op from the caller's perspective — a nil-check returning a clear error would be friendlier. Non-blocking; purely a caller-ergonomics nit.

Stop's async-goroutine-then-return-nil pattern is the right move for letting the RPC reply flush before the socket tears down, and it matches what terminal's `Controller.Stop` does — good consistency.

LGTM to merge — the items above are non-blocking.

PR #137, opened 2026-04-23
feat: load profiles from a directory tree

### LINKED ISSUE #128: Profile - Load profiles from a directory tree (replaces single file)
## Context
Today, all profiles live in a single multi-document YAML file at `~/.sbsh/profiles.yaml`. Two pains:

1. One large file is awkward to edit, share, or version per project / per integration (e.g. kukeon).
2. The loader is all-or-nothing: a single YAML or schema error in the file makes the loader return zero profiles. Shell completion for `sbsh -p` then silently shows no candidates, with no signal to the user that anything is wrong — they just see an empty list and assume there are no profiles.

## Proposal
Replace the single-file model with a directory-scan model whose loader is resilient and surfaces problems without aborting.

### Layout and discovery
- Default location: `~/.sbsh/profiles.d/`.
- Override via flag `--profiles-dir` or environment variable `SBSH_PROFILES_DIR` (flag wins over env, env wins over default).
- Directory is scanned recursively; subdirectories are allowed for organization.
- Only files matching `*.yaml` or `*.yml` are loaded; everything else is ignored.
- Each file may contain one or many YAML documents.
- The legacy `~/.sbsh/profiles.yaml` single-file path is removed. No implicit migration; document the change in the release notes.

### Resilient loading and warnings
The loader must always return the set of *successfully loaded* profiles plus a list of warnings. A failure in one file must never abort the load of others.

Categories of warning, each naming the offending file (and document index where applicable):
- Malformed YAML (parse error).
- Schema-invalid document (e.g. missing required fields, unknown apiVersion).
- Duplicate profile name — first occurrence in deterministic load order wins; subsequent duplicates are skipped, naming both the skipped file and the file that won.

### Surfacing warnings to the user
- For interactive commands: warnings print to **stderr** at load time, prefixed (e.g. `sbsh: warning: …`).
- For shell completion (`sbsh -p` and friends): completion must always return whatever loaded successfully — never fail or return empty just because one file was bad. Completion scripts typically swallow stderr, so warnings must also be retrievable on demand: `sb validate profiles` (or equivalent) lists every warning with file path and reason, so a user with empty/short completion can run it to diagnose.
- Document this loop in the release notes and the configuration guide: "if completion looks wrong, run `sb validate profiles`."

### Discovery and overrides
- [ ] `sbsh` and `sb` load profiles from `~/.sbsh/profiles.d/` by default.
- [ ] `--profiles-dir <path>` overrides the default; `SBSH_PROFILES_DIR` overrides the default but loses to the flag.
- [ ] Recursive directory traversal works; nested subdirectories are walked.
- [ ] Only `.yaml` and `.yml` files are parsed; other files are skipped silently.
- [ ] Multi-document YAML files inside the directory are fully supported.
- [ ] Load order is deterministic (e.g. lexicographic by full path) so "first wins" is reproducible.
- [ ] Missing or empty directory is not an error — `sbsh` starts with zero profiles loaded.
- [ ] `~/.sbsh/profiles.yaml` is no longer read; removed from defaults, docs, and examples.

### Resilience
- [ ] A malformed YAML file does not prevent other files from loading.
- [ ] A schema-invalid document does not prevent other documents (in the same file or others) from loading.
- [ ] Duplicate profile names produce a warning; the first-loaded profile is kept; each duplicate is skipped with a message naming both the skipped file and the file that won.
- [ ] The loader returns `(profiles, []warnings)` — callers (CLI, completion, validate) consume both.

### User-facing reporting
- [ ] Interactive commands print warnings to stderr at load time, each line naming the file (and document index where relevant) and the reason.
- [ ] Shell completion for `-p` always returns the successfully loaded profile names, even when warnings exist; never returns empty solely because some files failed.
- [ ] `sb validate profiles` lists every warning (malformed, invalid, duplicate) with file path, document index, and reason — so a user can diagnose silent completion.
- [ ] Docs (`README`, `docs/configuration*`, examples) updated to reflect the new layout and the "completion empty? run validate" diagnostic flow.

## Notes
- Related: #30 (Configuration - Introduce ConfigurationDoc config.yaml) currently plans a `profilesFile` key — once this lands, #30 should reference `profilesDir` instead.
- Related: #126 / `sb validate profiles` is the natural home for the warning-listing requirement above.
- Breaking change for existing users; call out clearly in release notes for the version that ships this.

eminwux:
## Summary
- Replaces the single `~/.sbsh/profiles.yaml` layout with a recursive directory scan of `~/.sbsh/profiles.d/` (`*.yaml` / `*.yml`), overridable via `--profiles-dir` or `SBSH_PROFILES_DIR`.
- Makes the loader resilient: malformed YAML, schema-invalid docs, and duplicate profile names become `ProfileWarning`s instead of aborting — callers (CLI, completion, `sb validate profiles`) decide how to surface them. Interactive commands print warnings to stderr; completion swallows them so it never silently empties when one file is bad.
- Extends `sb validate profiles` to walk the configured directory, validate every file in deterministic (lexicographic) order, and list loader warnings — the diagnostic flow for users whose `-p` completion looks empty.
- Renames the config/env/flag plumbing everywhere: `ConfigurationSpec.profilesFile` → `profilesDir`, `SB_PROFILES_FILE` / `SBSH_PROFILES_FILE` → `SB_PROFILES_DIR` / `SBSH_PROFILES_DIR`, `--profiles-file` → `--profiles-dir`, builder option `WithProfileFile` → `WithProfilesDir`.
- Updates user-facing docs (README, profiles guide, configuration guide, faq, cli reference, cicd/container guides, tutorials, concepts, gap-analysis, ROADMAP field list) to reflect the new layout and the "completion empty? run `sb validate profiles`" diagnostic. The legacy single-file path is no longer read; breaking change is called out in the profiles guide.

## Test plan
- [x] `make sbsh-sb` and `file ./sbsh` confirm an ELF executable.
- [x] `go build ./...` and `go vet ./...` clean.
- [x] `go test $(go list ./... | grep -v '/e2e$' | grep -v 'internal/terminal$')` passes — including new `pkg/discovery` coverage for recursive walk, malformed YAML, duplicate first-wins, unsupported kind, empty/missing directory, and a directory fixture for `sb validate profiles`.
- [x] `go test -tags=integration ./cmd/sb/get/...` passes.
- [ ] `go test ./internal/terminal/...` — **pre-existing flake unrelated to this change**: `Test_HandleEvent_EvCmdExited` hangs to the 10-minute timeout. Reproduces on `origin/main` (7a60316) in an isolated worktree. Not introduced here.

## Release notes
The repo has no CHANGELOG yet, so this PR description serves as the release note. Breaking change:
- `~/.sbsh/profiles.yaml` is no longer read. Move contents into `~/.sbsh/profiles.d/` — a single combined file or many small files both work, the loader scans recursively.
- `SBSH_PROFILES_FILE` / `SB_PROFILES_FILE` env vars and the `--profiles-file` flag are renamed to `SBSH_PROFILES_DIR` / `SB_PROFILES_DIR` / `--profiles-dir`.
- `ConfigurationSpec.profilesFile` is now `profilesDir`.

Closes #128

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -18,7 +18,7 @@
 - [ ] **A** API - Add SupervisorMetadata to SupervisorDoc
 - [ ] **A** API - Add TerminalMetadata to TerminalDoc
 - [ ] **A** Configuration - Introduce ConfigurationDoc config.yaml
-    (fields: runPath, profilesFile, logLevel)
+    (fields: runPath, profilesDir, logLevel)
 - [ ] **A** Configuration - Introduce profilesDirectory in ConfigurationDoc with conflict resolution
 - [ ] **A** Configuration - Introduce default TerminalProfileDoc in ConfigurationDoc
 - [ ] **A** Supervisor - Introduce SupervisorDoc and SupervisorProfileDoc
--- a/cmd/config/autocomplete.go
+++ b/cmd/config/autocomplete.go
@@ -26,20 +26,23 @@ import (
 	"github.com/eminwux/sbsh/pkg/api"
 )
 
-func AutoCompleteListProfileNames(ctx context.Context, logger *slog.Logger, profilesFile string) ([]string, error) {
+// AutoCompleteListProfileNames returns every profile name successfully loaded
+// from profilesDir. Warnings (malformed files, duplicate names, schema-invalid
+// documents) are swallowed so completion never fails or returns an empty list
+// just because one file is broken — users can diagnose those with
+// `sb validate profiles`.
+func AutoCompleteListProfileNames(ctx context.Context, logger *slog.Logger, profilesDir string) ([]string, error) {
 	// logger is not set on autocomplete calls
 	if logger == nil {
 		logger = logging.NewNoopLogger()
 	}
 
-	profiles, err := discovery.LoadProfilesFromPath(ctx, logger, profilesFile)
+	profiles, _, err := discovery.LoadProfilesFromDir(ctx, logger, profilesDir)
 	if err != nil {
-		if logger != nil {
-			logger.ErrorContext(ctx, "ListProfiles: failed to load profiles", "path", profilesFile, "error", err)
-		}
+		logger.ErrorContext(ctx, "ListProfiles: failed to load profiles", "dir", profilesDir, "error", err)
 		return nil, err
 	}
-	if profiles == nil {
+	if len(profiles) == 0 {
 		return nil, errors.New("no profiles found")
 	}
 
--- a/cmd/config/configuration.go
+++ b/cmd/config/configuration.go
@@ -44,8 +44,8 @@ func ApplyConfigurationDocEnv(cfgDoc *api.ConfigurationDoc) {
 	}
 	setIfUnset(SB_ROOT_RUN_PATH.EnvVar(), cfgDoc.Spec.RunPath)
 	setIfUnset(SBSH_ROOT_RUN_PATH.EnvVar(), cfgDoc.Spec.RunPath)
-	setIfUnset(SB_GET_PROFILES_FILE.EnvVar(), cfgDoc.Spec.ProfilesFile)
-	setIfUnset(SBSH_ROOT_PROFILES_FILE.EnvVar(), cfgDoc.Spec.ProfilesFile)
+	setIfUnset(SB_GET_PROFILES_DIR.EnvVar(), cfgDoc.Spec.ProfilesDir)
+	setIfUnset(SBSH_ROOT_PROFILES_DIR.EnvVar(), cfgDoc.Spec.ProfilesDir)
 	setIfUnset(SB_ROOT_LOG_LEVEL.EnvVar(), cfgDoc.Spec.LogLevel)
 	setIfUnset(SBSH_ROOT_LOG_LEVEL.EnvVar(), cfgDoc.Spec.LogLevel)
 }
--- a/cmd/config/configuration_test.go
+++ b/cmd/config/configuration_test.go
@@ -53,7 +53,7 @@ metadata:
   name: default
 spec:
   runPath: /var/run/sbsh
-  profilesFile: /etc/sbsh/profiles.yaml
+  profilesDir: /etc/sbsh/profiles.d
   logLevel: debug
 `
 	if err := os.WriteFile(path, []byte(content), 0o644); err != nil {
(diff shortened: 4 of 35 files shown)


eminwux:
**PR #137 Review — `feat: load profiles from a directory tree`**

Substantive change, but cleanly scoped: the new recursive walk + warning-as-first-class-citizen model in `pkg/discovery/profiles.go` is well-structured, the config/env/flag/builder rename (`profilesFile` → `profilesDir`, `WithProfileFile` → `WithProfilesDir`, etc.) is consistent across every call site I checked, and `sb validate profiles` turning into the diagnostic escape hatch for "why is my completion empty?" is a good UX pivot. Table/JSON/YAML output, per-doc index in warnings, and lexicographic first-wins deduplication all look right. No sensitive-data leaks — warnings print file paths, which are user-configured and expected in diagnostics. Test coverage in `pkg/discovery/profiles_test.go` hits the interesting branches (recursive walk, malformed YAML, duplicates, unsupported apiVersion/kind, empty/missing dir, `FindProfileByNameInDir`).

- **Dead public-API** (`pkg/discovery/profiles.go:279-315` and its wrapper at `internal/discovery/profiles.go:78-84`): `LoadProfilesFromReaderWithContext` had one caller pre-PR (via the deleted `LoadProfilesFromPath`) and is now unreferenced everywhere — production, tests, and e2e. The docstring you updated even hints at this (*"exists primarily for tests and callers that already have a stream in hand"*) but no such caller exists. Worth deleting in a follow-up since `decodeProfileStream` covers the internal need and leaving an exported-but-unused helper in `pkg/discovery` is a soft API commitment. Not blocking since this is leftover from before, not newly introduced.
- **Comment overclaims** (`cmd/sb/validate/profiles.go:203-206`): the comment says loader warnings *"go to stdout… then also to stderr so interactive pipelines see them"* but the code only writes them to stdout (line 210). Either drop the stderr clause or add the missing `fmt.Fprintln(stderr, …)` — pick whichever matches intent.
- **Double directory walk** (`cmd/sb/validate/profiles.go:114-147`): `expandPath` first calls `LoadProfilesFromDir` (which internally calls `collectYAMLFiles`), then runs its own `filepath.WalkDir` to collect the same file list. Cheap operation, but re-walking the tree twice per invocation is wasteful; consider having `LoadProfilesFromDir` return the resolved file list, or expose `collectYAMLFiles` so the validate command can reuse it.
- **Stale test-local names** (`cmd/sb/get/get_integration_test.go:87, 319`): `profilesFile` is now misleading since the loader only accepts directories — the tests at lines 324/334 that still pass file paths happen to return the "not a directory" error path, which the tests then accept as "expected error". Works today but semantically muddled; worth renaming + narrowing what those subtests actually verify.

Stop's async-goroutine-then-return-nil pattern is the right move for letting the RPC reply flush before the socket tears down, and it matches what terminal's `Controller.Stop` does — good consistency. (sorry, wrong PR — ignore that last sentence.)

LGTM to merge — all items above are non-blocking follow-ups.

PR #139, opened 2026-04-24
fix(controller): atomic Close guard + closeReqCh priority fix deadlock (#138)

### LINKED ISSUE #138: bug: internal/terminal Controller.Close races shuttingDown flag, causing Test_HandleEvent_EvCmdExited to deadlock
## Summary

`Test_HandleEvent_EvCmdExited` in `internal/terminal/controller_test.go:537` deadlocks
deterministically and the test goroutine never exits. CI/local `make test` is forced
to wait for the per-run timeout (60s isolated, 300s in the full suite) before failing.
Reproduces on a clean `origin/main` worktree at commit `7a60316`.

## Reproduction

```bash
go test -timeout=60s -run '^Test_HandleEvent_EvCmdExited$' -v ./internal/terminal/
# panic: test timed out after 1m0s
```

## Goroutine trace (excerpt)

```
goroutine 17 [chan receive]:
github.com/eminwux/sbsh/internal/terminal.Test_HandleEvent_EvCmdExited
        internal/terminal/controller_test.go:593

goroutine 18 [chan send]:
github.com/eminwux/sbsh/internal/terminal.(*Controller).Close
        internal/terminal/controller.go:264                       <-- c.closeReqCh <- reason
github.com/eminwux/sbsh/internal/terminal.(*Controller).Run
        internal/terminal/controller.go:196                       <-- Close called from <-c.ctx.Done() branch
```

## Root cause analysis

`Close` guards its body with `if !c.shuttingDown.Load()`, but the flag is flipped
in a separate goroutine that reads from `closingCh` (controller.go:185-189).
That leaves a race window where `Close` runs twice before `shuttingDown` becomes
`true`:

1. EvCmdExited fires → `handleEvent` → `onClosed` → `Close(err)`. The first call:
   - `c.cancel(reason)` cancels the controller context.
   - `c.closingCh <- reason` succeeds (buffer cap 1, now full).
   - `c.closeReqCh <- reason` succeeds (buffer cap 1, now full).
   - Returns. `shuttingDown` is **still false** because the watcher goroutine
     hasn't been scheduled yet.
2. `Run` loops back to its `select`. `c.ctx.Done()` is now ready (was just
   canceled in step 1) and is picked before `closeReqCh`.
3. Inside the `<-c.ctx.Done()` branch, `Run` calls `c.Close(c.ctx.Err())` at
   controller.go:196. `shuttingDown` is still false, so `Close` enters the body
   again and tries `c.closeReqCh <- reason` — but the buffer is full and the
   only reader (the same `Run` goroutine) is sitting inside this very `Close`
   call. **Deadlock.**

## Suggested fix direction

Make the shutdown gate atomic with the action it guards, e.g.:

- Replace the `closingCh`-reader goroutine with `c.shuttingDown.CompareAndSwap(false, true)`
  at the top of `Close`, so only the first caller proceeds.
- Or make `closeReqCh` non-blocking (`select { case c.closeReqCh <- reason: default: }`)
  since it's a one-shot wakeup.

The same pattern exists in `internal/client/controller.go` and is worth auditing
in the same change.

## Impact

- Local `make test` always waits the full 5-minute test timeout before failing.
- CI `Unit tests` step is unreliable / slow for the same reason.
- Any new code path that triggers `EvCmdExited` followed by ctx cancellation will
  hit this.

SAY-5:
## Summary

`Test_HandleEvent_EvCmdExited` in `internal/terminal` deadlocked deterministically because `Close` relied on a watcher goroutine to flip `shuttingDown` asynchronously. When `EvCmdExited` triggered `Close`, the context cancel inside `Close` immediately woke `Run`'s `<-c.ctx.Done()` branch, which called `Close` again before the watcher had been scheduled. Both invocations saw `shuttingDown=false` and the second blocked forever on `closeReqCh` (buffer 1, already filled by the first caller; only reader was `Run`, sitting inside the second `Close`).

## Fix

1. **Atomic shutdown guard.** Replace `shuttingDown.Load()`-then-body with a single `CompareAndSwap` at the top of `Close`. The guard is now atomic with the body it guards, so re-entries return immediately instead of re-filling the one-shot channel.
2. **closeReqCh priority.** In `Run`'s main loop, prioritise `closeReqCh` over `ctx.Done` with a non-blocking pre-check. When `Close` cancels ctx and sends on `closeReqCh`, both branches become ready simultaneously; Go's default select picks randomly, which could surface `ErrContextDone` for a close we initiated ourselves. Deterministic priority makes the returned error reliable.
3. **Drop the watcher goroutine.** `closingCh -> shuttingDown.Store` goroutine is now redundant; `WaitClose` still reads `closingCh` for observability.
4. **Parallel fix in `internal/client/controller.go`** (flagged in the issue as "worth auditing in the same change").

Fixes #138

## Test plan

- [x] `go test -run '^Test_HandleEvent_EvCmdExited$' -timeout 30s ./internal/terminal/` — passes in 0.01s (was deadlocking with the 60s hard timeout)
- [x] `go test -timeout 120s -race ./internal/terminal/` — full terminal suite green with `-race`
- [x] `go test -timeout 120s -race ./internal/client/` — client controller suite green with `-race`
- [x] Traced both close callers (handleEvent/onClosed and Run's ctx.Done branch) to confirm CAS short-circuits the second entry

### DIFF
--- a/internal/client/controller.go
+++ b/internal/client/controller.go
@@ -222,13 +222,27 @@ func (s *Controller) Run(doc *api.ClientDoc) error {
 	s.logger.Info("controller ready, entering main event loop")
 	close(s.ctrlReadyCh)
 
-	go func() {
-		errC := <-s.closingCh
-		s.shuttingDown.Store(true)
-		s.logger.Warn("controller closing", "reason", errC)
-	}()
+	// shuttingDown is now flipped atomically via CompareAndSwap at the
+	// top of Close, so we no longer need a background goroutine to
+	// observe closingCh and set the flag. Leaving a watcher opened the
+	// same re-entrancy deadlock documented in issue #138 for the
+	// terminal controller.
 
 	for {
+		// closeReqCh takes precedence over ctx.Done so an internal
+		// shutdown (handleEvent -> Close) returns ErrCloseReq rather
+		// than racing with the ctx cancellation that Close itself just
+		// performed.
+		select {
+		case errClose := <-s.closeReqCh:
+			s.logger.Warn("close request received", "error", errClose)
+			if errClose == nil {
+				return nil
+			}
+			return fmt.Errorf("%w: %w", errdefs.ErrCloseReq, errClose)
+		default:
+		}
+
 		select {
 		case <-s.ctx.Done():
 			var errDone error
@@ -359,22 +373,27 @@ func (s *Controller) onClosed(_ api.ID, err error) {
 }
 
 func (s *Controller) Close(reason error) error {
-	if !s.shuttingDown.Load() {
-		s.logger.Info("initiating shutdown sequence", "reason", reason)
-		// Set closing reason
-		s.closingCh <- reason
-
-		// Notify terminal runner to close all terminals
-		_ = s.sr.Close(reason)
-
-		// Notify Run to exit
-		s.closeReqCh <- reason
-
-		// Mark controller as closed
-		close(s.closedCh)
-	} else {
+	// CompareAndSwap guarantees only the first caller enters the
+	// shutdown body. See terminal.Controller.Close for the full
+	// explanation (issue #138).
+	if !s.shuttingDown.CompareAndSwap(false, true) {
 		s.logger.Info("shutdown sequence already in progress, ignoring duplicate request", "reason", reason)
+		return nil
 	}
+
+	s.logger.Info("initiating shutdown sequence", "reason", reason)
+
+	// Set closing reason
+	s.closingCh <- reason
+
+	// Notify terminal runner to close all terminals
+	_ = s.sr.Close(reason)
+
+	// Notify Run to exit
+	s.closeReqCh <- reason
+
+	// Mark controller as closed
+	close(s.closedCh)
 	return nil
 }
 
--- a/internal/terminal/controller.go
+++ b/internal/terminal/controller.go
@@ -182,13 +182,27 @@ func (c *Controller) Run(spec *api.TerminalSpec) error {
 	c.logger.InfoContext(c.ctx, "controller ready")
 	close(c.ctrlReadyCh)
 
-	go func() {
-		err := <-c.closingCh
-		c.shuttingDown.Store(true)
-		c.logger.WarnContext(c.ctx, "controller closing", "reason", err)
-	}()
+	// shuttingDown is now flipped atomically at the top of Close via
+	// CompareAndSwap, so we no longer need a background goroutine to
+	// observe closingCh and set the flag. Leaving a watcher here opened
+	// a race where Close could be re-entered between the closingCh send
+	// and the Store — deadlocking on the one-shot closeReqCh buffer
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #139 Review — `fix(controller): atomic Close guard + closeReqCh priority fix deadlock`**

Clean, targeted fix for the #138 deadlock. The CAS replaces a racy Load-then-act pattern and correctly makes re-entry a no-op, and the priority pre-check on `closeReqCh` at the top of `Run`'s loop catches the internal `handleEvent → onClosed → Close → return` path deterministically. Verified locally: `Test_HandleEvent_EvCmdExited` passes in ~0.01s (was deadlocking), full `./internal/terminal/` and `./internal/client/` suites green. No sensitive-data leaks in the new logs. LGTM.

- **Scope of the priority claim** (`internal/terminal/controller.go:193-207`, same shape in client `controller.go:230-243`): the priority `closeReqCh` pre-check only runs when `Run` re-enters the loop — i.e. after an internal `handleEvent → Close` returns. For a close triggered from outside `Run`'s goroutine, `Close` calls `c.cancel(reason)` *before* sending on `closeReqCh`, so `Run` can already be woken on `ctx.Done` alone (closeReqCh not yet filled) and still return `ErrContextDone`. The deadlock is fixed either way, but the PR description's "deterministic priority makes the returned error reliable" holds only for the internal-close path that the test exercises, not for concurrent external callers. Worth tightening the claim in the commit message / comment, or ignoring if external Close isn't a real caller today.
- **Comment path** (`internal/client/controller.go:231`): comment says `handleEvent -> Close`, but the actual path is `handleEvent -> onClosed -> Close`. One-word fix.

LGTM as-is — these are both wording nits, not behavior issues.

eminwux:
One more pass before merge — bouncing this back for a light cleanup. The fix itself is right; the asks are style-only:

- **Tighten the new comment blocks.** The three added blocks (watcher-removal rationale in `Run`, priority-check rationale, CAS rationale in `Close`) each run 5–8 lines and retell the same #138 story. One short line per site is enough — e.g. `// CAS ensures a single shutdown entrant; see #138.` and `// closeReqCh wins over ctx.Done so internal Close returns ErrCloseReq.` The issue carries the long version. Same applies to the duplicated blocks in `internal/client/controller.go`.
- **Fix the close-path comment** in `internal/client/controller.go:231`: says `handleEvent -> Close`, actual path is `handleEvent -> onClosed -> Close`.
- **Scope the priority claim** in the commit message (and the comment above the pre-check) to "internal close path" — external callers still hit the `ctx.Done`/`closeReqCh` race since `cancel()` fires before the `closeReqCh` send. Doesn't change the fix, just the claim.

No code changes needed beyond the above. Flip back to `ready-for-review` once pushed.

eminwux:
Solved in #185. Thanks @SAY-5 for contributing, but we needed to ship this quicky; it randomly blocked testing. Happy to provide you with some other issue you can work on with more time.

PR #143, opened 2026-04-24
feat: PID-1 reaper, signal forwarder, and graceful shutdown

### LINKED ISSUE #120: terminal: proper PID 1 behavior - reap orphans and forward signals to child
## Motivation

When `sbsh terminal` runs as PID 1 of a Linux container (the target use case for **`eminwux/kukeon#57` UC2** — PTY-attached workloads under kukeon-injected sbsh), today's supervisor behavior is subtly wrong in two specific ways. This issue also replaces one always-on default (immediate SIGKILL on shutdown) that's orthogonal to PID 1 but naturally belongs in the same change.

## Current behavior (factual)

- Own-process signal handling at `cmd/sbsh/terminal/terminal.go:311` installs `signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)` — SIGINT/SIGTERM on the sbsh process cancel the context, which triggers `Close`.
- `Close` kills the child via `sr.cmd.Process.Kill()` at `internal/terminal/terminalrunner/lifecycle.go:105-108` — which sends **SIGKILL**, not SIGTERM.
- Child is waited on via `os/exec.Cmd.Wait()` only. No `Wait4(-1, ...)` / `SIGCHLD` reaper.
- Multi-arch static binary (`CGO_ENABLED=0`, amd64/arm64) is already produced by `make release-build`. Release-pipeline work (arch-tagged release assets for kukeon to fetch) is **out of scope here** — if kukeon's resolver needs a specific asset layout, that lives in a separate issue.

### Gap A — Zombie reaping

As PID 1 of a PID namespace, sbsh inherits any orphaned descendant (e.g. a grandchild whose parent exited). Go's runtime and `os/exec` only wait on processes sbsh itself spawned via `exec.Cmd`. Orphans that reparent to PID 1 are never reaped and become `<defunct>` entries that accumulate indefinitely. This is the classic problem `tini` / `dumb-init` solve.

### Gap B — Signal forwarding (not just own-process signal handling)

Today, SIGTERM to sbsh triggers an immediate `Process.Kill()` = SIGKILL on the child — no graceful shutdown path for the workload. Other signals relevant to typical workloads (SIGHUP, SIGQUIT, SIGINT when the container is attached to a TTY) are not forwarded at all. For `kuke stop cell` to ever be graceful, the container's PID 1 must forward SIGTERM and give the workload a chance to exit cleanly before escalating.

## Proposed design

Three semantically independent sub-changes, bundled in one PR because they touch the same code paths (lifecycle, command construction, signal plumbing). Each has a distinct activation gate — call them out explicitly rather than pretending there's a single gate.

### Sub-change 1 — PID-1-gated reaper

- Activates when `os.Getpid() == 1`. No user-facing flag. Linux-only (darwin/freebsd compile a no-op via build tags).
- Goroutine listens for `SIGCHLD`. On each signal (and at least once at startup), drain with `syscall.Wait4(-1, &status, unix.WNOHANG, nil)` in a loop until the call returns `0` or `ECHILD`.
- **Reaper vs `os/exec.Cmd.Wait()` ECHILD race.** With the reaper active, kernel-level `SIGCHLD` consumption means `os/exec.Cmd.Wait()` can race and hit `ECHILD` before seeing the child's real exit status. When the reaped PID matches the tracked `cmd.Process.Pid`, feed the status back to the existing `waitOnTerminal` path via a channel instead of letting `cmd.Wait()` handle it. This is the trickiest part of the change and deserves its own dedicated test — see acceptance.

### Sub-change 2 — PID-1-gated signal forwarder

- Same activation gate as the reaper (`os.Getpid() == 1`). Outside PID 1, host-side signal handling is unchanged — SIGINT/SIGHUP/SIGQUIT on host `sbsh` are **not** forwarded to the child, same as today.
- Minimum signal set forwarded: **SIGTERM, SIGINT, SIGHUP, SIGQUIT**.
- Target: the tracked child's **process group** (`syscall.Kill(-pgid, sig)`), not just the child PID — catches misbehaving descendants. Requires the child to be launched with `SysProcAttr.Setpgid: true` (verify and set if not already).
- SIGWINCH stays out of this path — PTY resize is already handled separately.

### Sub-change 3 — Graceful shutdown (always on)

- **Different activation gate from subs 1–2**: applies *whenever* sbsh initiates shutdown of the child (context cancel, `Close(reason)`), regardless of PID 1 status. Removing the "always hard-kill" default is correct everywhere.
- Replace today's immediate SIGKILL path with: SIGTERM to the child (or its process group, at PID 1) → wait up to `--shutdown-grace` for the child to exit → escalate to SIGKILL if still running.
- `--shutdown-grace` is a duration flag on `sbsh terminal`, bound via viper like the existing flags.
- **Default: 30s**, matching Kubernetes' `terminationGracePeriodSeconds`. kukeon (#57) is the primary injector; aligning the default makes the common integration predictable.

### Test hook

For unit/e2e tests that need to exercise the PID-1 paths without actually being PID 1, expose an internal `initmode.Enable(bool)` helper. Must be called before `terminalrunner.Start`. Unit-level only — real PID-1 assertions need a namespace (see acceptance).

## Non-goals

- Not a generic `tini` replacement. sbsh still owns exactly one tracked child + a PTY; this issue does not introduce supervision trees, restart policies, or multi-child semantics.
- No changes to darwin/freebsd runtime behavior.
- No changes to the release pipeline. Multi-arch binary already exists.
- No `PR_SET_CHILD_SUBREAPER` — we rely solely on actually being PID 1.
- No user-facing `--init-mode` flag.
- No extension of signal forwarding to non-PID-1 sbsh. Host-side signal behavior is unchanged except for the SIGTERM-then-SIGKILL shutdown default.

## Acceptance

Split by test level so unit and e2e responsibilities are unambiguous.

### Unit (`go test`, `initmode.Enable(true)`)

- **Reaper drains orphans:** synthetic orphan fork is reaped; `Wait4` loop terminates on `ECHILD`/`0`; no defunct leaks within the test's scope.
- **Reaper / `cmd.Wait()` race is resolved:** when the reaper consumes the tracked child's `SIGCHLD`, the `waitOnTerminal` path still receives the real exit status (not `ECHILD`, not a silent zero).
- **Signal forwarder targets the process group:** a child launched with `Setpgid: true` plus a sibling in the same pgroup both receive forwarded SIGTERM.
- **Graceful shutdown — happy path:** child exits on SIGTERM before the grace period; sbsh does not send SIGKILL.
- **Graceful shutdown — escalation:** child ignores SIGTERM; SIGKILL is sent at `--shutdown-grace`; sbsh exits promptly afterward.

### E2E (real PID 1 via `unshare -Urp` or throwaway container)

- **Orphan reaping at PID 1:** child forks-then-exits an orphan grandchild; `ps` inside the namespace shows no `<defunct>` within 100 ms of the orphan's exit.
- **Graceful shutdown at PID 1:** SIGTERM to sbsh reaches the child's process group; a `trap TERM` handler in a shell child observes it; sbsh exits after the child exits OR after `--shutdown-grace`, whichever comes first.
- **No regression outside PID 1:** existing e2e tests pass unchanged. Host-side `sbsh terminal` shutdown now sends SIGTERM first (then SIGKILL after the grace period) instead of immediate SIGKILL — intentional, user-visible improvement, documented in the changelog.

## Related

- **Hard blocker for `eminwux/kukeon#57`** (UC2 — inject sbsh as PID 1 for PTY-attached workloads).
- Foundation (shared control plane) lands in `eminwux/kukeon#39` (UC1); does not block or get blocked by this issue.
- `eminwux/kukeon#46` (Session) — soft integration via sbsh's terminal-I/O observation API as an `idleTimeout` activity signal. Not a blocker in either direction.
- Independent of #118 (library consumer API).

eminwux:
## Summary

- **Zombie reaping** (`internal/terminal/terminalrunner/reaper_linux.go`): when `initmode.IsInit()` is true (default: `os.Getpid() == 1`), a SIGCHLD-driven `Wait4(-1, WNOHANG, ...)` loop drains orphans that reparent to sbsh. No `PR_SET_CHILD_SUBREAPER`; we rely solely on actually being PID 1.
- **Reaper / `cmd.Wait` race resolved**: when the reaper is active, `watchChildExit` sources the tracked child's exit status from the reaper's `TrackedExitCh` (not `os/exec.Cmd.Wait()`, which would race to `ECHILD`). `cmd.Process.Release()` drops the handle afterward. Outside init mode the path is unchanged: plain `cmd.Wait()`.
- **Signal forwarder** (`signals_linux.go`, same init-mode gate): forwards SIGTERM/SIGINT/SIGHUP/SIGQUIT to the tracked child's process group via `syscall.Kill(-pgid, …)`. `Setsid: true` on the child already makes it the pgroup leader, so no explicit `Setpgid` is needed. SIGWINCH stays on its dedicated PTY-resize path.
- **Graceful shutdown (always on)**: `Exec.Close` now runs SIGTERM → wait up to `ShutdownGrace` → SIGKILL on the child's process group, replacing the unconditional `Process.Kill()`. `cmd.Cancel` is a no-op so ctx-cancel doesn't pre-empt the grace period.
- **`--shutdown-grace` duration flag** on `sbsh terminal` (default 30s, matching Kubernetes `terminationGracePeriodSeconds`), bound via viper and threaded through `TerminalSpec.ShutdownGrace` + `profile.DefaultShutdownGrace`.
- **Test hook**: new `internal/initmode` package with `Enable(bool)` / `Reset()` so unit tests can exercise PID-1 code paths without actually being PID 1.

Build-tag split: Linux is the real implementation; non-Linux files (`*_other.go`) are no-op stubs. sbsh only runs as PID 1 in Linux containers.

## Test plan

- [x] `make sbsh-sb` + `file ./sbsh | grep 'ELF 64-bit LSB executable'`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=60s ./internal/initmode/... ./internal/terminal/terminalrunner/... ./internal/profile/... ./cmd/sbsh/... ./cmd/sb/...` — all green
- [x] `go test -tags=integration ./cmd/sb/get/...` — ok
- [x] `./sbsh terminal --help | grep shutdown-grace` — flag appears with `default 30s`
- [x] Reaper unit tests: `TestReaper_DrainsOrphan`, `TestReaper_TrackedExitResolvesRace` (exit code 7 survives the SIGCHLD race)
- [x] Signal forwarder unit test: `TestSignalForwarder_TargetsProcessGroup` (leader + sibling subshell in same pgroup both run their HUP trap via filesystem marker)
- [x] Graceful shutdown unit tests: `TestShutdownChild_HappyPath` (SIGTERM trap exits before grace, no SIGKILL) and `TestShutdownChild_Escalation` (SIGTERM ignored, SIGKILL at grace)
- [x] Library-consumer example still builds (`docs/examples/library-consumer`)

### Known pre-existing failure

`go test ./internal/terminal/...` deadlocks on `Test_HandleEvent_EvCmdExited` with the exact same stack trace on `origin/main` (verified via `git worktree`). That deadlock is the subject of the already-open PR #139 and is unrelated to this change.

### Out of scope (per issue)

- No release-pipeline or multi-arch asset changes.
- No user-facing `--init-mode` flag.
- No generic tini replacement — sbsh still owns exactly one tracked child + PTY.
- E2E tests with `unshare -Urp` (CI does not run `./e2e`; the five unit-level acceptance scenarios from the issue are all covered).

Closes #120

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -204,4 +204,6 @@ var (
 	SBSH_TERM_SPEC = DefineKV("SBSH_TERM_SPEC", "sbsh.terminal.spec")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_DISABLE_SET_PROMPT = DefineKV("SBSH_TERM_DISABLE_SET_PROMPT", "sbsh.terminal.disableSetPrompt")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_TERM_SHUTDOWN_GRACE = DefineKV("SBSH_TERM_SHUTDOWN_GRACE", "sbsh.terminal.shutdownGrace")
 )
--- a/cmd/sbsh/terminal/terminal.go
+++ b/cmd/sbsh/terminal/terminal.go
@@ -92,6 +92,7 @@ func buildTerminalSpecFromFlags(cmd *cobra.Command, logger *slog.Logger) (*api.T
 			LogLevel:         viper.GetString(config.SBSH_TERM_LOG_LEVEL.ViperKey),
 			SocketFile:       viper.GetString(config.SBSH_TERM_SOCKET.ViperKey),
 			DisableSetPrompt: viper.GetBool(config.SBSH_TERM_DISABLE_SET_PROMPT.ViperKey),
+			ShutdownGrace:    viper.GetDuration(config.SBSH_TERM_SHUTDOWN_GRACE.ViperKey),
 		},
 	)
 
@@ -363,6 +364,13 @@ func setupTerminalCmdFlags(terminalCmd *cobra.Command) {
 
 	terminalCmd.Flags().Bool("disable-set-prompt", false, "Disable setting the prompt")
 	_ = viper.BindPFlag(config.SBSH_TERM_DISABLE_SET_PROMPT.ViperKey, terminalCmd.Flags().Lookup("disable-set-prompt"))
+
+	terminalCmd.Flags().Duration(
+		"shutdown-grace",
+		profile.DefaultShutdownGrace,
+		"How long to wait after SIGTERM before escalating to SIGKILL on the child",
+	)
+	_ = viper.BindPFlag(config.SBSH_TERM_SHUTDOWN_GRACE.ViperKey, terminalCmd.Flags().Lookup("shutdown-grace"))
 }
 
 func runTerminal(
--- a/internal/initmode/initmode.go
+++ b/internal/initmode/initmode.go
@@ -1,0 +1,58 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package initmode reports whether the current sbsh process should take on
+// PID-1 responsibilities (zombie reaping, signal forwarding). It exists so
+// those code paths can be exercised from tests without actually being PID 1.
+package initmode
+
+import (
+	"os"
+	"sync/atomic"
+)
+
+// override, when non-zero, forces IsInit to return true (1) or false (2)
+// regardless of os.Getpid(). Tests flip this via Enable.
+var override atomic.Int32
+
+// IsInit reports whether this process should act as container init. Default
+// is os.Getpid() == 1. Tests can force the value via Enable.
+func IsInit() bool {
+	switch override.Load() {
+	case 1:
+		return true
+	case 2:
+		return false
+	default:
+		return os.Getpid() == 1
+	}
+}
+
+// Enable forces IsInit to return on. It must be called before any subsystem
+// that reads IsInit has started. Passing false forces IsInit off.
+func Enable(on bool) {
+	if on {
+		override.Store(1)
+	} else {
+		override.Store(2)
+	}
+}
+
+// Reset clears any Enable override, returning IsInit to the os.Getpid()
+// default. Intended for test teardown.
+func Reset() {
+	override.Store(0)
+}
--- a/internal/initmode/initmode_test.go
+++ b/internal/initmode/initmode_test.go
@@ -1,0 +1,53 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package initmode
+
+import (
+	"os"
+	"testing"
+)
+
+func TestIsInit_DefaultMatchesPID(t *testing.T) {
+	Reset()
+	t.Cleanup(Reset)
+
+	want := os.Getpid() == 1
+	if got := IsInit(); got != want {
+		t.Fatalf("IsInit()=%v, want %v (pid=%d)", got, want, os.Getpid())
+	}
+}
+
+func TestEnable_ForcesValue(t *testing.T) {
+	Reset()
+	t.Cleanup(Reset)
+
+	Enable(true)
+	if !IsInit() {
+		t.Fatalf("Enable(true): IsInit()=false, want true")
+	}
+
+	Enable(false)
+	if IsInit() {
+		t.Fatalf("Enable(false): IsInit()=true, want false")
+	}
+
+	Reset()
+	want := os.Getpid() == 1
+	if got := IsInit(); got != want {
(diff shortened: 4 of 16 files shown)


eminwux:
Re-running failed `Build & Test` CI job once ([run](https://github.com/eminwux/sbsh/actions/runs/24902157121)). Will revisit when it finishes; if it fails again the PR will be marked `ci-stalled`.

eminwux:
**PR #143 Review — `feat: PID-1 reaper, signal forwarder, and graceful shutdown`**

Thoughtful, well-scoped PID-1 implementation — the `initmode` toggle makes the whole thing testable without being PID 1, the Linux/non-Linux build-tag split is clean, the reaper/Cmd.Wait race is explicitly documented and tested (`TestReaper_TrackedExitResolvesRace` survives exit code 7), and the unit-level acceptance scenarios from #120 are all covered. No sensitive-data leaks in the new logs/errors — PIDs and signals only, as appropriate for diagnostics. No dead exported API in the diff.

A few small, non-blocking items:

- **Robustness** (`internal/terminal/terminalrunner/lifecycle.go:watchChildExit`, init-mode branch): the receive on `sr.reaper.TrackedExitCh()` is a single-arm blocking read. In the SIGKILL-escalation path, there is a narrow window where `Close` can reach `reaper.Stop()` before the kernel has generated `SIGCHLD` for the forced-killed child; the reaper drains and exits, nothing ever publishes to `TrackedExitCh`, and `watchChildExit` leaks. Wrapping in `select { case code, ok := <-sr.reaper.TrackedExitCh(): … case <-sr.ctx.Done(): … }` closes the leak. Very narrow window in practice — non-blocking.
- **Concurrency nit** (`internal/terminal/terminalrunner/reaper_linux.go:Stop`): the check-then-`close(r.stop)` is not atomic; two concurrent `Stop()` calls can both pass the select and both call `close`, panicking on the second. `Exec.Close` is the only caller today so it is theoretical, but a `sync.Once` guard is cheap insurance.
- **Test noise** (`internal/terminal/terminalrunner/lifecycle_shutdown_test.go` end of `TestShutdownChild_Escalation`): the trailing `_ = api.Exited` plus comment ("the runner did not synthesize a TerminalState api.Exited transition…") reads like a leftover — nothing is actually being asserted, and `api` is otherwise unused in the test. Either drop the line and the `pkg/api` import, or turn it into a real assertion.
- **CI status**: the `Build & Test` job failed on the first run; I re-triggered it once via `ci-retried` and it went green on the retry, so no action needed from you.

LGTM to proceed — the items above are all non-blocking, and can land in a follow-up or be dropped at your discretion.

PR #144, opened 2026-04-24
fix(attach): accept --id/--name without positional arg

### LINKED ISSUE #142: spawn: NewClient attach by ID fails — `sb attach --id` rejected without positional arg
## Repro

Reproduced on `main` at 156f8e9 while developing the library-consumer example (PR #141).

1. Boot a terminal and obtain its ID+Name (any path).
2. Call `pkg/spawn.NewClient` with a `ClientDoc` whose `Spec.TerminalSpec.ID` is set (and optionally `Name` too):

   ```go
   doc, _ := builder.BuildClientDoc(ctx, logger, stateRoot,
       builder.WithClientMode(api.AttachToTerminal),
       builder.WithClientTerminalSpec(&api.TerminalSpec{
           ID: termSpec.ID, // <-- any non-empty ID triggers this
       }),
   )
   _, err := spawn.NewClient(ctx, doc, spawn.ClientOptions{
       BinaryPath: sbPath,
       ExtraArgs:  []string{"--run-path", stateRoot},
       ...
   })
   // WaitReady never succeeds; child exits immediately.
   ```

   Equivalent direct CLI repro: `./sb attach --id any-id --socket /tmp/any`.

## Observed

The spawned `sb attach` process prints to stderr and exits non-zero:

```
Error: no terminal identifier provided; terminal name or ID must be specified
```

`WaitReady` then returns `ErrProcessExited: exit status 1`. The example in PR #141 works around this by attaching by `Name` only.

## Root cause

Two places interact:

1. `pkg/spawn/client.go` `buildClientAttachArgs` prefers `--id` over positional when `term.ID != ""`:

   ```go
   switch {
   case term.ID != "":
       args = append(args, "--id", string(term.ID))
   case term.Name != "":
       args = append(args, term.Name)
   }
   ```

   Result: `sb --run-path … attach --socket … --id <id>` (no positional arg).

2. `cmd/sb/attach/attach.go` `RunE` unconditionally errors when positional args are empty:

   ```go
   switch {
   case len(args) == 0:
       return errdefs.ErrNoTerminalIdentifier
   case len(args) == 1:
       if cmd.Flags().Changed("id") { … reject … }
       if cmd.Flags().Changed("name") { … reject … }
   ```

   So `--id` is rejected when passed alongside a positional arg **and** rejected on its own — there is no valid way to attach by `--id` today.

## Suggested fix

Accept `--id` or `--name` as substitutes for the positional arg in `cmd/sb/attach/attach.go`:

```go
case len(args) == 0:
    if !cmd.Flags().Changed("id") && !cmd.Flags().Changed("name") {
        return errdefs.ErrNoTerminalIdentifier
    }
```

The `run()` function below already handles the flag-only path (`terminalIDFlag != ""` branches), so the downstream code does not need changes. Add a CLI test covering `sb attach --id <id> --socket <sock>` and an integration test that exercises `pkg/spawn.NewClient` with an ID-only `TerminalSpec`.

eminwux:
## Summary
- `sb attach --id <id>` (with no positional) returned `ErrNoTerminalIdentifier` and exited 1, making `pkg/spawn.NewClient` unusable with an ID-only `TerminalSpec` — `buildClientAttachArgs` emits `--id` without a positional and the CLI rejected it.
- Relax the args-zero branch to pass validation when either `--id` or `--name` is set; the existing flag-only handling in `run()` already supports this path.
- Extract the inline arg/flag validation into `validateAttachArgs` so the contract is testable in isolation.
- Add unit tests covering the flag-only success paths for both `--id` and `--name`.

## Test plan
- [x] `go test ./cmd/sb/attach/...`
- [x] `go test ./pkg/spawn/...`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make sbsh-sb` + `file ./sbsh` (ELF confirmed)
- [x] `make test` — only pre-existing failures surface (`Test_HandleEvent_EvCmdExited` fixed in unmerged PR #139; `TestSignalForwarder_TargetsProcessGroup` is known-flaky per the shell-signal tests guidance and passes on retry)

Closes #142

### DIFF
--- a/cmd/sb/attach/attach.go
+++ b/cmd/sb/attach/attach.go
@@ -64,25 +64,8 @@ to quickly create a Cobra application.`,
 				return errdefs.ErrLoggerNotFound
 			}
 
-			switch {
-			case len(args) == 0:
-				return errdefs.ErrNoTerminalIdentifier
-			case len(args) == 1:
-				// If user passed -n when listing, reject it
-				if cmd.Flags().Changed("id") {
-					return fmt.Errorf(
-						"%w: the --id flag is not valid when using positional terminal name",
-						errdefs.ErrInvalidFlag,
-					)
-				}
-				if cmd.Flags().Changed("name") {
-					return fmt.Errorf(
-						"%w: the --name flag is not valid when using positional terminal name",
-						errdefs.ErrInvalidFlag,
-					)
-				}
-			case len(args) > 1:
-				return errdefs.ErrTooManyArguments
+			if err := validateAttachArgs(cmd, args); err != nil {
+				return err
 			}
 
 			logger.DebugContext(cmd.Context(), "attach command invoked",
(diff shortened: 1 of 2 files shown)


eminwux:
Re-running failed **Build & Test** job once ([run 24912794536](https://github.com/eminwux/sbsh/actions/runs/24912794536)). Will revisit when it finishes; if it fails again the PR will be marked `ci-stalled`.

eminwux:
**PR #144 Review — `fix(attach): accept --id/--name without positional arg`**

Tight, well-scoped fix. The bug is real — `pkg/spawn/client.go:170`'s `buildClientAttachArgs` emits `--id <ID>` with no positional, and the old `len(args) == 0` branch returned `ErrNoTerminalIdentifier` unconditionally, breaking spawn's `--id`-only callers. The new helper preserves the args==1 contract verbatim and the `run()` socket-resolution path (`case terminalIDFlag != ""`) was already wired for the flag-only case. No sensitive data, no dead symbols, no scope creep.

Existing `Test_ErrNoTerminalIdentifier` still exercises the args==0/no-flags path through `RunE`, so the integration coverage is intact alongside the new unit tests on `validateAttachArgs`. Test plan is fully checked.

LGTM to merge.

PR #147, opened 2026-04-25
feat(terminal): replace --command with positional argv after --

### LINKED ISSUE #146: terminal: replace --command with positional argv after --
## Summary

Today `sbsh terminal` takes the wrapped command as a single string flag:

```
sbsh terminal --socket /run/.../socket --command "/bin/sleep 3600"
```

Replace `--command` with positional argv after a `--` separator, the way `docker run`, `kubectl exec`, `runc exec`, `ctr run`, `sudo`, `env`, `nsenter`, etc. all do:

```
sbsh terminal --socket /run/.../socket -- /bin/sleep 3600
```

Everything after `--` is the workload's `argv` (argv[0] is the binary, the rest are args). If no positional argv is given, fall back to the existing default (`/bin/bash`). The `--command` flag is removed.

## Why

`--command` taking a single string forces every caller that already has a parsed `argv` to flatten it back into a shell-quoted string and then have sbsh's `/bin/sh -c` re-tokenize it. That round-trip is lossy:

- `[\"printf\", \"%s\\n\", \"a b\"]` and other args that contain whitespace, quotes, or shell metacharacters need full POSIX shell-quoting on the way in and would still surprise callers who expected `execve(argv)` semantics.
- It diverges from every other tool in the container/runtime ecosystem. `kuke attach`, `docker exec`, `crictl exec`, `kubectl exec`, `runc exec`, `nsenter`, `sudo` — all of them take positional argv after `--`.
- It blocks straightforward integration from OCI-spec wrappers. `internal/ctr/attachable.go` in eminwux/kukeon (see kukeon#79) wraps a container's `process.args` with sbsh; the workload's argv is already an `[]string`, and the wrapper would just `append([]string{sbsh, \"terminal\", \"--socket\", ..., \"--\"}, original...)`. With `--command`, the wrapper has to either pick a quoting scheme or hand-roll an escaper, and neither survives uncommon argv well.

## Suggested shape

`spf13/cobra` supports this directly: declare the command with `Args: cobra.ArbitraryArgs` and read `cmd.ArgsLenAtDash()` to find the boundary. Roughly:

```go
var cmdTerminal = &cobra.Command{
    Use:   \"terminal [flags] [-- command [args...]]\",
    Short: \"Run a new sbsh terminal\",
    Args:  cobra.ArbitraryArgs,
    RunE: func(cmd *cobra.Command, args []string) error {
        var workload []string
        if dash := cmd.ArgsLenAtDash(); dash >= 0 {
            workload = args[dash:]      // everything after `--`
        }
        // resolution order:
        //   1. positional argv after `--` (if present)
        //   2. default /bin/bash
        ...
    },
}
```

`cmd.ArgsLenAtDash()` returns `-1` if no `--` was given, otherwise the index of the first arg after `--`.

## Breaking change

`--command` is removed. There is no deprecation window — any caller still passing `--command \"...\"` must switch to `-- ...` in the same release. Update README, examples, and any in-tree profiles or scripts that reference the old flag.

## Acceptance criteria

- [ ] `sbsh terminal ... -- <argv...>` runs the workload via `execve` semantics with no shell re-tokenization.
- [ ] `sbsh terminal ...` with no positional argv after `--` still defaults to `/bin/bash`.
- [ ] `--command` flag is removed from `sbsh terminal` (cobra rejects it as an unknown flag).
- [ ] README, `--help` output, and any in-tree examples no longer reference `--command`.
- [ ] Tests cover: positional argv with whitespace/quoting in args, default fallback, and rejection of `--command`.

## Cross-reference

Once this lands, eminwux/kukeon#79 collapses from \"redesign how the wrapper passes argv to sbsh\" to a one-line flag-name correction (`--terminal-socket` → `--socket`, etc.) plus appending `original` to the wrapped argv after `--`. The new e2e test in eminwux/kukeon#75 will be the regression guard on the kukeon side.

eminwux:
## Summary
- `sbsh terminal` now takes the workload as positional argv after a `--` separator (`sbsh terminal --socket ... -- /bin/sleep 3600`), matching how `docker run`, `kubectl exec`, `runc exec`, `ctr run`, `nsenter`, `sudo`, and `env` all wrap their child argv. The workload runs via `execve` semantics with no shell re-tokenization, which preserves args containing whitespace, quotes, or shell metacharacters.
- The `--command` flag is removed (no deprecation window — cobra now rejects it as `unknown flag`). The `SBSH_TERM_COMMAND` viper key was the only consumer of that flag and is dropped along with it.
- Default fallback is unchanged: with no positional argv, the profile default (`/bin/bash`) still applies.
- Help text and `Use` line updated to `terminal [flags] [-- command [args...]]` with examples that reflect the new shape.

## Test plan
- [x] `make sbsh-sb` produces ELF binaries (`sbsh` and the `sb` hardlink).
- [x] `go vet ./...` and `go build ./...` clean.
- [x] `make test` (unit + integration + e2e) green, including new `Test_splitWorkload`, `Test_TerminalCmd_RejectsCommandFlag`, `Test_TerminalCmd_PositionalArgvAfterDash`, `Test_TerminalCmd_DefaultFallback_NoWorkload`.
- [x] Manual smoke: `./sbsh terminal --command /bin/bash` → `Error: unknown flag: --command`. `./sbsh terminal --help` shows `Usage: sbsh terminal [flags] [-- command [args...]]` and the new examples.

Closes #146

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -191,8 +191,6 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_PROFILE = DefineKV("SBSH_TERM_PROFILE", "sbsh.terminal.profile")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
-	SBSH_TERM_COMMAND = DefineKV("SBSH_TERM_COMMAND", "sbsh.terminal.command")
-	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_CAPTURE_FILE = DefineKV("SBSH_TERM_CAPTURE_FILE", "sbsh.terminal.captureFile")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_LOG_FILE = DefineKV("SBSH_TERM_LOG_FILE", "sbsh.terminal.logFile")
--- a/cmd/sbsh/terminal/terminal.go
+++ b/cmd/sbsh/terminal/terminal.go
@@ -54,9 +54,22 @@ func checkFlag(cmd *cobra.Command, flag string, err string) error {
 	return nil
 }
 
+// splitWorkload partitions positional args around a "--" separator. dash is
+// the index of the first arg after "--" (i.e. cobra.Command.ArgsLenAtDash);
+// pass -1 when no "--" was given. Pre-dash args keep their existing meaning
+// (e.g. "-" for stdin); post-dash args are the workload's argv (argv[0] is
+// the binary, the rest are its args), preserved verbatim with no shell
+// re-tokenization.
+func splitWorkload(args []string, dash int) (preDash, workload []string) {
+	if dash < 0 {
+		return args, nil
+	}
+	return args[:dash], args[dash:]
+}
+
 func checkStdInUsage(cmd *cobra.Command, _ []string) error {
 	errorMessage := "positional argument '-'"
-	flagsToCheck := []string{"id", "name", "command", "profile", "log-file", "log-level", "socket", "file"}
+	flagsToCheck := []string{"id", "name", "profile", "log-file", "log-level", "socket", "file"}
 	for _, flag := range flagsToCheck {
 		if err := checkFlag(cmd, flag, errorMessage); err != nil {
 			return err
@@ -67,7 +80,7 @@ func checkStdInUsage(cmd *cobra.Command, _ []string) error {
 
 func checkFileUsage(cmd *cobra.Command, _ []string) error {
 	errorMessage := "the --file flag"
-	flagsToCheck := []string{"id", "name", "command", "profile", "log-file", "log-level", "socket"}
+	flagsToCheck := []string{"id", "name", "profile", "log-file", "log-level", "socket"}
 	for _, flag := range flagsToCheck {
 		if err := checkFlag(cmd, flag, errorMessage); err != nil {
 			return err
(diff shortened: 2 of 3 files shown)


eminwux:
Re-running failed CI jobs once — `Build & Test` (Test workflow) failed. Rerun: https://github.com/eminwux/sbsh/actions/runs/24937385652

Will revisit when checks finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #147 Review — `feat(terminal): replace --command with positional argv after --`**

Clean refactor — the `--command "..."` → `-- argv...` pivot lines up with how `docker run`, `kubectl exec`, `runc exec`, `nsenter`, `sudo`, and `env` all hand off child argv, and the implementation routes through `cobra.ArgsLenAtDash` so post-dash argv is preserved verbatim (no shell re-tokenization). `SBSH_TERM_COMMAND` is fully gone — grep confirms zero remaining references in source. Default fallback to `/bin/bash -i` is preserved via `internal/profile/profile.go:167`. No correctness, sensitive-data, or dead-code concerns.

- **Input validation, edge case** (`cmd/sbsh/terminal/terminal.go:298-311`): when a stdin/file spec is provided alongside positional argv (`sbsh terminal - -- /bin/bash` or `sbsh terminal --file foo.json -- /bin/bash`), `processSpec` short-circuits on the spec and silently drops the workload. The pre-PR equivalent (`--command X -`) errored via `checkStdInUsage`. Defensible either way (spec wins is reasonable), but a one-line check that errors when both a stdin/file spec and a non-empty workload are supplied would preserve the old "incompatible inputs are loud" contract.
- **Doc drift** (`docs/site/cli/sbsh.md:9`, `docs/site/cli/sbsh.md:110-115`): synopsis still reads `sbsh terminal [flags]`. The new `Use:` line (`terminal [flags] [-- command [args...]]`) and the `-- /bin/zsh` example shape aren't reflected here. Help text is correct; only the static doc is slightly stale. Worth a follow-up.

Tests cover the right surface: `Test_splitWorkload` exercises the dash<0, dash==0, and pre-dash-with-extras shapes plus whitespace/quote preservation; `Test_TerminalCmd_RejectsCommandFlag` pins the removal; `Test_TerminalCmd_PositionalArgvAfterDash` and `Test_TerminalCmd_DefaultFallback_NoWorkload` cover the two main runtime paths.

LGTM.

PR #156, opened 2026-04-26
feat: add pkg/attach in-process attach façade for embedders

### LINKED ISSUE #155: pkg/attach: expose in-process attach loop façade for embedders
## Context
External Go programs (notably kukeon's `kuke attach`) want to drive an interactive sbsh attach session from inside their own binary, but today the only path is to spawn the `sb` binary as a subprocess. The interactive attach loop (PTY/copy, detach-keystroke, SIGWINCH, signal forwarding) lives in `internal/client.ClientController`, which is not importable.

## Proposal
Add a public `pkg/attach` package that exposes a single high-level entrypoint, e.g.:

```go
package attach

type Options struct {
    SocketPath              string       // absolute path to terminal control socket (required)
    Stdin                   *os.File     // typically os.Stdin
    Stdout                  *os.File     // typically os.Stdout
    Stderr                  *os.File     // typically os.Stderr
    DisableDetachKeystroke  bool
    Logger                  *slog.Logger // optional
}

func Run(ctx context.Context, opts Options) error
```

`Run` performs the full attach: connects the control socket, runs the bidirectional PTY copy loop, handles detach keystroke / SIGWINCH / SIGINT / SIGTERM, and returns when the session ends. Internals (`ClientController`, `ClientDoc` construction, naming) stay private. `cmd/sb/attach` is refactored to call `pkg/attach.Run` so there is one code path.

## Acceptance criteria
- [ ] `pkg/attach.Run(ctx, Options) error` exists and is the only export an external embedder needs.
- [ ] Feature parity with `sb attach`: detach keystroke (toggleable), SIGWINCH-driven resize, SIGINT/SIGTERM forwarding, clean exit on context cancel.
- [ ] `cmd/sb/attach` calls `pkg/attach.Run` (no duplicate loop).
- [ ] Unit test exercises the façade against a fake control socket (no real PTY required).
- [ ] Godoc on the package describes the embedder contract.

## Notes
Driver: kukeon's `kuke attach` wants to become single-binary by dropping its on-host `sb` dependency. This issue is the sbsh-side enabler; the kukeon-side refactor is filed separately on the kukeon repo.

eminwux:
## Summary
- External Go programs (notably kukeon's `kuke attach`) need to drive an interactive sbsh attach session in-process without spawning the `sb` binary as a subprocess. Today the attach loop lives in `internal/client.ClientController`, which is not importable.
- Add `pkg/attach.Run(ctx, Options) error` as the lone public entrypoint for embedders. Internals (`ClientController`, `ClientDoc` construction, naming, run-dir housekeeping) stay private behind the package.
- Wire `cmd/sb/attach` to call `pkg/attach.Run` so there is one code path for the interactive attach loop. The CLI keeps ownership of `signal.NotifyContext(SIGINT/SIGTERM)`; pkg/attach itself deliberately does not trap signals (godoc'd) so embedders pick their own policy.
- Plumb caller-supplied stdin/stdout/stderr `*os.File` handles through `internal/client.NewClientControllerWithIO` → `clientrunner.NewClientRunnerExecWithIO` → `Exec.{stdin,stdout,stderr}`. The `os.Std*` references in `io.go`, `lifecycle.go`, `terminal.go` are replaced with the runner-scoped fields; existing callers go through the nil-defaulting overload, so behaviour is unchanged for `sb`.
- Add a SocketFile-only short-circuit in `Controller.createAttachTerminal`: embedders that already know the absolute control-socket path skip the metadata-discovery dance.

## Acceptance criteria
- [x] `pkg/attach.Run(ctx, Options) error` exists and is the only export an external embedder needs.
- [x] Feature parity with `sb attach`: detach keystroke (toggleable via `Options.DisableDetachKeystroke`), SIGWINCH-driven resize (installed by `forwardResize` against `Options.Stdin`), SIGINT/SIGTERM handled by the caller's ctx (CLI wires `signal.NotifyContext`), clean exit on context cancel (returns `errdefs.ErrContextDone`).
- [x] `cmd/sb/attach` calls `pkg/attach.Run` (no duplicate loop; `buildClientDoc` and the inline controller wiring are gone).
- [x] Unit test exercises the façade against a fake control socket (`startFakeTerminalServer` runs a JSON-RPC server with a `fakeTerminalController`; no real PTY required).
- [x] Godoc on the package describes the embedder contract (`doc.go` covers Run, Options.Stdin TTY requirement, signal-policy ownership, socket lifetime, pre-v1 stability tag).

## Test plan
- [x] `make sbsh-sb` — produces ELF executables for `sbsh` and `sb`.
- [x] `file ./sbsh ./sb` — both `ELF 64-bit LSB executable`.
- [x] `go vet ./...` — clean.
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full unit suite green (matches CI's `Build & Test` step), including new `pkg/attach` tests.
- [x] `go test -tags=integration ./cmd/sb/get/...` — green (matches CI's integration step).
- [x] `(cd docs/examples/library-consumer && go build ./... && go vet ./...)` — green (matches CI's library-consumer build step).
- [ ] `-race` not run locally (no gcc on this host); CI covers it.

Closes #155

### DIFF
--- a/cmd/sb/attach/attach.go
+++ b/cmd/sb/attach/attach.go
@@ -23,18 +23,15 @@ import (
 	"log/slog"
 	"os"
 	"os/signal"
-	"path/filepath"
 	"syscall"
 	"time"
 
 	"github.com/eminwux/sbsh/cmd/config"
 	"github.com/eminwux/sbsh/cmd/sb/get"
 	"github.com/eminwux/sbsh/cmd/types"
-	"github.com/eminwux/sbsh/internal/client"
 	"github.com/eminwux/sbsh/internal/defaults"
 	"github.com/eminwux/sbsh/internal/errdefs"
-	"github.com/eminwux/sbsh/internal/naming"
-	"github.com/eminwux/sbsh/pkg/api"
+	"github.com/eminwux/sbsh/pkg/attach"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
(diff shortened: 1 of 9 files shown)


eminwux:
**PR #156 Review — `feat: add pkg/attach in-process attach façade for embedders`**

Clean façade extraction: CLI-only concerns (flag→socket resolution, signal trapping, exit-code handling) stay in `cmd/sb/attach`, while `pkg/attach.Run` takes only an absolute socket path. IO-handle plumbing is done with nil-defaulting overloads (`NewClientControllerWithIO`, `NewClientRunnerExecWithIO`) so no existing caller breaks. The new `SocketFile` branch in `Controller.Run`/`createAttachTerminal` is gated correctly. Test coverage exercises the façade end-to-end against a fake JSON-RPC terminal server with no real PTY required. CI is green; no correctness bugs, no sensitive-data leaks, no dead code, no convention violations.

A few non-blocking items:

- **CLI behavior delta** (`cmd/sb/attach/attach.go:209`): the old `run()` returned `errdefs.ErrContextDone` from the ctx-done branch; the new code swallows it (`if … errors.Is(runErr, errdefs.ErrContextDone) { return nil }`), so cobra no longer prints an `Error: context has been cancelled` line on Ctrl-C. Probably an intentional UX win, but worth a sentence in the PR description so it doesn't surprise anyone bisecting later.
- **Empty terminal ID in SocketFile-only path** (`internal/client/controller.go:289-299`): when called via `pkg/attach.Run`, `doc.Spec.TerminalSpec.ID` is `""` and that empty string ends up as the map key in `terminalstore.Add` and in event/log fields (`io.go:89,129,255-265`). No functional bug since pkg/attach only adds one terminal per Controller, but consider populating a synthetic `naming.RandomID()` in the SocketFile branch so logs/`m.current` aren't keyed by `""`.
- **Stale inline comment** (`internal/client/controller_test.go:1059`): says `// Both empty should trigger ErrAttachNoTerminalSpec` — with the new validation it's all three of ID/Name/SocketFile.
- **Non-sentinel validation error** (`pkg/attach/attach.go:600`): the `SocketPath == ""` check returns a bare `errors.New("pkg/attach: …")`. The package godoc tells callers they can `errors.Is` against `errdefs.*` sentinels — this one isn't matchable. Either export an `ErrSocketPathRequired` from `pkg/attach` or note in the doc that programmer-error validations are unwrapped.
- **`docs/examples/library-consumer` doesn't exercise the new façade**: the PR's test plan checks it builds, but the example doesn't call `pkg/attach.Run`, so an embedder reading it for guidance won't see the new attach path. Worth a follow-up to add an attach-aware example to close the loop with the kukeon driver named in #155.

LGTM to proceed — items above are all non-blocking.

PR #161, opened 2026-04-26
fix(client): wait on rpcReadyCh with ctx.Done so Run can't pin on RPC startup

### LINKED ISSUE #149: test: Test_ErrContextDone client mock drops rpcReadyCh signal under -race scheduling
## Repro

Observed once on `fix/controller-close-deadlock-138` (the PR #148 branch) with `gcc 14.2.0`:

```bash
CGO_ENABLED=1 go test -timeout=300s -race ./internal/terminal/ ./internal/client/...
```

The full `internal/client` package timed out at 300s. The same test run a second time passed; `-count=20` of `Test_ErrContextDone` in isolation also passed. Reproduction is timing-dependent and load-sensitive, not deterministic.

## Observed

Goroutine trace at the timeout:

```
goroutine 42 [chan receive]:
github.com/eminwux/sbsh/internal/client.Test_ErrContextDone
        internal/client/controller_test.go:373    <-- waiting on <-exitCh

goroutine 43 [chan receive]:
github.com/eminwux/sbsh/internal/client.(*Controller).Run
        internal/client/controller.go:130 +0x12e5  <-- blocked on <-s.rpcReadyCh
        internal/client/controller_test.go:366
```

`Run` is sitting on the unbuffered `<-s.rpcReadyCh` receive at `controller.go:130`, so `Run` never returns and the test never exits.

## Root cause

The `StartServerFunc` mock used by `Test_ErrContextDone` (and several sibling tests) signals readiness with a non-blocking send (`internal/client/controller_test.go:291-296`):

```go
StartServerFunc: func(_ context.Context, _ *clientrpc.ClientControllerRPC, readyCh chan error, _ chan error) {
    select {
    case readyCh <- nil:
    default:
    }
},
```

`StartServer` runs in a goroutine spawned just before `<-s.rpcReadyCh` (`controller.go:128-130`). If that goroutine is scheduled and runs the `default` branch *before* `Run` reaches the receive — possible under `-race` because the runtime randomizes goroutine scheduling more aggressively — the ready signal is silently dropped and `Run` blocks forever. The test's `cancel()` doesn't help because the receive isn't a `select` over `s.ctx.Done()`.

The same mock pattern is used by other client tests (e.g. `Test_ErrRPCServerExited` shape at `controller_test.go:389-394`) and is latent there too.

## Suggested fix

Either:
- Make the mock's send blocking (`readyCh <- nil`) — readers of `rpcReadyCh` either receive or the surrounding test will time out cleanly via the test goroutine, not via a forever-block.
- Or harden production `Run` to wait on `rpcReadyCh` inside a `select` that also watches `s.ctx.Done()`, so context cancellation can break the hang regardless of mock behavior. This second option is the more robust fix because it removes a real-world cliff: if production `StartServer` ever crashes before sending on `readyCh`, today's `Run` hangs the same way.

eminwux:
## Summary
- `client.Controller.Run` waited on `<-s.rpcReadyCh` outside any select, so a `StartServer` that crashed before its send — or a test mock that hit its non-blocking-send `default` branch — would pin Run forever, even after the parent context cancelled. Wrap the receive in a select that also watches `s.ctx.Done()` so context cancellation always unblocks startup.
- Buffer `rpcReadyCh` (cap 1). Two reasons: (a) the production `StartServer.readyCh <- nil` send no longer blocks forever if Run took the new ctx.Done branch (no goroutine leak); (b) the existing tests' `select { case readyCh <- nil: default: }` mock can't drop the signal anymore — the buffered slot is always available. This eliminates the race surface called out in #149.
- Add `Test_ErrContextDone_RPCNeverReady` regression: a `StartServerFunc` that never signals readiness must still let `Run` return with `ErrContextDone` when the parent context is cancelled. Without the new select arm, this hangs until the test's 2s deadline trips.

## Test plan
- [x] `go test -timeout=120s -count=200 -run '^Test_ErrContextDone(_RPCNeverReady)?$|^Test_ErrRPCServerExited$|^Test_ErrCloseReq$' ./internal/client/` — 200/200 green
- [x] `go test -timeout=180s ./internal/client/...` — full client suite green
- [x] `go test -timeout=5m -skip 'Test_HandleEvent_EvCmdExited' $(go list ./... | grep -v '/e2e$')` — full unit suite green
- [x] `make sbsh-sb` — `./sbsh` is `ELF 64-bit LSB executable`
- [x] `go vet ./...` — clean

`Test_HandleEvent_EvCmdExited` (`internal/terminal`) is the pre-existing #138 deadlock — verified to fail 5/5 on `origin/main` (`468aabb`) in an isolated worktree, unrelated to this PR's diff. Open PRs #148 and #139 fix it.

Closes #149

### DIFF
--- a/internal/client/controller.go
+++ b/internal/client/controller.go
@@ -85,8 +85,12 @@ func NewClientControllerWithIO(
 		closedCh:    make(chan struct{}),
 		closingCh:   make(chan error, 1),
 		ctrlReadyCh: make(chan struct{}),
-		rpcReadyCh:  make(chan error),
-		rpcDoneCh:   make(chan error),
+		// rpcReadyCh is buffered (cap 1) so StartServer can deposit its
+		// one-shot ready signal without blocking on a receiver. Run watches
+		// both this channel and ctx.Done below; without the buffer, a Run
+		// that bailed on ctx.Done would pin StartServer forever on its send.
+		rpcReadyCh: make(chan error, 1),
+		rpcDoneCh:  make(chan error),
 		closeReqCh:  make(chan error, 1),
 		//nolint:mnd // event channel buffer size
 		eventsCh: make(chan clientrunner.Event, 32),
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #161 Review — `fix(client): wait on rpcReadyCh with ctx.Done so Run can't pin on RPC startup`**

Targeted fix for #149: the `select` over `rpcReadyCh`/`ctx.Done()` plus the cap-1 buffer correctly closes both the Run-pin-on-startup hang and the StartServer-send-leak that would follow it. The new `Test_ErrContextDone_RPCNeverReady` exercises exactly the failure mode (StartServer never signals, parent ctx cancelled) and would hang on the unfixed code, so the regression has real teeth. No sensitive-data concerns. Diff is small and focused.

One small non-blocking nit:

- **Error rendering** (`internal/client/controller.go:257`): in the new `<-s.ctx.Done()` branch, when `s.Close` succeeds (the common case for plain ctx cancellation), `errDone` stays `nil` and `fmt.Errorf("%w: %w", errdefs.ErrContextDone, errDone)` renders as `"context has been cancelled: %!w(<nil>)"`. `errors.Is(err, errdefs.ErrContextDone)` still works (so the new test passes), but the user-facing message is malformed. Consider wrapping `s.ctx.Err()` instead — it's the actual cause and is always non-nil here, so the message stays well-formed and the deadline-vs-cancel distinction propagates to callers. This same pattern already exists at `internal/terminal/controller.go:200`, so it's a pre-existing style rather than a regression introduced here.

LGTM.

Issue #108, opened 2026-04-20
test: `Test_HandleEvent_EvCmdExited` hangs


eminwux:
`Test_HandleEvent_EvCmdExited` in \`internal/terminal/controller_test.go\` hangs when run as part of the full test suite, forcing contributors to use \`go test ./... -skip Test_HandleEvent_EvCmdExited\` to get a clean run (see PR #106 CI workaround).

Acknowledged as unrelated to the `sb stop terminal` work in PR #106; splitting out so it's tracked independently.

**Reproduction**

\`\`\`
go test ./internal/terminal/ -run Test_HandleEvent_EvCmdExited -timeout 30s
\`\`\`

**Expected:** test completes (pass or fail) within seconds.
**Actual:** test does not return; suite times out.

**Suspected area:** event-loop / channel shutdown in the `EvCmdExited` handler path — likely a missing close or unbuffered send after controller teardown. Needs investigation.

eminwux:
Fixed by PR #185 (commit bce0bd2 — atomic shutdown gate in Controller.Close). The deadlock was filed in more detail as #138 and resolved there. Closing as completed.

PR #163, opened 2026-04-26
fix(controller): wrap ctx.Err() in ErrContextDone return so message isn't %!w(<nil>)

### LINKED ISSUE #162: client/terminal controller: ErrContextDone wrap renders `%!w(<nil>)` when Close succeeds
Surfaced during review of #161.

In `internal/client/controller.go:257`:

```go
return fmt.Errorf("%w: %w", errdefs.ErrContextDone, errDone)
```

When `s.Close(s.ctx.Err())` succeeds (the common case for plain ctx cancellation), `errDone` stays `nil` and the resulting error string renders as `"context has been cancelled: %!w(<nil>)"`. `errors.Is(err, errdefs.ErrContextDone)` still returns true (so tests pass), but the user-facing message is malformed.

Same shape exists at `internal/terminal/controller.go:200`:

```go
return fmt.Errorf("%w: %w", errdefs.ErrContextDone, err)
```

where `err` is also nil when Close succeeded.

Suggested fix in both places: wrap `ctx.Err()` (always non-nil here, distinguishes Canceled vs DeadlineExceeded) and only fold in the close error when one actually occurred. Sketch:

```go
if errC := s.Close(s.ctx.Err()); errC != nil {
    return fmt.Errorf("%w: %w: %w", errdefs.ErrContextDone, s.ctx.Err(), errC)
}
return fmt.Errorf("%w: %w", errdefs.ErrContextDone, s.ctx.Err())
```

eminwux:
## Summary

- The three `case <-ctx.Done()` branches in `internal/client/controller.go` and `internal/terminal/controller.go` wrapped a `var err error` (nil when `Close` succeeded) into the returned error, so the message rendered as `"context has been cancelled: %!w(<nil>)"`. `errors.Is(err, errdefs.ErrContextDone)` still matched, so the bug was cosmetic but tests using `Errorf` formatting saw the malformed string (visible in `Test_HandleEvent_EvCmdExited`'s failure output during occasional flakes).
- Wrap `ctx.Err()` (always non-nil here, distinguishes `context.Canceled` from `context.DeadlineExceeded`) instead of the never-set local. When `Close` returns an error, fold it in as a third wrapped error so the cause stays in the chain. The intermediate `errdefs.ErrOnClose` sentinel is dropped from this path — no test or production caller does `errors.Is(..., ErrOnClose)`, so this is a deliberate simplification rather than a behavior break.
- Issue called out two sites; the same pattern exists at `internal/client/controller.go:171` (the RPC-readiness select arm) — fixed there too with the equivalent shape that preserves the original `close(s.ctrlReadyCh)` ordering.

## Test plan

- [x] `make sbsh-sb` (ELF binary verified)
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test -timeout=5m -skip Test_HandleEvent_EvCmdExited $(go list ./... | grep -v '/e2e$')` — all green
- [x] `Test_HandleEvent_EvCmdExited` skipped: pre-existing flaky deadlock tracked by #138 / #108 with open fixes #139 and #148. Reproduced on `origin/main` (~40% deadlock over 5 runs) — unrelated to this PR's scope.

Closes #162

### DIFF
--- a/internal/client/controller.go
+++ b/internal/client/controller.go
@@ -161,14 +161,15 @@ func (s *Controller) Run(doc *api.ClientDoc) error {
 			return fmt.Errorf("%w: %w", errdefs.ErrStartRPCServer, errRPC)
 		}
 	case <-s.ctx.Done():
-		var errDone error
+		ctxErr := s.ctx.Err()
+		errReturn := fmt.Errorf("%w: %w", errdefs.ErrContextDone, ctxErr)
 		s.logger.Warn("parent context canceled while waiting for RPC server start")
-		if errC := s.Close(s.ctx.Err()); errC != nil {
+		if errC := s.Close(ctxErr); errC != nil {
 			s.logger.Error("error during Close after context done", "error", errC)
-			errDone = fmt.Errorf("%w: %w", errdefs.ErrOnClose, errC)
+			errReturn = fmt.Errorf("%w: %w: %w", errdefs.ErrContextDone, ctxErr, errC)
 		}
 		close(s.ctrlReadyCh)
-		return fmt.Errorf("%w: %w", errdefs.ErrContextDone, errDone)
+		return errReturn
 	}
 
 	var terminal *api.AttachedTerminal
@@ -266,13 +267,13 @@ func (s *Controller) Run(doc *api.ClientDoc) error {
 	for {
 		select {
 		case <-s.ctx.Done():
-			var errDone error
+			ctxErr := s.ctx.Err()
 			s.logger.Warn("parent context canceled, shutting down controller")
-			if errC := s.Close(s.ctx.Err()); errC != nil {
+			if errC := s.Close(ctxErr); errC != nil {
 				s.logger.Error("error during Close after context done", "error", errC)
-				errDone = fmt.Errorf("%w: %w", errdefs.ErrOnClose, errC)
+				return fmt.Errorf("%w: %w: %w", errdefs.ErrContextDone, ctxErr, errC)
 			}
-			return fmt.Errorf("%w: %w", errdefs.ErrContextDone, errDone)
+			return fmt.Errorf("%w: %w", errdefs.ErrContextDone, ctxErr)
 
 		case ev := <-s.eventsCh:
 			s.logger.Debug(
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #163 Review — `fix(controller): wrap ctx.Err() in ErrContextDone return so message isn't %!w(<nil>)`**

Fix is correct and tightly scoped to the three `<-ctx.Done()` arms in `internal/client/controller.go` and `internal/terminal/controller.go`. Wrapping `ctx.Err()` instead of the never-set `errDone`/`err` resolves the `"%!w(<nil>)"` rendering, and folding `errC` in as a third wrapped error when `Close` fails keeps the close cause in the chain. No sensitive-data exposure surface and the error wrappings are well-formed. Bonus that `errors.Is(err, context.Canceled)` and `errors.Is(err, context.DeadlineExceeded)` now work on these returns.

A few observations, none blocking:

- **`errdefs.ErrOnClose` drop verified** (`internal/errdefs/errdefs.go:34`): grepped the tree — no `errors.Is(..., ErrOnClose)` callers in production or tests, so dropping it from these three paths is purely a string-shape change. The sentinel is still wrapped at ~15 other sites in `internal/client/controller.go` / `internal/terminal/controller.go`, so it stays load-bearing for the rest of the package and isn't a candidate for removal here.
- **Test compatibility preserved**: `internal/terminal/controller_test.go:322` and `internal/client/controller_test.go:279,381` only assert `errors.Is(err, errdefs.ErrContextDone)` — still matches under the new wrap shape. `cmd/sbsh/sbsh_test.go:170,313` use `errdefs.ErrOnClose` only as an arbitrary `WaitCloseFunc` return value and assert on `ErrWaitOnClose`, not `ErrOnClose`, so those are unaffected too.
- **`ctxErr` caching is fine**: `s.ctx.Err()` is stable and non-nil after `<-s.ctx.Done()` returns, so the single read into `ctxErr` is safe.

Flaky-deadlock skip on `Test_HandleEvent_EvCmdExited` is reasonable given #138/#108 are tracked separately and you reproduced on `origin/main` — not a gate for this PR.

LGTM to proceed.

PR #164, opened 2026-04-26
fix(terminal): reject positional argv after -- when stdin/file spec supplied

### LINKED ISSUE #150: terminal: error when stdin/file spec is supplied alongside positional argv
Surfaced during review of #147 (non-blocking).

`cmd/sbsh/terminal/terminal.go:298-311`: when a stdin/file spec is provided alongside positional argv (e.g. `sbsh terminal - -- /bin/bash` or `sbsh terminal --file foo.json -- /bin/bash`), `processSpec` short-circuits on the spec and silently drops the workload. The pre-PR equivalent (`--command X -`) errored via `checkStdInUsage`.

Suggested fix: add a one-line check that errors when both a stdin/file spec and a non-empty workload are supplied, preserving the old "incompatible inputs are loud" contract.

eminwux:
## Summary

- `cmd/sbsh/terminal/terminal.go`'s `processSpec` short-circuited on a non-nil spec from stdin/`--file` and ignored `workload`, so `sbsh terminal - -- /bin/bash` and `sbsh terminal --file foo.json -- /bin/bash` silently dropped the positional argv.
- Added a guard at the top of the `*spec != nil` branch that returns `errdefs.ErrInvalidArgument` when a workload accompanies a stdin/file spec, restoring the pre-#147 "incompatible inputs are loud" contract that `checkStdInUsage` used to enforce.
- The existing comment "Spec provided via stdin" is technically stale (spec can also come from `--file`), but updating it is out of scope for this fix.

## Test plan

- [x] `make sbsh-sb` (ELF binary verified)
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./cmd/sbsh/terminal/...` — green, including new `Test_processSpec_RejectsWorkloadWithSpec`
- [x] `go test -timeout=5m -skip Test_HandleEvent_EvCmdExited \$(go list ./... | grep -v '/e2e\$')` — green (skip per known #138 deadlock)
- [x] Manual: `./sbsh terminal --file /tmp/spec.json -- /bin/zsh` and `echo '{...}' | ./sbsh terminal - -- /bin/zsh` both error with `invalid positional argument: positional argv after '--' is not valid when a terminal spec is supplied via stdin or --file`
- [ ] **Pre-existing flake unrelated to this PR:** `TestSignalForwarder_TargetsProcessGroup` in `internal/terminal/terminalrunner` flakes on \`origin/main\` (~80% over 20 runs in an isolated worktree); reproduces independently of this branch's diff. Will surface separately via \`/reflect\`.

Closes #150

### DIFF
--- a/cmd/sbsh/terminal/terminal.go
+++ b/cmd/sbsh/terminal/terminal.go
@@ -215,6 +215,14 @@ func setLoggingVarsFromFlags() {
 func processSpec(cmd *cobra.Command, spec **api.TerminalSpec, workload []string) error {
 	// Check if spec is already provided
 	if *spec != nil {
+		// A stdin/file spec already supplies the workload; positional argv
+		// after `--` would be silently dropped, so reject it loudly.
+		if len(workload) > 0 {
+			return fmt.Errorf(
+				"%w: positional argv after '--' is not valid when a terminal spec is supplied via stdin or --file",
+				errdefs.ErrInvalidArgument,
+			)
+		}
 		// Spec provided via stdin
 		err := logging.SetupFileLogger(cmd, (*spec).LogFile, (*spec).LogLevel)
 		if err != nil {
--- a/cmd/sbsh/terminal/terminal_test.go
+++ b/cmd/sbsh/terminal/terminal_test.go
@@ -491,3 +491,23 @@ func Test_ValidJSONSpec(t *testing.T) {
 		t.Fatalf("expected Name %s; got: %s", validSpec.Name, spec.Name)
 	}
 }
+
+func Test_processSpec_RejectsWorkloadWithSpec(t *testing.T) {
+	cmd := &cobra.Command{}
+	cmd.SetContext(context.Background())
+
+	spec := &api.TerminalSpec{
+		ID:      api.ID("test-id"),
+		Kind:    api.TerminalLocal,
+		Name:    "test-name",
+		Command: "/bin/bash",
+	}
+
+	err := processSpec(cmd, &spec, []string{"/bin/zsh"})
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #164 Review — `fix(terminal): reject positional argv after -- when stdin/file spec supplied`**

Tight, well-scoped fix. The guard at `cmd/sbsh/terminal/terminal.go:218-225` lands in the right place — at the very top of the `*spec != nil` branch, before any logger setup, so the error surfaces before the side effect of opening a log file. Wrapping `errdefs.ErrInvalidArgument` keeps the sentinel discoverable via `errors.Is` and matches the wrap pattern already used at `cmd/sbsh/terminal/terminal.go:157`. No sensitive-data exposure surface — error string only echoes the static guidance.

A few small observations, none blocking:

- **Test coverage** (`cmd/sbsh/terminal/terminal_test.go:494-513`): `Test_processSpec_RejectsWorkloadWithSpec` exercises the guard directly via `processSpec`, which is the right unit. Optional belt-and-suspenders: an end-to-end via `processInput` to confirm the error propagates through the stdin/file load path, but the existing test is sufficient since `processSpec` is a small narrow seam.
- **Stale comment on line 219** (`// Spec provided via stdin`): you flagged this in the PR description and chose to defer — agreed it's out of scope for the bugfix. Will file as `followup-nits` per the review workflow.
- **Test plan checkbox** (the unchecked `TestSignalForwarder_TargetsProcessGroup` flake): pre-existing on `origin/main` per your repro and unrelated to this diff — not a gate.

LGTM to proceed.

PR #167, opened 2026-04-27
test: fix TestSignalForwarder_TargetsProcessGroup sibling-trap race

### LINKED ISSUE #165: test: TestSignalForwarder_TargetsProcessGroup flakes on sibling-trap race
## Repro

Reproduced on `origin/main` at `fe5c44d` in an isolated worktree (no local diff):

```bash
git worktree add /tmp/sbsh-main-check origin/main
cd /tmp/sbsh-main-check
go test -timeout=60s -count=20 -run '^TestSignalForwarder_TargetsProcessGroup$' ./internal/terminal/terminalrunner/...
```

Result: 4/20 runs failed with `sibling did not receive SIGHUP (marker /tmp/.../sibling absent)`. The first run (the implicit `-count=1` invocation that surfaced this during a `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` sweep) also failed.

## Observed

```
--- FAIL: TestSignalForwarder_TargetsProcessGroup (0.01s)
    signals_linux_test.go:96: sibling did not receive SIGHUP (marker /tmp/TestSignalForwarder_TargetsProcessGroup26595257/001/sibling absent)
```

## Root cause

`internal/terminal/terminalrunner/signals_linux_test.go:50-58` has a TOCTOU between the leader's `echo ready` and the sibling subshell installing its HUP trap:

```sh
trap 'touch "$LEADER"' HUP
(trap 'touch "$SIBLING"; exit 0' HUP; sleep 10 & wait) &
child=$!
echo ready
wait $child
```

`(...) &` only guarantees the subshell has been forked, not that it has executed its first statement. When the test reads `ready` and then calls `syscall.Kill(os.Getpid(), syscall.SIGHUP)`, SIGHUP can be delivered to the pgroup before the subshell has installed its `trap '... HUP'` handler — in that window, the subshell takes the default action for SIGHUP (terminate without running the trap), so `$SIBLING` is never touched. The leader trap is unaffected because it is installed before `echo ready`.

## Suggested fix

Synchronise on the inner trap, not the outer shell's `echo ready`. Two cheap options:

1. Have the subshell signal readiness itself (e.g. a `touch "$SIBLING_READY"` immediately after `trap`, with the leader spinning on its presence before printing `ready`).
2. Replace the `&`/`wait` pattern with a fifo or a shared file the subshell touches *after* installing the trap, and have the leader block on it.

Either makes the readiness signal causally downstream of the trap install, removing the race.

(Surfaced while running the full unit suite for PR #164 / issue #150. Not in #164's scope — fix is independent.)

eminwux:
## Summary

`TestSignalForwarder_TargetsProcessGroup` flakes (~20% on `origin/main`) because the shell script the test fans signals through has *three* races, not the one called out in #165. This patch closes all three:

1. **Subshell trap-install race (the one identified in #165).** `(trap '…' HUP; sleep 10 & wait) &` only guarantees the subshell has been forked, not that it has executed its `trap`. SIGHUP arriving before the trap is installed takes the default action (terminate) and the marker is never touched. The subshell now `touch "$SIBLING_READY"` *after* installing its trap; the leader spins on that marker before printing `ready`, so `syscall.Kill(SIGHUP)` only ever lands once the trap is in place.
2. **Subshell `wait` premature-exit race.** With `sleep 10 & wait`, when SIGHUP reaches the `sleep` child first, dash reaps it via `waitpid` and exits the subshell *before* the pending SIGHUP for the subshell itself dispatches the trap, so `$SIBLING` is never touched. Replaced with `while :; do sleep 1; done` so the subshell stays alive across sleep deaths and the trap's `exit 0` is the only termination path.
3. **Leader `wait` early-return race.** `wait $child` in dash returns when the *leader's own* HUP trap fires — not when `$child` exits. The leader was therefore printing `done` while the subshell's `touch "$SIBLING"` was still in flight; the Go-side marker check then raced an in-flight write. Wrapped in a `while kill -0 $child …; do wait $child; done` loop so `done` is pinned to subshell death.

The third race is what surfaced when patching only #1 and #2 — the failure rate actually went *up* with the loop subshell because it widened the trap-completion window, and the early-`done` race dominated. Confirmed by an isolated diagnostic harness (run 200×) showing `trap_fired=true sibling=false` ≈30% of the time despite the trap entering and writing its first marker.

## Test plan

- [x] `go test -timeout=300s -count=500 -run '^TestSignalForwarder_TargetsProcessGroup\$' ./internal/terminal/terminalrunner/...` — 0 failures over 500 runs (was ~20% on `origin/main`)
- [x] `go test ./internal/terminal/terminalrunner/...` — green
- [x] `go test -timeout=5m -skip Test_HandleEvent_EvCmdExited \$(go list ./... | grep -v '/e2e\$')` — green (skip per known #138 deadlock; addressed by #148)
- [x] `make sbsh-sb` — `./sbsh` and `./sb` both ELF executables
- [x] `go vet ./...` — clean

Closes #165

### DIFF
--- a/internal/terminal/terminalrunner/signals_linux_test.go
+++ b/internal/terminal/terminalrunner/signals_linux_test.go
@@ -41,23 +41,49 @@ func TestSignalForwarder_TargetsProcessGroup(t *testing.T) {
 	dir := t.TempDir()
 	leaderMarker := filepath.Join(dir, "leader")
 	siblingMarker := filepath.Join(dir, "sibling")
+	siblingReadyMarker := filepath.Join(dir, "sibling_ready")
 
-	// Subshell uses "sleep & wait" so the shell does not tail-call-exec
-	// sleep (dash optimizes "exec sleep" when sleep is the last command,
-	// which would leave no shell alive to run the SIGHUP trap).
-	// Markers are written to disk so the test does not depend on stdout
-	// flush ordering of the subshell vs. the leader.
+	// Three constraints must hold to deliver a deterministic HUP-via-trap
+	// signal under both the leader AND the subshell:
+	//
+	//  1. The subshell's HUP trap must be installed before the test sends
+	//     SIGHUP. `(...) &` only guarantees the subshell has been forked, not
+	//     that it has executed its first statement. The leader therefore polls
+	//     "$SIBLING_READY", which the subshell touches *after* installing the
+	//     trap, before printing "ready".
+	//  2. The subshell must stay alive until its trap actually runs. A bare
+	//     `sleep 10 & wait` is racy: when SIGHUP reaches the sleep child first,
+	//     dash reaps it via waitpid and exits the subshell before the pending
+	//     SIGHUP for the subshell itself dispatches the trap, so $SIBLING is
+	//     never touched. A `while :; do sleep 1; done` loop keeps the shell
+	//     alive across sleep deaths until the trap's `exit 0` terminates it.
+	//  3. The leader must not print "done" until the subshell's trap has
+	//     actually finished (i.e. the subshell process is dead). `wait $child`
+	//     returns when *the leader's own* HUP trap fires, not when $child
+	//     exits. Re-waiting in a `kill -0`-gated loop pins "done" to subshell
+	//     death, ensuring the marker check in the Go side never races with an
+	//     in-flight `touch "$SIBLING"`.
+	//
+	// Markers are written to disk so the test does not depend on stdout flush
+	// ordering of the subshell vs. the leader.
 	script := `
 trap 'touch "$LEADER"' HUP
-(trap 'touch "$SIBLING"; exit 0' HUP; sleep 10 & wait) &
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #167 Review — `test: fix TestSignalForwarder_TargetsProcessGroup sibling-trap race`**

Solid analysis — the three races are real and each fix attacks the right primitive (readiness handshake, sleep-loop liveness, kill-0-gated rewait). Diff is contained to one test file, CI is green, and the 500-run reproduction check is exactly the right validation. No correctness, security, or convention concerns.

A couple of small non-blocking notes:

- **Style nit** (`signals_linux_test.go:71`): `while kill -0 $child 2>/dev/null; do wait $child 2>/dev/null; done` is correct, but the inner `2>/dev/null` is redundant — by the time `wait` would emit its "not a child" stderr, `kill -0` has already failed and the loop has exited. Can drop it; not worth a respin.
- **Style nit** (`signals_linux_test.go:54`): `sleep 0.01` is a GNU coreutils extension. Fine here under `//go:build linux`, but worth a one-liner comment if/when this script gets cargo-culted into a non-Linux test.

LGTM to proceed.

PR #173, opened 2026-04-27
fix(client): synthesize terminal ID for SocketFile-only attach

### LINKED ISSUE #157: pkg/attach: populate synthetic terminal ID for SocketFile-only attach path

When `pkg/attach.Run` is invoked, `Controller.createAttachTerminal` (`internal/client/controller.go:289-299`) takes the SocketFile-only short-circuit and copies `doc.Spec.TerminalSpec` into a fresh `*api.AttachedTerminal`. Embedders only know the absolute control socket path, so `Spec.ID` ends up as the empty string `""`.

That empty ID then flows into:

- `terminalstore.Add(terminal)` — used as the map key (`internal/client/terminalstore/session_store.go`); first-call also sets `m.current` to `api.ID("")`.
- Event records in `internal/client/clientrunner/io.go:89,129`.
- Log fields in `internal/client/clientrunner/io.go:255-265`.

No functional bug today because each `pkg/attach.Run` call gets its own Controller and adds exactly one terminal — the empty key never collides. But the empty-ID propagation is unsatisfying in logs/event payloads emitted by embedded clients.

Suggested fix: in the SocketFile-only branch of `createAttachTerminal`, if `spec.ID == ""`, fill in a synthetic `naming.RandomID()` before calling `terminalstore.Add`. Keeps logs/`m.current` keyed by something distinct per session and stays internal (the embedder never sees this ID).

eminwux:
## Summary
- `internal/client.Controller.createAttachTerminal` now fills in a synthetic `naming.RandomID()` when the SocketFile-only short-circuit runs without a caller-provided `Spec.ID`. Embedders that hand `pkg/attach` only the absolute control-socket path no longer end up with an empty-string ID flowing into the `terminalstore` map key, `m.current`, event records (`internal/client/clientrunner/io.go`), and log fields.
- The synthetic ID lives on a local copy of the spec — the embedder's `doc.Spec.TerminalSpec` is not mutated. A caller-supplied ID still takes precedence (only the `spec.ID == ""` branch synthesizes).
- No functional bug today (each `pkg/attach.Run` gets its own Controller and adds exactly one terminal, so empty keys don't collide), but logs/events stay coherent across embedded sessions.

## Test plan
- [x] `go test -timeout=60s -run 'Test_CreateAttachTerminal' -v ./internal/client/...` — both new tests pass (synthesizes when empty; preserves when provided)
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full unit suite green
- [x] `docs/examples/library-consumer`: `go build ./... && go vet ./...` — clean

Closes #157

### DIFF
--- a/internal/client/controller.go
+++ b/internal/client/controller.go
@@ -30,6 +30,7 @@ import (
 	"github.com/eminwux/sbsh/internal/client/terminalstore"
 	"github.com/eminwux/sbsh/internal/discovery"
 	"github.com/eminwux/sbsh/internal/errdefs"
+	"github.com/eminwux/sbsh/internal/naming"
 	"github.com/eminwux/sbsh/pkg/api"
 )
 
@@ -314,6 +315,9 @@ func (s *Controller) createAttachTerminal(doc *api.ClientDoc) (*api.AttachedTerm
 	if doc.Spec.TerminalSpec.SocketFile != "" {
 		s.logger.Debug("attach by socket path", "socket", doc.Spec.TerminalSpec.SocketFile)
 		spec := *doc.Spec.TerminalSpec
+		if spec.ID == "" {
+			spec.ID = api.ID(naming.RandomID())
+		}
 		terminal := terminalstore.NewSupervisedTerminal(&spec)
 		if err := s.ss.Add(terminal); err != nil {
 			return nil, fmt.Errorf("%w: %w", errdefs.ErrTerminalStore, err)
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #173 Review — `fix(client): synthesize terminal ID for SocketFile-only attach`**

Tight, well-scoped fix that matches the suggestion in #157: synthesize an ID inside the `SocketFile != ""` branch only when the caller didn't supply one, on a local copy of the spec so the embedder's `doc.Spec.TerminalSpec` stays untouched. Tests cover both the synthesize-when-empty and preserve-when-provided paths, plus an explicit assertion that the caller's spec is not mutated. Imports remain correctly grouped/alphabetized; no new exported surface; no sensitive data in any new log/error paths.

LGTM to proceed.

PR #174, opened 2026-04-27
test: refresh stale comment in Test_ErrAttachNoIDOrName

### LINKED ISSUE #158: internal/client: refresh stale comment in Test_ErrAttachNoIDOrName

In `internal/client/controller_test.go:1059`, the inline comment reads:

\`\`\`go
TerminalSpec: &api.TerminalSpec{ID: "", Name: ""}, // Both empty should trigger ErrAttachNoTerminalSpec
\`\`\`

After #156 the validation in `Controller.Run` requires *all three* of `ID`, `Name`, and `SocketFile` to be empty before returning `ErrAttachNoTerminalSpec`. The test still passes (none of the three are set), but the comment should say "all three empty" so the rationale matches the code.

One-line edit; non-blocking.

eminwux:
## Summary
- The inline comment in `Test_ErrAttachNoIDOrName` (`internal/client/controller_test.go:1124`) read "Both empty should trigger ErrAttachNoTerminalSpec", but after #156 the validation in `internal/client/controller.go:208` checks all three of `ID`, `Name`, and `SocketFile`. Refresh the comment to reflect the actual condition.
- Pure comment edit; the test still passes (none of the three are set on the literal — `SocketFile` is the zero value).

## Test plan
- [x] `go test -timeout=5m \$(go list ./... | grep -v '/e2e\$')` — full unit suite green
- [x] `docs/examples/library-consumer`: `go build ./... && go vet ./...` — clean

Closes #158

### DIFF
--- a/internal/client/controller_test.go
+++ b/internal/client/controller_test.go
@@ -1121,7 +1121,7 @@ func Test_ErrAttachNoIDOrName(t *testing.T) {
 			ID:           api.ID(clientID),
 			LogFile:      "/tmp/sbsh-logs/s0",
 			RunPath:      viper.GetString("global.runPath"),
-			TerminalSpec: &api.TerminalSpec{ID: "", Name: ""}, // Both empty should trigger ErrAttachNoTerminalSpec
+			TerminalSpec: &api.TerminalSpec{ID: "", Name: ""}, // All three (ID, Name, SocketFile) empty should trigger ErrAttachNoTerminalSpec
 			ClientMode:   api.AttachToTerminal,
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #174 Review — `test: refresh stale comment in Test_ErrAttachNoIDOrName`**

Pure documentation refresh — the inline comment in `internal/client/controller_test.go:1124` now matches the three-field validation at `internal/client/controller.go:208`. No behavior change, test plan fully ticked, all CI green.

LGTM.

PR #180, opened 2026-04-27
test(signals_linux_test): drop redundant 2>/dev/null on inner wait

### LINKED ISSUE #169: test(signals_linux_test): drop redundant 2>/dev/null on inner wait
Surfaced during review of #167.

`internal/terminal/terminalrunner/signals_linux_test.go:71` (post-merge) — the inner `2>/dev/null` on `wait $child` is redundant:

\`\`\`sh
while kill -0 \$child 2>/dev/null; do
    wait \$child 2>/dev/null
done
\`\`\`

By the time `wait` would emit its "not a child" stderr, `kill -0` has already failed and the loop has exited. Drop the inner redirect — keeps the outer one only.

Deferred from #167; non-blocking style cleanup.

eminwux:
## Summary
- `internal/terminal/terminalrunner/signals_linux_test.go:76` had `wait $child 2>/dev/null` inside a `while kill -0 $child 2>/dev/null; do … done` loop. The inner stderr redirect is dead weight: the only stderr `wait` would emit is "not a child of this shell", and by the time that condition holds, `kill -0` on the same pid has already failed and the loop has exited. Outer redirect on `kill -0` is the one that actually suppresses output (during normal child-still-alive polling, that's the noisy call).
- Drop the inner `2>/dev/null`. Pure style cleanup — no behavior change.

## Test plan
- [x] `go test -count=1 -run TestSignalForwarder_TargetsProcessGroup ./internal/terminal/terminalrunner/ -v` — pass
- [x] `make sbsh-sb` — `./sbsh` and `./sb` both ELF executables (verified via `file`)
- [x] `go test -count=1 -timeout=300s -skip Test_HandleEvent_EvCmdExited $(go list ./... | grep -v '/e2e$')` — full unit suite green (skip is the known #138 deadlock currently in flight via PR #148; matches what CI sees today)
- [x] `E2E_BIN_DIR=$(pwd) go test -count=1 -timeout=600s ./e2e` — green (2.7s)

Closes #169

### DIFF
--- a/internal/terminal/terminalrunner/signals_linux_test.go
+++ b/internal/terminal/terminalrunner/signals_linux_test.go
@@ -73,7 +73,7 @@ child=$!
 while [ ! -f "$SIBLING_READY" ]; do sleep 0.01; done
 echo ready
 while kill -0 $child 2>/dev/null; do
-	wait $child 2>/dev/null
+	wait $child
 done
 echo done
 exit 0


eminwux:
Re-running failed CI jobs once (Build & Test, run [24989195257](https://github.com/eminwux/sbsh/actions/runs/24989195257)). Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #180 Review — `test(signals_linux_test): drop redundant 2>/dev/null on inner wait`**

LGTM. The inner `2>/dev/null` on `wait $child` is genuinely dead weight: the outer `while kill -0 $child 2>/dev/null` guard means `wait` is only called while `$child` is still a child of this shell, so the only "not a child of this shell" stderr path is unreachable. Interruption by the leader's own HUP trap returns wait silently (no stderr in dash or bash), and reaping a zombie child after a missed signal also produces no stderr. Outer redirect on `kill -0` correctly stays — that's the noisy call during normal polling. Pure style fix, no behavior change. Test plan fully checked.

LGTM to proceed.

PR #182, opened 2026-04-27
fix(terminal): fail fast on --terminal-name collision with active terminal

### LINKED ISSUE #114: Terminal - Document `--terminal-name` collision behavior (fail-fast or suffix)
Specify and document what happens when `sbsh terminal --terminal-name <name>` is invoked with a name that already belongs to an active terminal. Either fail fast with a clear error, or append a numeric suffix. Update CLI help and docs accordingly.

**Why:** Agents running scripted launches can't predict the outcome today.

From gap-analysis.md §10 (P3).

eminwux:
## Summary

Spec and document what happens when \`sbsh terminal --name <name>\` (and \`sbsh --terminal-name <name>\`) is invoked with a name already in use by an active terminal: **fail fast** with a clear error rather than silently spawn a shadow. Names from terminals that have exited (or whose owner process is gone, reconciled to \`Exited\` on scan) can be reused.

The check applies symmetrically to all three name-source paths called out in the issue and its comment thread:
- \`--name\` / \`--terminal-name\` flag
- \`sbsh terminal -\` reading the spec from stdin
- \`--file\` JSON spec

Why fail-fast over a numeric suffix: every other tool in the cohort the project takes its CLI cues from (\`docker run --name\`, \`kubectl\` resource names, \`tmux new-session -s\`) rejects the collision. A silent suffix corrupts what the user asked for, breaks scripted re-fetches by name, and leaves \`sb attach <name>\` ambiguous; \`FindTerminalByName\` resolves to first match today, so a shadow terminal would only be addressable by ID.

### Implementation

- New \`errdefs.ErrTerminalNameInUse\`.
- New \`internal/discovery.VerifyTerminalNameAvailable(ctx, logger, runPath, name)\` that scans \`runPath/terminals/*\`, runs the existing \`ReconcileTerminals\` (so PID-gone terminals stuck on a non-Exited state don't block reuse), and returns \`ErrTerminalNameInUse\` on the first active match. Empty name or empty runPath short-circuits to nil — random-name generation in \`BuildTerminalSpec\` runs *before* this check, so by the time we look the spec always has a name.
- Both CLI entry points (\`cmd/sbsh/terminal\` subcommand and the \`sbsh\` root which embeds a terminal in a client) call \`VerifyTerminalNameAvailable\` after the spec is built and before any controller is started. The check uses \`spec.RunPath\` so the JSON-spec / stdin-spec paths get the same treatment as the flag path.
- \`--name\` and \`--terminal-name\` flag descriptions, the \`sbsh terminal\` Long block, and \`docs/site/cli/sbsh.md\` all call out the uniqueness rule. The docs page gets a new "Name uniqueness" subsection that includes the error wording, the exited-name-reuse rule, and the documented best-effort caveat for the racing-launches edge case.

The check is best-effort: two simultaneous launches with the same name can both pass it. Documented explicitly — downstream tools (\`sb attach\` / \`detach\` / \`stop\` \`<name>\`) resolve the duplicate by first match, so the racing loser remains addressable by ID. A truly atomic guarantee would need either a name-keyed directory or a runPath-wide lock, both of which are larger architectural changes outside this issue's scope.

### Test plan

- [x] \`make sbsh-sb\` — \`./sbsh\` and \`./sb\` both ELF executables (verified via \`file\`)
- [x] \`go build ./...\` / \`go vet ./...\` — clean
- [x] \`go test -count=1 -timeout=60s ./internal/discovery/...\` — new \`TestVerifyTerminalNameAvailable\` passes (8 subtests covering empty inputs, no-terminals, active collision, non-colliding name, exited reuse, stale-PID reconciliation, and exited+active mix)
- [x] \`go test -count=1 -timeout=300s -skip Test_HandleEvent_EvCmdExited \$(go list ./... | grep -v '/e2e\$')\` — full unit suite green (skip is the known #138 deadlock currently in flight via PRs #148/#139; matches what CI sees today)
- [x] \`go test -count=1 -tags=integration -timeout=120s ./cmd/sb/get/...\` — green
- [x] \`E2E_BIN_DIR=\$(pwd) go test -count=1 -timeout=600s ./e2e\` — green (2.95s)
- [x] \`cd docs/examples/library-consumer && go build ./... && go vet ./...\` — clean

Closes #114

### DIFF
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -174,6 +174,16 @@ You can also use sbsh with parameters. For example:
 				return buildErr
 			}
 
+			if errAvail := discovery.VerifyTerminalNameAvailable(
+				cmd.Context(),
+				logger,
+				terminalSpec.RunPath,
+				terminalSpec.Name,
+			); errAvail != nil {
+				logger.Error("Terminal name collision", "error", errAvail)
+				return errAvail
+			}
+
 			logger.Debug("Built terminal spec", "terminalSpec", fmt.Sprintf("%+v", terminalSpec))
 
 			// Define a new ClientDoc
@@ -303,7 +313,7 @@ func setTerminalFlags(rootCmd *cobra.Command) error {
 		return err
 	}
 
-	rootCmd.Flags().String("terminal-name", "", "Optional name for the terminal")
+	rootCmd.Flags().String("terminal-name", "", "Optional name for the terminal (must be unique across active terminals)")
 	if err := viper.BindPFlag(config.SBSH_ROOT_TERM_NAME.ViperKey, rootCmd.Flags().Lookup("terminal-name")); err != nil {
 		return err
 	}
--- a/cmd/sbsh/terminal/terminal.go
+++ b/cmd/sbsh/terminal/terminal.go
@@ -296,6 +296,11 @@ Examples:
 If no terminal name is provided, a random name will be generated.
 If no command is provided, /bin/bash will be used by default.
 If no log filename is provided, a default path under the run directory will be used.
+
+Terminal names must be unique across active (non-Exited) terminals in the
+same run path; the launch fails fast when the requested name (passed via
+--name, --file, or stdin spec) is already in use. Names from terminals
+that have exited can be reused.
 `,
 		Args:         cobra.ArbitraryArgs,
 		SilenceUsage: true,
@@ -344,6 +349,15 @@ If no log filename is provided, a default path under the run directory will be u
 				return errdefs.ErrTerminalSpecNotFound
 			}
 
+			if errAvail := discovery.VerifyTerminalNameAvailable(
+				cmd.Context(),
+				logger,
+				terminalSpec.RunPath,
+				terminalSpec.Name,
+			); errAvail != nil {
+				return errAvail
+			}
+
 			logger.Debug("Built terminal spec", "terminalSpec", fmt.Sprintf("%+v", terminalSpec))
 
 			if logger.Enabled(context.Background(), slog.LevelDebug) {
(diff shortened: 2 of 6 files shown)


eminwux:
Re-running failed CI jobs once — `Build & Test` failed on the [latest run](https://github.com/eminwux/sbsh/actions/runs/25000543226). Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #182 Review — `fix(terminal): fail fast on --terminal-name collision with active terminal`**

LGTM. The new `VerifyTerminalNameAvailable` is symmetrically applied across all three name-source paths (`--name`/`--terminal-name` flag, `sbsh terminal -` stdin, `--file` JSON), the sentinel is wrapped with `%w` so `errors.Is(err, ErrTerminalNameInUse)` works downstream, and the test suite exercises the load-bearing edge cases — including stale-PID reconciliation flipping a `Ready` doc to `Exited` so the name becomes reusable. Docs and flag descriptions are updated consistently. No correctness, sensitive-data, or dead-code concerns:
- `errdefs.ErrTerminalNameInUse` is returned in `internal/discovery/terminals.go:311` and asserted in the new test.
- `discovery.VerifyTerminalNameAvailable` has callers in `cmd/sbsh/sbsh.go:177` and `cmd/sbsh/terminal/terminal.go:352`.
- `ReconcileTerminals` mutates in place via `&terminals[i]`, so the post-reconcile `t.Status.State` read in the verify loop sees fresh state.

- **Logging asymmetry** (`cmd/sbsh/sbsh.go:177` vs `cmd/sbsh/terminal/terminal.go:352`): the `sbsh` root path logs `"Terminal name collision"` before returning, but the `sbsh terminal` path returns the error without a structured log line. Cobra prints the error either way, so behavior is the same from the user's perspective; pure consistency nit.
- **Test's `deadPid = 0x7fffffff`** (`internal/discovery/terminals_test.go`): relies on `pid_max` being well below `INT_MAX`. Realistic on every Linux config (default 32K, configurable up to 2^22), but if anyone ever bumps `kernel.pid_max` toward `INT_MAX` on a CI runner the test could collide with a live process. Practically a non-issue — flagging only because the constant is unexplained beyond the comment.

Both are non-blocking. Nice work on the doc updates — the "Name uniqueness" subsection and the explicit best-effort caveat are exactly the right level of detail.

PR #184, opened 2026-04-28
fix(terminalrunner): prune dead writers instead of killing the PTY

### LINKED ISSUE #183: bug: ungracefully-disconnected client leaks dead writer in multiOutW; later broken pipe tears down whole terminal
## Summary

When a client disconnects abruptly (socket EOF, no explicit `sb detach`), its `pipeOutW` writer is left in `DynamicMultiWriter`. The next write that fails on that dead pipe is treated as a fatal terminal error and the entire PTY session is torn down — killing the long-running child process.

In a real session, this killed a multi-hour `claude` agent the moment a *new* `sb attach` produced its first output, ~4 seconds after reattach.

## Reproduction (from log `~/.sbsh/run/terminals/ab92232c/log`)

| Time | Event |
|---|---|
| 16:55:23 | session start, attacher A connects → `writer added total_writers=2` |
| 17:25:45 | attacher B connects → `total_writers=3` |
| 20:18:40 | B drops abruptly (`stdin->socket read error error=EOF`); `client connection handler exiting` fires — **no `writer removed` log line follows** |
| 20:18:40 → 20:54:11 | small periodic writes continue at `writers_count=3` — B's dead pipe is still in the slice (kernel buffer not yet full) |
| 20:54:11 | new attacher C connects → `total_writers=4` |
| 20:54:15 | `write to writer failed index=2 error=write \|1: broken pipe` (the leaked B writer) |
| 20:54:15 | `terminal EvError` → `initiating shutdown sequence` → `child process has exited` |

Synthetic repro: attach two clients, `kill -9` the second one (or just close its terminal), wait for the kernel pipe buffer to fill (or write enough data), reattach a third client. The next byte of PTY output destroys the session.

## Root cause — two cascading bugs

**1. Writer leak on abrupt disconnect.** `cleanupClient` in `internal/terminal/terminalrunner/connections.go:26-35` removes the client from `sr.clients` and closes the conn but never calls `multiOutW.Remove(client.pipeOutW)`. The only place that calls `multiOutW.Remove` is the explicit `Detach()` path at `internal/terminal/terminalrunner/lifecycle.go:294`. Anything that bypasses Detach (socket EOF, attacher process death, network drop) leaks a dead `*os.File` into the fan-out forever.

**2. One bad writer fails the whole multiwrite.** `DynamicMultiWriter.Write` at `internal/terminal/terminalrunner/terminal_multiwriter.go:45-51` returns on the first writer error rather than continuing past it / dropping the bad writer. The caller `terminalManagerReader` at `internal/terminal/terminalrunner/terminal.go:260-275` then sends `Event{Type: EvError}`, the controller closes, and the child shell is reaped. A single stale fan-out endpoint kills the session.

Bug 1 makes Bug 2 reachable. Bug 2 turns Bug 1 from a slow leak into an immediate session-kill.

## Suggested fix direction

- **Primary**: have `cleanupClient` call `multiOutW.Remove(client.pipeOutW)` (and close `pipeOutW`) so abrupt disconnects clean up the same way `Detach()` does. Consolidate the two paths so they can't drift again.
- **Defense in depth**: in `DynamicMultiWriter.Write`, on a writer error, log + auto-`Remove(w)` the failing writer and continue with the rest. Return an error from `Write` only if *all* writers failed (or none remain). The single dead client/subscriber should never be able to kill the PTY pipeline.
- **Tests**: unit test for multiwriter that one failing `io.Writer` is dropped and others continue; integration test that closing a client socket without `Detach` does not bring down the terminal on the next read.

## Acceptance criteria

- [ ] On socket EOF / abrupt client drop, the client's writer is removed from `multiOutW` (verified via `writer removed` log line and `total_writers` decrement).
- [ ] A failing writer in `DynamicMultiWriter.Write` does not propagate as `EvError` to `terminalManagerReader` while at least one writer remains; the failing writer is dropped from the slice.
- [ ] Regression test: attach client B, kill B without `sb detach`, fill enough output to break its pipe, reattach C — terminal session and child process survive.

eminwux:
## Summary

- An ungracefully-disconnected client used to leave its `pipeOutW` in `DynamicMultiWriter`; the next broken-pipe error there cascaded into a fatal `EvError` and killed the PTY along with the user's child process.
- `cleanupClient` now removes the client's writer from `multiOutW` before closing the conn — mirroring what the explicit `Detach()` path already did.
- `DynamicMultiWriter.Write` now drops a failing writer from the slice and continues; only when **every** writer has failed does it return the new `ErrAllWritersFailed`. A single dead fan-out endpoint can no longer take down the producer.
- `terminalManagerReader` only escalates to `EvError` on `ErrAllWritersFailed` — the log file alone keeps the session usable while attachers come and go.

Closes #183.

## Test plan

- [x] `make sbsh-sb` (verified `file ./sbsh` is `ELF 64-bit LSB executable`)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/terminal/terminalrunner/...` — green, including four new `TestDynamicMultiWriter_*` cases (drop-and-continue, all-fail returns `ErrAllWritersFailed`, empty multiwriter, redundant Remove is a no-op).
- [x] `make test` — full project suite passes.
- [ ] `make test-race` — `internal/terminal` `Test_HandleEvent_EvCmdExited` deadlocks at `controller.go:264`. **Reproduces on `origin/main` with the exact CI command (`go test -timeout=5m $(go list ./... | grep -v '/e2e$')`)** — pre-existing race in `Controller.Close` (sends to `closingCh` / `closeReqCh` after `cancel()`, with the goroutine that reads `closingCh` not yet having flipped `shuttingDown`). Filing separately.

## Notes for reviewer

- New exported symbol: `ErrAllWritersFailed`. Callers outside `terminalrunner` can detect the terminal-fatal case with `errors.Is`. The only in-tree caller updated here is `terminalManagerReader`.
- Per-write success log now includes a `failed=N` field so it's obvious in the log when a writer was just pruned.
- I considered closing `client.pipeOutR/pipeOutW` in `cleanupClient` to fully release the FDs, but `Detach()` doesn't close them either, and doing so would expand scope beyond the issue's acceptance criteria. Worth a follow-up if FD leaks become measurable.

### DIFF
--- a/internal/terminal/terminalrunner/connections.go
+++ b/internal/terminal/terminalrunner/connections.go
@@ -25,6 +25,19 @@ import (
 
 func (sr *Exec) cleanupClient(client *ioClient) {
 	sr.logger.Info("client connection handler exiting", "client", client.id)
+
+	// Drop this client's pipe from the fan-out *before* closing the conn so
+	// no further PTY output targets a dying writer. Without this, an abrupt
+	// disconnect (socket EOF, attacher kill, network drop) leaves the dead
+	// pipeOutW in multiOutW; a later broken pipe used to cascade into a
+	// fatal terminal event and reap the child process.
+	sr.ptyPipesMu.RLock()
+	multiOutW := sr.ptyPipes.multiOutW
+	sr.ptyPipesMu.RUnlock()
+	if multiOutW != nil && client.pipeOutW != nil {
+		multiOutW.Remove(client.pipeOutW)
+	}
+
 	if cerr := client.conn.Close(); cerr != nil {
 		sr.logger.Warn("error closing client connection", "err", cerr, "client", client.id)
 	}
--- a/internal/terminal/terminalrunner/terminal.go
+++ b/internal/terminal/terminalrunner/terminal.go
@@ -259,7 +259,10 @@ func (sr *Exec) terminalManagerReader(multiOutW io.Writer) {
 				sr.logger.Debug("writing to multiOutW")
 				_, errWrite := multiOutW.Write(buf[:n])
 				sr.logger.Debug("writing to multiOutW done")
-				if errWrite != nil {
+				// Only escalate when every fan-out endpoint has gone away —
+				// the log file alone keeps the session usable, and a single
+				// dropped attacher must not kill the PTY.
+				if errors.Is(errWrite, ErrAllWritersFailed) {
 					sr.logger.Error("error writing raw data to client", "err", errWrite)
 					errReturn := fmt.Errorf(
 						"terminalManagerReader multiWriter write error: %w: %w",
--- a/internal/terminal/terminalrunner/terminal_multiwriter.go
+++ b/internal/terminal/terminalrunner/terminal_multiwriter.go
@@ -17,10 +17,17 @@
 package terminalrunner
 
 import (
+	"errors"
 	"io"
 	"log/slog"
 	"sync"
 )
+
+// ErrAllWritersFailed is returned by DynamicMultiWriter.Write when every writer
+// in the snapshot returned an error. Callers may use this to detect the
+// terminal-fatal case (no remaining sinks) versus the routine case of a single
+// stale fan-out endpoint that has just been pruned.
+var ErrAllWritersFailed = errors.New("all writers failed")
 
 type DynamicMultiWriter struct {
 	mu      sync.RWMutex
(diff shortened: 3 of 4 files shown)


eminwux:
**PR #184 Review — `fix(terminalrunner): prune dead writers instead of killing the PTY`**

Clean, minimal fix for the cascade described in #183. Symmetric with the existing `Detach` path (`lifecycle.go:196-199`), the new `ErrAllWritersFailed` sentinel cleanly separates "single dead attacher" from "no sinks left", and the four new multiwriter tests cover the meaningful states (drop-and-continue, all-fail, empty, redundant Remove). No sensitive-data exposure in the new logs (`failed=N` is a count, not identity). Test plan box left unchecked is a pre-existing race documented with a `main`-branch reproduction — not introduced here.

- **Doc precision** `internal/terminal/terminalrunner/terminal_multiwriter.go:47-48`: "returns ErrAllWritersFailed only when no writer accepted the payload" reads as if an empty multiwriter would also error; the gate is actually `len(ws) > 0 && len(failed) == len(ws)`. Could tighten to "returns ErrAllWritersFailed only when at least one writer was registered and all of them failed." Cosmetic.
- **Exported sentinel without an external consumer** `internal/terminal/terminalrunner/terminal_multiwriter.go:30`: `ErrAllWritersFailed` is exported on speculation that callers outside the package will want `errors.Is`; the only in-tree consumer is `terminalManagerReader` in the same package. Fine as a sentinel pattern, just flagging that nothing today actually needs the export.

LGTM.

PR #185, opened 2026-04-29
fix: atomic shutdown gate in Controller.Close prevents deadlock

### LINKED ISSUE #138: bug: internal/terminal Controller.Close races shuttingDown flag, causing Test_HandleEvent_EvCmdExited to deadlock

eminwux:
## Summary
- `Controller.Close` in `internal/terminal/controller.go` and `internal/client/controller.go` gated its body on `shuttingDown.Load()`, but the flag was flipped in a separate goroutine that read `closingCh`. Two concurrent `Close` callers (typically `handleEvent → onClosed` racing the `Run` loop's `ctx.Done` branch) could both enter the body before the watcher set the flag, blow through the cap-1 sends to `closingCh` / `closeReqCh`, and pin `Run` forever — see issue #138 for the full goroutine trace.
- Replace the `Load`-guarded entry with `shuttingDown.CompareAndSwap(false, true)` so the flag flip is fused atomically with the act it gates: only the first caller proceeds, every subsequent caller short-circuits before any channel send. The `closingCh`-reader goroutine is removed (its only job was setting the flag).
- After the gate fix, `Close` cancels `ctx` (to unblock `WaitReady`) before sending on `closeReqCh`, so `Run` can wake on either signal. To keep the canonical close-request exit path (and surface the original reason instead of a generic `"context canceled"`), the `ctx.Done` branch in `Run` now drains `closeReqCh` non-blockingly first; falls through to the existing `ErrContextDone` path when the parent really was the canceller.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=60s -run '^Test_HandleEvent_EvCmdExited$' -v ./internal/terminal/` — previously deadlocked, now passes deterministically
- [x] `go test -race -count=100 -timeout=120s -run '^Test_HandleEvent_EvCmdExited$' ./internal/terminal/` — 100/100 race-clean
- [x] `go test -race -timeout=120s ./internal/terminal/... ./internal/client/...`
- [x] `make sbsh-sb` produced ELF executables (`file ./sbsh ./sb` confirms `ELF 64-bit LSB executable`)
- [x] `make test` (unit + integration + e2e)
- [x] `make test-race`

Closes #138

### DIFF
--- a/internal/client/controller.go
+++ b/internal/client/controller.go
@@ -92,7 +92,7 @@ func NewClientControllerWithIO(
 		// that bailed on ctx.Done would pin StartServer forever on its send.
 		rpcReadyCh: make(chan error, 1),
 		rpcDoneCh:  make(chan error),
-		closeReqCh:  make(chan error, 1),
+		closeReqCh: make(chan error, 1),
 		//nolint:mnd // event channel buffer size
 		eventsCh: make(chan clientrunner.Event, 32),
 		NewClientRunner: func(ctx context.Context, logger *slog.Logger, doc *api.ClientDoc, evCh chan<- clientrunner.Event) clientrunner.ClientRunner {
@@ -259,15 +259,21 @@ func (s *Controller) Run(doc *api.ClientDoc) error {
 	s.logger.Info("controller ready, entering main event loop")
 	close(s.ctrlReadyCh)
 
-	go func() {
-		errC := <-s.closingCh
-		s.shuttingDown.Store(true)
-		s.logger.Warn("controller closing", "reason", errC)
-	}()
-
 	for {
 		select {
 		case <-s.ctx.Done():
+			// See matching comment in internal/terminal/controller.go: prefer
+			// the close-request signal when it is already queued so the exit
+			// error carries the original reason rather than "context canceled".
+			select {
+			case errClose := <-s.closeReqCh:
+				s.logger.Warn("close request received", "error", errClose)
+				if errClose == nil {
+					return nil
+				}
+				return fmt.Errorf("%w: %w", errdefs.ErrCloseReq, errClose)
+			default:
+			}
 			ctxErr := s.ctx.Err()
 			s.logger.Warn("parent context canceled, shutting down controller")
 			if errC := s.Close(ctxErr); errC != nil {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #185 Review — `fix: atomic shutdown gate in Controller.Close prevents deadlock`**

CompareAndSwap-on-the-flag is the right fix — only the first caller enters the body, every subsequent one short-circuits before the cap-1 sends, killing the deadlock root cause from #138. Repro test passes deterministically and `-race -count=100` is clean. No correctness or sensitive-data concerns.

A couple of non-blocking nits:

- **Race window** `internal/terminal/controller.go` (`Close`, the `cancel(reason)` → `closeReqCh <- reason` ordering): `cancel` fires `ctx.Done` before `closeReqCh` receives `reason`, with a `closingCh` send and a potentially-slow `sr.Close(reason)` in between. If `Run` wakes on `ctx.Done` during that window, the new peek finds `closeReqCh` empty and falls through to `ErrContextDone` wrapping nil — losing the original reason on the exit error. Reordering the `closeReqCh <- reason` before `cancel(reason)` would close that gap; functionally a deadlock-free degradation today, just an observability nit.
- **Dead infra** `internal/client/controller.go:52,72,362`: removing the `closingCh`-watcher goroutine left `closingCh` with no production readers in the client (terminal's `WaitClose` still reads it, but client's `WaitClose` only reads `closedCh`). The cap-1 send in `Close` is harmless but the field, the `make`, and the send can all be removed (tests at `controller_test.go:1751/1794` would need to drop their drain goroutines too).

LGTM unconditional — both items are cleanups, not blockers.

PR #189, opened 2026-05-01
fix(terminal): configurable control-socket mode and gid

### LINKED ISSUE #188: terminal: socket hard-coded to 0o600 owner-only — block kukeon-group attach across UID boundaries
`sbsh terminal` chmods its control socket to `0o600` after `Listen()` and never sets a group, so any caller whose uid is not the listener's uid is rejected at `connect()` — even when a group-traversable directory is set up around it. This blocks the kukeon attach path end-to-end whenever the host operator and the wrapped container do not share a uid (the common case: non-root operator in `kukeon` group, container with `USER claude`).

## Repro

`sbsh` ref: `v0.8.0` (also `main` at this writing).

Stand up a terminal as one uid and try to dial it as a different uid that *is* a member of the socket directory's group:

```
$ sudo install -d -m 02750 -o 1000 -g 986 /tmp/sbtest
$ sudo -u '#1000' sbsh terminal \
    --run-path /tmp/sbtest \
    --socket /tmp/sbtest/socket \
    --capture-file /tmp/sbtest/cap \
    --log-file /tmp/sbtest/log \
    -- /bin/sleep 60 &

# Different uid (uid 1001), member of gid 986:
$ id -u; id -G | tr ' ' '\n' | grep -F 986
1001
986
$ ls -l /tmp/sbtest/socket
srw------- 1 1000 1000 0 May  1 16:00 /tmp/sbtest/socket
$ sbsh attach --socket /tmp/sbtest/socket
Error: error attaching client: ping failed: dial unix /tmp/sbtest/socket: connect: permission denied
```

The directory was deliberately mode `02750 1000:986` so group members have traverse + read access; the socket itself, however, is `0o600` and nobody outside the listener's uid can connect.

## Observed

`internal/terminal/terminalrunner/sockets.go:63`:

```go
if errChmod := os.Chmod(socketFile, 0o600); errChmod != nil {
    sr.logger.Error(\"OpenSocketCtrl: failed to chmod socket file\", \"socket\", socketFile, \"error\", errChmod)
    _ = ctrlLn.Close()
    return errChmod
}
```

There is no path to relax this mode and no path to set the socket's group, so any deployment that wants to share a terminal across uids in the same group (kukeon's `kukeon` operator group is the concrete example, but nothing here is kukeon-specific) cannot. Linux `connect(2)` on a Unix domain socket requires write permission on the socket inode; `0o600` with the wrong uid means denied, full stop.

## Root cause

The socket mode and gid were hard-coded for the original single-user case. The wrapping use case (kukeon, but also any future \"shared dev terminal owned by a service uid\") needs to (a) widen the mode to allow group access and (b) set the socket's group to a host-side group the attacher is a member of.

## Suggested fix

Make socket ownership configurable via three layers — flags, env, and the `TerminalProfile` YAML — with safe defaults that preserve today's behavior.

### 1. New CLI flags on `sbsh terminal`

`cmd/sbsh/terminal/terminal.go` already binds each flag through viper; follow the same pattern:

```go
terminalCmd.Flags().String(\"socket-mode\", \"\",
    \"Octal mode applied to the control socket after Listen (default 0600)\")
_ = viper.BindPFlag(config.SBSH_TERM_SOCKET_MODE.ViperKey,
    terminalCmd.Flags().Lookup(\"socket-mode\"))

terminalCmd.Flags().Int(\"socket-gid\", -1,
    \"Numeric GID applied to the control socket via chown after Listen; -1 leaves group unchanged\")
_ = viper.BindPFlag(config.SBSH_TERM_SOCKET_GID.ViperKey,
    terminalCmd.Flags().Lookup(\"socket-gid\"))
```

`--socket-mode` is taken as a string and parsed with `strconv.ParseUint(s, 8, 32)` so users can write `0660` or `660` and either form works; reject any mode with bits outside `0o7777` to keep the surface small. `--socket-gid` defaults to `-1` (sentinel for \"don't chown\") so the omission case is identical to current behavior.

### 2. New env keys in `cmd/config/env.go`

Mirror the existing `SBSH_TERM_SOCKET` pair:

```go
SBSH_TERM_SOCKET_MODE = DefineKV(\"SBSH_TERM_SOCKET_MODE\", \"sbsh.terminal.socketMode\")
SBSH_TERM_SOCKET_GID  = DefineKV(\"SBSH_TERM_SOCKET_GID\",  \"sbsh.terminal.socketGid\")
```

### 3. Profile YAML schema — `pkg/api/profile.go`

Add an optional `socket` block on `TerminalProfileSpec` so the same configuration can travel with the profile (kukeon writes profiles per cell; this is the cleanest plumbing point):

```go
type TerminalProfileSpec struct {
    RunTarget     RunTarget     `json:\"runTarget\"        yaml:\"runTarget\"`
    RestartPolicy RestartPolicy `json:\"restartPolicy\"    yaml:\"restartPolicy\"`
    Shell         ShellSpec     `json:\"shell\"            yaml:\"shell\"`
    Stages        StagesSpec    `json:\"stages\"           yaml:\"stages\"`
    Socket        *SocketSpec   `json:\"socket,omitempty\" yaml:\"socket,omitempty\"`
}

// SocketSpec configures the control socket's filesystem permissions. Both
// fields are optional; omitting the block keeps the legacy 0600 owner-only
// behavior. mode is an octal string (\"0660\") to avoid YAML's int/octal
// ambiguity; gid is a numeric GID resolved by the caller.
type SocketSpec struct {
    Mode string `json:\"mode,omitempty\" yaml:\"mode,omitempty\"`
    Gid  *int   `json:\"gid,omitempty\"  yaml:\"gid,omitempty\"`
}
```

Resolution order at runtime: explicit `--socket-mode`/`--socket-gid` flags win; otherwise the profile's `spec.socket` values are used; otherwise defaults (mode `0o600`, no chown).

### 4. Apply at `Listen()` time — `internal/terminal/terminalrunner/sockets.go`

Replace the hard-coded line:

```go
mode := sr.metadata.Spec.SocketMode  // resolved value, default 0o600
if errChmod := os.Chmod(socketFile, mode); errChmod != nil {
    _ = ctrlLn.Close()
    return errChmod
}
if gid := sr.metadata.Spec.SocketGid; gid >= 0 {
    if errChown := os.Chown(socketFile, -1, gid); errChown != nil {
        _ = ctrlLn.Close()
        return errChown
    }
}
```

`-1` in `Chown(socket, -1, gid)` leaves the uid unchanged; the listener's uid stays the owner, only the group is rewritten.

### 5. Docs — `docs/profiles/README.md`

Add a `socket` subsection to the profile field reference, e.g.:

```yaml
spec:
  # ...
  socket:
    mode: \"0660\"   # Octal string; defaults to \"0600\" if omitted
    gid: 986         # Numeric host GID; omit to leave group unchanged
```

Plus a one-paragraph note that the default keeps owner-only access (the current behavior) and that widening either field is opt-in for callers that share the terminal across uids in a group.

### 6. Backward compatibility

Existing invocations and existing profiles continue to produce `srw------- listener-uid:listener-gid` sockets — none of the new fields are required, the env keys default to unset, and the listener uses the legacy mode when no override is configured. The kukeon side then closes its own gap (eminwux/kukeon#260) by passing `--socket-mode 0660 --socket-gid <kukeon-gid>` (or writing those into the per-cell profile) when wrapping a container.

## Out of scope

- Setting the *uid* of the socket (only group/mode are configurable here). Callers that want a different owner-uid should pre-create the socket dir and run `sbsh terminal` as that uid; that is what kukeon does today via `oci.WithImageConfig`.
- Validating that the supplied gid exists on the host — the host-side caller is responsible. Out of band, an unknown gid will surface as `chown: invalid argument` from the kernel and abort `Listen()` cleanly.

eminwux:
## Summary
- Resolves the kukeon group-attach blocker where the control socket was hard-coded to `0600` and unreachable across uid boundaries.
- Three configuration layers (CLI flags, env vars, profile YAML) for socket mode/gid, all funneled through `BuildTerminalSpecParams` and applied after `Listen` in `OpenSocketCtrl` (chmod always, chown only when gid is set).
- Default behavior preserved: with no flag, no env, no `socket` block, the socket still gets `0600` and the group is left untouched.
- `BuildTerminalSpecParams.SocketGID` is `*int` (not `int`) so that "unset" is distinguishable from "explicit gid 0 = root" — a plain `int` made the zero-value caller path silently chown to root and reject every default-profile launch under non-root.
- `viper.GetInt` is unreliable for negative-default pflags (BindPFlag forwards a *changed* pflag; an unchanged `-1` falls through to viper's int zero value), so the gid is read directly from cobra with an env-var fallback in `readSocketGID` / `readRootSocketGID`.
- The flag pair is exposed at both the root (`sbsh --terminal-socket-mode/--terminal-socket-gid`) and the `terminal` subcommand (`sbsh terminal --socket-mode/--socket-gid`), each backed by its own env var.

## Test plan
- [x] `make sbsh-sb` and `file ./sbsh` confirms an ELF executable
- [x] `go vet ./...`
- [x] `make test` (full CI suite, including `./e2e`) — green
- [x] `TestSbsh_Default` re-run 5×, all green (initial implementation produced 5/5 socket-not-found failures before the `*int` sentinel + viper-bypass fix)
- [x] `./sbsh --help` and `./sbsh terminal --help` show the new flags with `(default -1)` for gid

Closes #188

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -160,6 +160,10 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_TERM_SOCKET = DefineKV("SBSH_ROOT_TERM_SOCKET", "sbsh.root.terminal.socket")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_ROOT_TERM_SOCKET_MODE = DefineKV("SBSH_ROOT_TERM_SOCKET_MODE", "sbsh.root.terminal.socketMode")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_ROOT_TERM_SOCKET_GID = DefineKV("SBSH_ROOT_TERM_SOCKET_GID", "sbsh.root.terminal.socketGid")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_TERM_DISABLE_SET_PROMPT = DefineKV(
 		"SBSH_ROOT_TERM_DISABLE_SET_PROMPT",
 		"sbsh.root.terminal.disableSetPrompt",
@@ -184,6 +188,10 @@ var (
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_SOCKET = DefineKV("SBSH_TERM_SOCKET", "sbsh.terminal.socket")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_TERM_SOCKET_MODE = DefineKV("SBSH_TERM_SOCKET_MODE", "sbsh.terminal.socketMode")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_TERM_SOCKET_GID = DefineKV("SBSH_TERM_SOCKET_GID", "sbsh.terminal.socketGid")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_ID = DefineKV("SBSH_TERM_ID", "sbsh.terminal.id")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -26,6 +26,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"strconv"
 	"syscall"
 	"time"
 
@@ -165,6 +166,8 @@ You can also use sbsh with parameters. For example:
 					LogFile:          viper.GetString(config.SBSH_ROOT_TERM_LOG_FILE.ViperKey),
 					LogLevel:         viper.GetString(config.SBSH_ROOT_TERM_LOG_LEVEL.ViperKey),
 					SocketFile:       viper.GetString(config.SBSH_ROOT_TERM_SOCKET.ViperKey),
+					SocketMode:       viper.GetString(config.SBSH_ROOT_TERM_SOCKET_MODE.ViperKey),
+					SocketGID:        readRootSocketGID(cmd),
 					DisableSetPrompt: viper.GetBool(config.SBSH_ROOT_TERM_DISABLE_SET_PROMPT.ViperKey),
 				},
 			)
@@ -349,6 +352,30 @@ func setTerminalFlags(rootCmd *cobra.Command) error {
 		return err
 	}
 
+	rootCmd.Flags().String(
+		"terminal-socket-mode",
+		"",
+		"Octal mode applied to the control socket after Listen (default 0600)",
+	)
+	if err := viper.BindPFlag(config.SBSH_ROOT_TERM_SOCKET_MODE.ViperKey, rootCmd.Flags().Lookup("terminal-socket-mode")); err != nil {
+		return err
+	}
+	if err := viper.BindEnv(config.SBSH_ROOT_TERM_SOCKET_MODE.ViperKey, config.SBSH_ROOT_TERM_SOCKET_MODE.EnvVar()); err != nil {
+		return err
+	}
+
+	rootCmd.Flags().Int(
+		"terminal-socket-gid",
+		-1,
+		"Numeric GID applied to the control socket via chown after Listen; -1 leaves group unchanged",
+	)
+	if err := viper.BindPFlag(config.SBSH_ROOT_TERM_SOCKET_GID.ViperKey, rootCmd.Flags().Lookup("terminal-socket-gid")); err != nil {
+		return err
+	}
+	if err := viper.BindEnv(config.SBSH_ROOT_TERM_SOCKET_GID.ViperKey, config.SBSH_ROOT_TERM_SOCKET_GID.EnvVar()); err != nil {
+		return err
+	}
+
 	rootCmd.Flags().Bool("terminal-disable-set-prompt", false, "Disable setting the prompt")
 	if err := viper.BindPFlag(config.SBSH_ROOT_TERM_DISABLE_SET_PROMPT.ViperKey, rootCmd.Flags().Lookup("terminal-disable-set-prompt")); err != nil {
 		return err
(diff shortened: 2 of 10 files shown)


eminwux:
**PR #189 Review — `fix(terminal): configurable control-socket mode and gid`**

Clean, well-motivated change. The three-layer precedence (CLI > env > profile > default 0o600) is wired consistently; `*int` for `SocketGID` correctly distinguishes "unset" from "explicit 0 = root"; the `readSocketGID`/`readRootSocketGID` cobra-direct fallback documents and dodges the `viper.GetInt` BindPFlag pitfall; `applySocketOverrides` runs after `addInputValuesToTerminal` and the latter doesn't touch socket fields, so no double-handling. No sensitive data in error logs. CI green.

A few non-blocking nits:

- **Public builder gap** `pkg/builder/terminal.go:178-191`: `pkg/builder.BuildTerminalSpec` does not forward `SocketMode`/`SocketGID` into the new `BuildTerminalSpecParams` fields, so external embedders using the public façade cannot override mode/gid the way CLI/env callers can. Profile-level config still flows through, so this is a coverage gap rather than a regression — but the comment at `pkg/builder/terminal.go:35` claims the struct mirrors the "subset … external callers are expected to control", and the new fields are arguably part of that subset. Worth a follow-up `WithSocketMode`/`WithSocketGID` pair.
- **Profile gid validation** `internal/profile/profile.go:113-116`: `profile.Spec.Socket.GID` is forwarded as-is. A nonsense value like `gid: -5` in YAML would surface only as a runtime `chown` failure (cast to uint32) rather than a profile-load error. Consider rejecting negative gids at parse time, symmetric with `parseSocketMode`'s mask check.
- **Missing unit test for `parseSocketMode`** `internal/profile/profile.go:130`: small pure function with octal edge cases ("0660" / "660" / out-of-mask / non-octal) — a table-driven test would be cheap and would lock in the documented behavior. CI exercises it transitively, but a direct test guards against future drift.

LGTM to merge — the items above are all follow-ups, not blockers.

eminwux:
Addressed the three follow-up nits in 9743946:

- **Public builder gap** — added `WithSocketMode(mode string)` and `WithSocketGID(gid int)` to `pkg/builder`, wired through `terminalConfig` into `BuildTerminalSpecParams.SocketMode` / `SocketGID`. `WithSocketGID` always sets the pointer when called, so `WithSocketGID(0)` is "explicit root" (matches the `*int` sentinel contract); omitting the option leaves it nil.
- **Profile gid validation** — `CreateTerminalFromProfile` now rejects negative `spec.socket.gid` values at parse time with a profile-named error, symmetric with `parseSocketMode`'s mask check. Zero stays valid (root).
- **`parseSocketMode` unit test** — table-driven test in `internal/profile/profile_test.go` covering leading-zero / no-leading-zero octal, the setgid bit, the full `0o7777` mask, out-of-mask (`10000`), non-octal digits (`0680`), non-numeric, and empty input.

Local validation: `make sbsh-sb` (ELF), `go vet ./...`, `go build ./...`, `make test` (full CI suite incl. `./e2e`) — all green.

eminwux:
**PR #189 Re-review — `fix(terminal): configurable control-socket mode and gid`**

Follow-up commit `9743946` cleanly addresses all three nits from the prior review:

- `internal/profile/profile.go:118-124`: negative-gid rejection at profile load, symmetric with the mode-mask check.
- `internal/profile/profile_test.go`: `TestParseSocketMode` table test covers leading-zero / no-leading-zero / setgid / full-mask / out-of-mask / non-octal-digit / non-numeric / empty.
- `pkg/builder/terminal.go:148-163`: `WithSocketMode`/`WithSocketGID` public builder options with docstrings calling out the unset-vs-explicit-zero distinction; both forwarded into `BuildTerminalSpecParams`.

LGTM — ready to merge.

PR #193, opened 2026-05-03
feat(attach): add ErrDetached/ErrPeerClosed sentinels for clean-vs-hangup branching

### LINKED ISSUE #192: pkg/attach: expose public sentinel for clean ^]^] detach so embedders can distinguish it from peer hangup
## Context

`pkg/attach.Run` returns a single error for two semantically distinct end-of-session paths:

1. **Clean operator detach** — user pressed `Ctrl+] Ctrl+]`; the `EvDetach` event fires (`internal/client/clientrunner/io.go:90`), the controller sends Detach, and the dual-copier tears down. The workload (remote terminal) stays alive — the user wants to re-attach later.
2. **Peer hangup / workload exit** — the remote terminal closes its end (shell `exit`, workload crash, kernel killed PID 1). The dual-copier tears down via the same shutdown path.

Both surface to embedders as the composed message `\"close requested: read/write routines exited\"` (the `EvError` event built at `internal/client/clientrunner/io.go:131` wrapped by the controller close path). `pkg/attach/errors.go` exposes only `ErrSocketPathRequired`; there is no public sentinel that lets embedders branch on \"detach\" vs \"hangup\".

## The cost in kukeon (downstream)

kukeon's `kuke attach` and `kuke run -a --rm` need this distinction:

- `kuke attach` (`cmd/kuke/attach/attach.go:161-168`) currently does `strings.Contains(err.Error(), \"close requested\") && strings.Contains(err.Error(), \"read/write routines exited\")` to map clean detach to exit code 0. Brittle: any reword of the sbsh error message silently breaks it.
- `kuke run -a --rm` (issue [eminwux/kukeon#279](https://github.com/eminwux/kukeon/issues/279)) needs to leave the cell alive on clean detach but kill it on peer hangup / shell exit. Today it conflates both as \"operator is done\" and kills the cell on every detach. Fixing kukeon#279 cleanly requires sbsh telling the embedder which case fired.

## Proposed shape

Add public sentinels in `pkg/attach/errors.go` (or wrap the controller errors so `errors.Is` works):

```go
// ErrDetached signals that Run returned because the operator pressed
// the in-band ^]^] detach keystroke (or some peer triggered Detach via
// the controller). The remote terminal is still alive; embedders can
// re-attach to the same SocketPath.
var ErrDetached = errors.New(\"pkg/attach: client detached\")

// ErrPeerClosed signals that Run returned because the remote terminal
// closed the connection — workload exited, peer crashed, or session
// otherwise ended on the far side. Re-attaching will fail.
var ErrPeerClosed = errors.New(\"pkg/attach: peer closed connection\")
```

Implementation sketch (no patch, just where to thread it):

- `internal/client/clientrunner/io.go` already emits two distinct events: `EvDetach` (line 90) for the keystroke path and the `EvError` with `\"read/write routines exited\"` (line 131) for any teardown. The dispatching site in `dualcopier`/`lifecycle` knows which fired.
- The controller's close path (`internal/terminal/terminalrunner/lifecycle.go:73` logs `\"terminal close requested\"`) currently composes the message-only error that bubbles to `pkg/attach.Run`. Plumb the cause forward so the public boundary can wrap with the right sentinel.

Embedders then use `errors.Is(err, attach.ErrDetached)` instead of `strings.Contains`.

## Suggested fix

1. Add `ErrDetached` and `ErrPeerClosed` to `pkg/attach/errors.go` with godoc that pins the contract (when each fires; what \"detached\" implies for the workload).
2. Thread the cause from the runner's two shutdown paths to `pkg/attach.Run`'s return — wrap with the matching sentinel via `fmt.Errorf(\"%w: ...\", attach.ErrDetached, …)` so existing string-based output stays readable while `errors.Is` works.
3. Add a unit test in `pkg/attach` (or `internal/client/clientrunner`) that drives both paths through a fake terminal and asserts the right sentinel surfaces.
4. Bump sbsh's minor version on release; cross-reference [eminwux/kukeon#279](https://github.com/eminwux/kukeon/issues/279) in the PR body so the kukeon-side cleanup can land once the new sbsh release is consumed.

## Backwards compatibility

Wrapping existing errors with the new sentinels keeps the human-readable text the same, so kukeon's current `isCleanDetach` string-match keeps working until kukeon migrates. No breaking change to embedders that don't opt in.

eminwux:
## Summary
- Add `attach.ErrDetached` and `attach.ErrPeerClosed` so embedders (notably kukeon) can `errors.Is` against the public sentinels instead of brittle string matching on the composed `\"close requested: read/write routines exited\"` message.
- Plumb the cause forward: a new `Controller.detachRequested` flag, set in `Detach()`, lets `handleEvent` tag the `EvError` close reason with `errdefs.ErrClientDetached` when a clean detach was requested, otherwise `errdefs.ErrPeerClosed`. `pkg/attach.Run` translates those internal sentinels to the public ones at the boundary.
- Cross-references kukeon's blocked clean-up in [eminwux/kukeon#279](https://github.com/eminwux/kukeon/issues/279). Setup-time errors (`errdefs.ErrAttach`, etc.) are intentionally **not** retagged so existing matchers keep firing.
- Backwards-compat: the wrapped chain still includes the human-readable substrings kukeon currently keys on (`close requested`, `read/write routines exited`), so the `strings.Contains` check at `cmd/kuke/attach/attach.go:161-168` keeps working until kukeon migrates to `errors.Is`.

## Test plan
- [x] `make sbsh-sb` (canonical build, `file ./sbsh` confirmed ELF executable, `sb` hardlink intact)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full suite including `./cmd/sb...`, `./cmd/sbsh...`, `./internal/terminal...`, `./internal/client...`, integration tag, and `./e2e` all pass
- [x] New unit coverage in `pkg/attach`: `TestClassifySessionEnd_NilPassthrough`, `TestClassifySessionEnd_ClientDetached`, `TestClassifySessionEnd_PeerClosed`, `TestClassifySessionEnd_SetupErrorPassesThrough` exercise the public translation, including the legacy substring assertion for kukeon's pre-migration string match.
- [x] New controller-level coverage in `internal/client`: `Test_EventError_PeerClosed` and `Test_EventDetachThenError_ClientDetached` drive both real-world flows (lone EvError vs. EvDetach-then-EvError) through the fake runner and assert the `Run` exit error matches the right `errdefs.*` sentinel.

Closes #192

### DIFF
--- a/internal/client/controller.go
+++ b/internal/client/controller.go
@@ -54,6 +54,14 @@ type Controller struct {
 	closedCh     chan struct{}
 	shuttingDown atomic.Bool
 
+	// detachRequested is set whenever Detach() runs (keystroke event or
+	// external RPC). The dual-copier teardown that follows surfaces as
+	// EvError "read/write routines exited" — indistinguishable from a
+	// peer hangup at that layer. handleEvent reads this flag to decide
+	// whether to tag the close reason as ErrClientDetached or
+	// ErrPeerClosed so pkg/attach can surface the right public sentinel.
+	detachRequested atomic.Bool
+
 	eventsCh chan clientrunner.Event
 
 	rpcReadyCh chan error
@@ -400,7 +408,7 @@ func (s *Controller) handleEvent(ev clientrunner.Event) {
 
 	case clientrunner.EvError:
 		s.logger.ErrorContext(s.ctx, "terminal EvError", "id", ev.ID, "err", ev.Err)
-		s.onClosed(ev.ID, ev.Err)
+		s.onClosed(ev.ID, s.classifySessionEnd(ev.Err))
 	case clientrunner.EvDetach:
 		s.logger.InfoContext(s.ctx, "terminal EvDetach", "id", ev.ID)
 		err := s.Detach()
@@ -450,11 +458,31 @@ func (s *Controller) WaitClose() error {
 }
 
 func (s *Controller) Detach() error {
+	// Mark before issuing the RPC: the terminal's Detach response triggers
+	// an asynchronous IO conn close, and the resulting EvError must be
+	// classified as a clean detach even when the runner returns an error
+	// from the RPC itself.
+	s.detachRequested.Store(true)
 	// Request detach from terminal
 	if err := s.sr.Detach(); err != nil {
 		return fmt.Errorf("%w: %w", errdefs.ErrDetachTerminal, err)
 	}
 	return nil
+}
+
+// classifySessionEnd wraps the dual-copier teardown error with the
+// matching internal sentinel so pkg/attach.Run can surface a public
+// detach-vs-hangup distinction. A nil cause still carries the sentinel
+// so embedders can errors.Is even when the runner emitted no detail.
+func (s *Controller) classifySessionEnd(cause error) error {
+	sentinel := errdefs.ErrPeerClosed
+	if s.detachRequested.Load() {
+		sentinel = errdefs.ErrClientDetached
+	}
+	if cause == nil {
+		return sentinel
+	}
+	return fmt.Errorf("%w: %w", sentinel, cause)
 }
 
 func (s *Controller) Ping(in *api.PingMessage) (*api.PingMessage, error) {
--- a/internal/client/controller_test.go
+++ b/internal/client/controller_test.go
@@ -1603,6 +1603,205 @@ func Test_EventDetachFailure(t *testing.T) {
 	<-exitCh
 }
 
+// Test_EventError_PeerClosed asserts that a bare EvError (no preceding
+// EvDetach) tags the close reason with errdefs.ErrPeerClosed so
+// pkg/attach.Run can surface attach.ErrPeerClosed at the public
+// boundary.
+func Test_EventError_PeerClosed(t *testing.T) {
+	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelDebug}))
+	sc := NewClientController(context.Background(), logger).(*Controller)
+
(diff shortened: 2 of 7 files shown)


eminwux:
**PR #193 Review — `feat(attach): add ErrDetached/ErrPeerClosed sentinels for clean-vs-hangup branching`**

Clean boundary translation: internal `errdefs.ErrClient­Detached`/`ErrPeerClosed` are set on the close reason at the controller, then `pkg/attach.Run` re-wraps them as the public `attach.ErrDetached`/`ErrPeerClosed` so embedders can `errors.Is` instead of substring-matching. Backwards-compat substrings preserved, setup errors deliberately untouched, no leaks. Tests exercise both real-world flows (lone EvError → PeerClosed, EvDetach→EvError → ClientDetached) and the helper's nil/passthrough/sentinel branches. CI green.

Minor observations (all non-blocking):

- **Sticky flag** `internal/client/controller.go:465`: `detachRequested` is set before the `sr.Detach()` RPC and never cleared. The doc comment justifies this for the RPC-failure case, but it also means any later EvError after an explicit Detach is force-classified as `ErrClientDetached` — fine because the controller's lifecycle ends at first close, but worth a thought if Run-loop reuse ever appears.
- **EvCmdExited bypass** `internal/client/controller.go:407`: only EvError routes through `classifySessionEnd`, so a non-nil `EvCmdExited` error reaches the public boundary unclassified. In practice the IO drop fires EvError next and that gets classified, so the visible behavior matches the doc — flagging in case the dual-event ordering ever changes.

LGTM.

PR #198, opened 2026-05-05
ci: make pre-commit pass on clean main and enforce in CI

eminwux:
## Summary
- Hooks failed on `main` with no local edits — `check-yaml` rejected `mkdocs.yml`'s `!!python/name` constructor and the multi-doc `profiles.yaml` files; `end-of-file-fixer`, `go fmt`, and `prettier` rewrote eight files of pre-existing drift. Configure `check-yaml` with `--unsafe` and `--allow-multiple-documents`, accept the formatter output for the eight drifted files, and add a `pre-commit run --all-files` step to `test.yaml` so future drift fails CI before it lands.
- The fixer-applied diffs are mechanical (gofmt struct/method alignment, prettier table padding, end-of-file trims, prettier wrap of an inline-code continuation in `gap-analysis.md`); no behavioral changes.

## Test plan
- [x] `pre-commit run --all-files` — passes
- [x] `make sbsh-sb` — produces ELF executable, `file ./sbsh` confirms
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')`
- [x] `go test -tags=integration ./cmd/sb/get/...`
- [x] `cd docs/examples/library-consumer && go build ./... && go vet ./...`

Closes #197

Issue #197, opened 2026-05-05
ci: pre-commit run --all-files fails on clean main


eminwux:
## Repro
Reproduced on clean `main` at `33e5bb0` (HEAD as of filing):

```
git checkout main && git pull
git status   # clean
pre-commit run --all-files
```

## Observed
Five hooks fail on files that have not been touched in this session:

- **fix end of files** modifies `internal/client/controller_rpcs_test.go`, `cmd/sb/validate/profiles.go`
- **check yaml** rejects `mkdocs.yml` (mkdocs `!!python/name` tag), `docs/profiles/profiles.yaml`, `docs/site/guides/profiles/profiles.yaml` (multi-doc YAML — `check-yaml` requires single-document)
- **go fmt** rewrites `internal/terminal/terminalrunner/reaper_other.go`, `cmd/sb/validate/profiles.go`
- **prettier** rewrites `docs/gap-analysis.md`, `docs/site/guides/configuration.md`, `docs/site/reference/library.md`, `.github/workflows/mkdocs.yaml`, `docs/examples/library-consumer/README.md`

`go fmt` and `prettier` rewrite the files in place and exit non-zero, so a contributor following `CLAUDE.md`'s "run `pre-commit run --all-files` before `git commit`" guidance will end up with ~8 unrelated, hook-induced edits in their working tree before they have staged anything.

## Root cause
Not yet pinpointed; candidate shortlist:
- `.pre-commit-config.yaml` `check-yaml` is configured to scan all YAML, but `mkdocs.yml` uses a custom `!!python/name` constructor and the two `profiles.yaml` files are intentionally multi-document — these need `--unsafe` and `--allow-multiple-documents` (or per-file excludes).
- The eight files flagged by `end-of-file-fixer` / `go-fmt` / `prettier` are real drift that pre-dates the hooks landing — they were never reformatted in the commit that introduced the hooks.
- CI does not enforce `pre-commit run --all-files` (`.github/workflows/test.yaml` runs `go vet` + `go test` only), so the drift accumulates unchecked.

## Suggested fix
Two-step: (1) one cleanup PR that runs each fixer against the offending files and updates `check-yaml` args (`--unsafe`, `--allow-multiple-documents`) so the hooks pass on a clean tree; (2) add a `pre-commit run --all-files` step to `test.yaml` so future drift fails CI before it lands. Step 1 unblocks the contributor flow described in the project `CLAUDE.md`; step 2 prevents regression.

PR #199, opened 2026-05-06
test(profile): cover CreateTerminalFromProfile socket gid validation

### LINKED ISSUE #191: profile: validate negative gid in spec.socket.gid at load time
Surfaced during review of #189.

`internal/profile/profile.go:113-116` forwards `profile.Spec.Socket.GID` (`*int`) as-is into `spec.SocketGID`. A nonsense value like `gid: -5` in profile YAML survives load and surfaces only as a runtime \`chown\` failure (Go casts the int to uint32, yielding 4294967291, which the kernel rejects with EINVAL).

This is asymmetric with the mode path: \`parseSocketMode\` rejects out-of-mask bits at load time so the user gets a clear \"invalid spec.socket.mode in profile X\" error.

**Suggested fix:** in the \`profile.Spec.Socket.GID != nil\` branch, reject \`*g < 0\` with \`fmt.Errorf(\"invalid spec.socket.gid in profile %q: negative gid %d\", profile.Metadata.Name, *g)\` before assigning. The CLI sentinel \`-1\` semantics live in \`readSocketGID\`/\`readRootSocketGID\` and don't apply to profile YAML — the docs say \"omit to leave group unchanged\", so negative values have no documented meaning here.

eminwux:
## Summary
- The negative-gid rejection in `CreateTerminalFromProfile` (added in #189 via 11aee39) had no test coverage. Lock the behavior in: `*spec.socket.gid = -5` returns an error mentioning the field and the profile name.
- Cover the happy path too (`nil`, explicit `0`, positive int) so future refactors can't silently regress the precedence rules around "unset vs explicit-zero gid".
- Code-only test file change; no production code touched (validation already lives at `internal/profile/profile.go:114-124`).

## Test plan
- [x] `pre-commit run --files internal/profile/profile_test.go` — passes (full-tree pre-commit fails on pre-existing main drift tracked by #197 / #198, not by this change)
- [x] `make sbsh-sb` — produces ELF executable, `file ./sbsh` confirms
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `cd docs/examples/library-consumer && go build ./... && go vet ./...`

Closes #191

### DIFF
--- a/internal/profile/profile_test.go
+++ b/internal/profile/profile_test.go
@@ -22,9 +22,11 @@ import (
 	"io"
 	"log/slog"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/eminwux/sbsh/internal/errdefs"
+	"github.com/eminwux/sbsh/pkg/api"
 )
 
 func TestBuildTerminalSpec_EmptyRunPath_ReturnsErrRunPathRequired(t *testing.T) {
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #199 Review — `test(profile): cover CreateTerminalFromProfile socket gid validation`**

Tight, code-only test addition. Locks the negative-gid rejection from #189 and covers the nil/zero/positive happy paths so the "unset vs explicit-zero" pointer semantics are pinned. Helper resolves `pkg/api` types correctly, the error-substring assertions match the actual `fmt.Errorf` template at `internal/profile/profile.go:117-121`, and the cases hit `BuildTerminalSpec`'s validation branch without depending on host gid state. No production code touched, no dead symbols, no sensitive data in errors.

LGTM.

PR #201, opened 2026-05-10
fix: honor mode/gid for capture and log files

### LINKED ISSUE #200: terminal: capture/log files hard-coded to 0o600 owner-only — block kukeon-group reads via kuke log
`sbsh terminal` opens the per-terminal capture and log files at mode `0o600` owned by the listener's uid:gid, with no path to widen the mode or set a different group. Same problem class as #188 (which fixed the control socket) for the sibling files in the same `--run-path` directory: a host operator whose uid is not the listener's uid, but who *is* a member of the surrounding directory's group, can `connect()` the socket (post-#188) but still cannot `read()` the capture transcript or the log. This blocks `kuke log` end-to-end for non-root operators in the `kukeon` group whenever the host operator and the wrapped container do not share a uid (the common case: non-root operator in `kukeon` group, container with `USER claude`).

## Repro

`sbsh` ref: current main (post-#188).

Stand up a terminal as one uid and try to read its capture file as a different uid that *is* a member of the file's directory group:

```
$ sudo install -d -m 02750 -o 1000 -g 986 /tmp/sbtest
$ sudo -u '#1000' sbsh terminal \
    --run-path /tmp/sbtest \
    --socket /tmp/sbtest/socket \
    --socket-mode 0660 --socket-gid 986 \
    --capture-file /tmp/sbtest/cap \
    --log-file /tmp/sbtest/log \
    -- /bin/sh -c 'echo hello; sleep 60' &

# Different uid (uid 1001), member of gid 986:
$ id -u; id -G | tr ' ' '\n' | grep -F 986
1001
986
$ ls -l /tmp/sbtest/{socket,cap,log}
srw-rw---- 1 1000 986  0 cap-time socket    # post-#188: group-rw, OK
-rw------- 1 1000 1000 6 cap-time cap       # owner-only, blocks kukeon-group
-rw------- 1 1000 1000 0 cap-time log       # owner-only, blocks kukeon-group
$ cat /tmp/sbtest/cap
cat: /tmp/sbtest/cap: Permission denied
```

The directory is `02750 1000:986` so group members have traverse + read access; the socket is `0660` (post-#188); the capture/log files are still `0o600` and nobody outside the listener's uid can read them.

## Observed

The capture and log files are opened with mode `0o600` and never chowned to a configurable gid. There is no flag, env key, or profile field to relax the mode or set the group.

## Root cause

The capture/log file modes were hard-coded for the original single-user case. The wrapping use case (kukeon, but also any future "shared dev terminal owned by a service uid") needs to (a) widen the mode to allow group read access and (b) set the file group to a host-side group the reader is a member of. #188 closed this for the socket; the capture and log siblings stayed on the legacy contract.

## Suggested fix

Mirror #188 exactly — flags + env + profile YAML — but for the capture and log files. Same defaults-preserve-legacy-behavior contract.

### 1. New CLI flags on `sbsh terminal`

```
--capture-mode      Octal mode applied to the capture file after Open (default 0600)
--capture-gid       Numeric GID applied to the capture file via chown after Open; -1 leaves group unchanged
--log-file-mode     Octal mode applied to the log file after Open (default 0600)
--log-file-gid      Numeric GID applied to the log file via chown after Open; -1 leaves group unchanged
```

`*-mode` parsed as octal string (accept `0640` or `640`); `*-gid` defaults to `-1` (no chown). Same shape as `--socket-mode`/`--socket-gid` from #188.

### 2. New env keys (mirror `SBSH_TERM_SOCKET_MODE`/`SBSH_TERM_SOCKET_GID`)

```
SBSH_TERM_CAPTURE_MODE     sbsh.terminal.captureMode
SBSH_TERM_CAPTURE_GID      sbsh.terminal.captureGid
SBSH_TERM_LOG_FILE_MODE    sbsh.terminal.logFileMode
SBSH_TERM_LOG_FILE_GID     sbsh.terminal.logFileGid
```

### 3. Profile YAML schema — extend the `socket` precedent on `TerminalProfileSpec`

```yaml
spec:
  socket:
    mode: "0660"
    gid: 986
  capture:
    mode: "0640"  # owner rw, group r — group only needs to tail
    gid: 986
  logFile:
    mode: "0640"
    gid: 986
```

Both blocks optional; omitting either keeps the legacy `0600` owner-only contract for that artifact.

Resolution order: explicit flags win; otherwise the profile's `spec.capture` / `spec.logFile` values; otherwise defaults (`0o600`, no chown). Identical to the socket resolution from #188.

### 4. Apply at open-file time

Wherever the capture and log files are opened in `sbsh terminal` (the same code paths that consume `--capture-file` / `--log-file`), apply the resolved mode via `os.Chmod` after `OpenFile`, and the resolved gid via `os.Chown(path, -1, gid)` (leave uid unchanged) when `gid >= 0`. Apply explicitly *after* open rather than relying on the `OpenFile` mode argument, since file rotation, restarts, and pre-existing inodes all silently bypass the mode-on-create path.

Do the chown/chmod on every `Open` (including reopens after rotation), not just the first creation, so a daemon restart that re-enters the path tightens or relaxes pre-existing inodes from a prior run uniformly.

### 5. Docs

Extend `docs/profiles/README.md` with `capture:` and `logFile:` subsections paralleling the `socket:` block from #188. One-paragraph note that the default keeps owner-only access (current behavior) and that widening either field is opt-in for callers that share the terminal across uids in a group.

### 6. Backward compatibility

Existing invocations and profiles continue to produce `-rw------- listener-uid:listener-gid` capture and log files — none of the new fields are required, the env keys default to unset, the listener uses the legacy mode when no override is configured. Identical contract shape to #188's compat note.

The kukeon side then closes its own gap (eminwux/kukeon#366 successor) by passing the new flags (or writing the values into the per-cell profile) when wrapping a container.

## Out of scope

- Setting the *uid* of the capture/log files (only group/mode are configurable). Callers that want a different owner-uid pre-create the directory and run `sbsh terminal` as that uid; same as the socket case.
- Validating that the supplied gid exists on the host — the host-side caller is responsible.

## Notes

- Depends on the same flag-resolution scaffolding from #188; the change is largely additive on top of that work.
- Downstream: `eminwux/kukeon#368` (closed, wrong-architecture host-side workaround) and the successor kukeon issue that plumbs these flags via `AttachableInjection`.

eminwux:
## Summary

- Capture transcript and per-terminal log file were hard-coded to `0o600` owner-only at open time, blocking the same group-read pattern PR #189 unlocked for the control socket. A `kuke`-style log viewer running under a separate uid in the kukeon group could see the socket but still got `EACCES` on `kuke log <cell>`.
- Wires the same flag/env/profile resolution surface (explicit flag > env > profile > default) to the capture file and log file. Defaults remain `0o600` with no chown so existing deployments are byte-identical.
- Profile schema gains `spec.capture.{mode,gid}` and `spec.logFile.{mode,gid}` blocks (parallel to the existing `spec.socket` block); CLI flags `--capture-mode`, `--capture-gid`, `--log-file-mode`, `--log-file-gid` are added to both the root `sbsh` (with `--terminal-` prefix) and the `sbsh terminal` subcommand.
- Apply happens via `chmod`/`chown` after `OpenFile` rather than via the `OpenFile` mode arg, so reopens (terminal restarts) re-assert the configured mode and gid. The capture path runs inside `startPty`; the log file path runs as a new step in `StartTerminal` because `SetupFileLogger` opens the log file in the parent process before the runner has the resolved spec.
- `SocketSpec` is preserved as a `type SocketSpec = FilePermSpec` alias for backward compatibility with PR #189 consumers.

## Test plan

- [x] `make sbsh-sb` (canonical build target; produces `./sbsh` ELF + `./sb` hardlink)
- [x] `file ./sbsh` confirms ELF executable, not Go static archive
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full CI test target — package tests + integration build tag + e2e)
- [x] `pre-commit run --files \$(git diff --name-only HEAD)` (golangci-lint, gofmt, addlicense, prettier all clean)
- [x] New unit coverage in `internal/profile/profile_test.go`: capture/logFile profile blocks accepted, negative gids rejected with `spec.<block>.gid` error path, omitted blocks leave runner defaults, `applyOnePermOverride` precedence + `ErrInvalidFlag` wrapping
- [x] New unit coverage in `pkg/builder/terminal_test.go`: `WithCaptureMode`/`WithCaptureGID`/`WithLogFileMode`/`WithLogFileGID` thread through, with the same nil-vs-zero discipline as `WithSocketGID`

Closes #200

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -164,6 +164,14 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_TERM_SOCKET_GID = DefineKV("SBSH_ROOT_TERM_SOCKET_GID", "sbsh.root.terminal.socketGid")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_ROOT_TERM_CAPTURE_MODE = DefineKV("SBSH_ROOT_TERM_CAPTURE_MODE", "sbsh.root.terminal.captureMode")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_ROOT_TERM_CAPTURE_GID = DefineKV("SBSH_ROOT_TERM_CAPTURE_GID", "sbsh.root.terminal.captureGid")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_ROOT_TERM_LOG_FILE_MODE = DefineKV("SBSH_ROOT_TERM_LOG_FILE_MODE", "sbsh.root.terminal.logFileMode")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_ROOT_TERM_LOG_FILE_GID = DefineKV("SBSH_ROOT_TERM_LOG_FILE_GID", "sbsh.root.terminal.logFileGid")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_TERM_DISABLE_SET_PROMPT = DefineKV(
 		"SBSH_ROOT_TERM_DISABLE_SET_PROMPT",
 		"sbsh.root.terminal.disableSetPrompt",
@@ -192,6 +200,14 @@ var (
 	SBSH_TERM_SOCKET_MODE = DefineKV("SBSH_TERM_SOCKET_MODE", "sbsh.terminal.socketMode")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_SOCKET_GID = DefineKV("SBSH_TERM_SOCKET_GID", "sbsh.terminal.socketGid")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_TERM_CAPTURE_MODE = DefineKV("SBSH_TERM_CAPTURE_MODE", "sbsh.terminal.captureMode")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_TERM_CAPTURE_GID = DefineKV("SBSH_TERM_CAPTURE_GID", "sbsh.terminal.captureGid")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_TERM_LOG_FILE_MODE = DefineKV("SBSH_TERM_LOG_FILE_MODE", "sbsh.terminal.logFileMode")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_TERM_LOG_FILE_GID = DefineKV("SBSH_TERM_LOG_FILE_GID", "sbsh.terminal.logFileGid")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_ID = DefineKV("SBSH_TERM_ID", "sbsh.terminal.id")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -168,6 +168,10 @@ You can also use sbsh with parameters. For example:
 					SocketFile:       viper.GetString(config.SBSH_ROOT_TERM_SOCKET.ViperKey),
 					SocketMode:       viper.GetString(config.SBSH_ROOT_TERM_SOCKET_MODE.ViperKey),
 					SocketGID:        readRootSocketGID(cmd),
+					CaptureMode:      viper.GetString(config.SBSH_ROOT_TERM_CAPTURE_MODE.ViperKey),
+					CaptureGID:       readRootCaptureGID(cmd),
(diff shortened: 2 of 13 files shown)


eminwux:
**PR #201 Review — `fix: honor mode/gid for capture and log files`**

LGTM. Clean parallel extension of the PR #189 socket-perms surface to the capture transcript and per-terminal log file, using a shared `FilePermSpec` and a single `applyOnePermOverride` helper. Pointer-vs-sentinel discipline is preserved end-to-end (CLI → builder → profile → spec → runner). Public-API parity is in place via `WithCaptureMode`/`WithCaptureGID`/`WithLogFileMode`/`WithLogFileGID` in `pkg/builder`, the `SocketSpec = FilePermSpec` alias keeps PR #189 consumers compiling, and the `kuke`-style group-read use case from #200 is unlocked. No security-sensitive data in errors/logs (only paths, modes, gids — all expected). Test coverage matches the new surface: profile-block validation and builder threading are both exercised. Single commit, GPG-verified. CI all green.

A couple of small non-blocking observations:

- **Placement** `internal/terminal/terminalrunner/sockets.go:51`: `applyArtifactPerms` lives in the file named for the socket but is explicitly the non-socket helper. The doc comment is accurate so it isn't misleading; just an organizational nit if a future PR wants to extract it (e.g. `perms.go`).
- **CLI vs profile gid validation** `internal/profile/profile.go:128`: profile rejects `gid < 0` with a friendly error, but `applyOnePermOverride` passes a CLI-supplied negative gid (other than the sentinel `-1`) straight through to `os.Chown`, which surfaces as a runtime errno later. Same shape as the pre-existing socket-gid path, so not a regression — flagging only because the asymmetry will eventually be worth a single shared validator.

LGTM unconditional — flipping to `ready-to-merge`.

PR #203, opened 2026-05-11
feat(terminal): expose RPC server as public package pkg/terminal/server

### LINKED ISSUE #202: feat: expose terminal RPC server as public package (pkg/terminal/server)
## Context

`internal/terminal/terminalrpc/{codec,rpc_server}.go` and `internal/terminal/terminalrunner/` implement the JSON-RPC + SCM_RIGHTS terminal protocol that `pkg/attach` (and `kuke attach`) consumes — codec, multi-subscriber state machine, `metadata.json` write-out, PID-1 reaper, signal forwarding, multiwriter fan-out. They are unexported, so an out-of-tree binary that wants to be a drop-in replacement for `sbsh terminal` on the *server* side has to reimplement ~2000-3000 LOC against the client's wire contract.

This is the situation kukeon's `kuketty` wrapper (eminwux/kukeon#165 phase 1, #410 phase 1b) is in today. Phase 1 ships a standalone kuketty binary that creates the attach-socket inode but does not yet serve the protocol; phase 1b is meant to land the server. Reimplementing the codec + runner from scratch is large, hard to keep wire-compatible with `pkg/attach`, and creates a maintenance fork — wire format changes in sbsh would have to be mirrored manually in kuketty.

The clean fix is to expose a small public facade so `kuketty` (and any other future external wrapper) consumes sbsh's server directly. The facade hides the multi-subscriber / codec / FD-passing internals so they remain refactorable; the input is `api.TerminalSpec` (already public) and a `net.Listener`.

## Proposal

New public package `github.com/eminwux/sbsh/pkg/terminal/server` with a minimal surface:

```go
package server

import (
    \"context\"
    \"log/slog\"
    \"net\"
    \"github.com/eminwux/sbsh/pkg/api\"
)

type Server struct { /* unexported */ }

// New constructs a server bound to the given TerminalSpec. The PTY is
// spawned on Serve, not here, so a caller can listen on the socket
// before paying the fork+exec cost.
func New(spec *api.TerminalSpec, logger *slog.Logger) (*Server, error)

// Serve accepts on listener and dispatches the TerminalController
// service surface (Ping, Resize, Attach, Detach, Metadata, State,
// Stop, Write, Subscribe). Blocks until ctx is cancelled, the wrapped
// child exits, or Stop is called. Returns the terminating cause.
func (s *Server) Serve(ctx context.Context, listener net.Listener) error

// Stop initiates graceful shutdown of the wrapped child (SIGTERM, then
// SIGKILL after spec.ShutdownGrace). Safe to call from another
// goroutine while Serve blocks.
func (s *Server) Stop(reason error) error

// Metadata returns the same TerminalDoc the RPC Metadata method
// returns — exposed on the Go API so callers can introspect without
// dialing themselves.
func (s *Server) Metadata() (*api.TerminalDoc, error)
```

Implementation: thin wrapper around the existing `internal/terminal/terminalrunner.Exec` + `internal/terminal/terminalrpc.TerminalControllerRPC`. No behavior change — the unit/integration tests under those internal packages cover everything; this issue is purely a visibility move + facade.

Cut as **v0.11.0** — minor bump (new public package, no breaking change to `pkg/api`, `pkg/attach`, or `pkg/rpcclient/terminal`).

## Acceptance criteria

- [ ] `pkg/terminal/server` package with `New`, `Serve`, `Stop`, `Metadata` (signatures above; exact shape dev's call within the constraint that it stays minimal).
- [ ] Facade wraps existing `internal/terminal/terminalrunner` + `internal/terminal/terminalrpc` — no duplicated logic, no behavior change.
- [ ] Unit test in `pkg/terminal/server` that boots a server, dials with `pkg/rpcclient/terminal`, exercises a `Ping` round trip + a `Write` + a `Subscribe` (FD passing), and a graceful `Stop`. End-to-end against the public surface only.
- [ ] No new public types beyond what `pkg/api` already exposes; the facade takes and returns existing types.
- [ ] Tagged v0.11.0 once merged.

## Out of scope

- Exposing the codec or the multiwriter as their own public packages — internals stay internal.
- Changing the wire format or the method set — `kuke attach` and `pkg/attach` are unchanged consumers.
- Reaper / signal forwarding / init-mode behavior — inherited verbatim from `terminalrunner.Exec`.
- The kukeon-side adapter that maps `kuketty.Metadata` → `api.TerminalSpec` — lives in kukeon, tracked via eminwux/kukeon#410.

## Notes

- Why a facade rather than promoting `internal/terminal/terminalrunner` wholesale? The runner exports a wide surface (`Exec`, `ptyPipes`, `ioClient`, `subscriberWriter`, the events channel, the closeReqCh / closedCh handshake) that should stay refactorable. A narrow facade lets us evolve the runner without breaking external consumers — only `New/Serve/Stop/Metadata` need to stay stable.
- Why `*api.TerminalSpec` as the input rather than a kuketty-specific shape? The spec is already the producer's contract: callers stamp Command/CommandArgs/Cwd/Env/Socket{File,Mode,GID}/Capture{File,Mode,GID}/LogFile{,Mode,GID}/ShutdownGrace and hand it over. Kukeon will render its kukeond-side metadata file into a TerminalSpec via a small adapter (kukeon-side, not sbsh's concern).
- Why hide PTY spawn inside `Serve` rather than at construction? So the caller can `net.Listen` on the socket, signal readiness, *then* pay the fork+exec — matching the phase 1 sequence kukeon's `claimSocketInode` already implements.
- Downstream: eminwux/kukeon#410 re-scopes to consume this facade. Re-sizing comment + plan: https://github.com/eminwux/kukeon/issues/410#issuecomment-4423254646

eminwux:
## Summary
- New `pkg/terminal/server` facade so out-of-tree binaries (e.g. kuketty) can serve the same JSON-RPC + SCM_RIGHTS terminal protocol `pkg/attach` consumes, without reimplementing the codec, multi-subscriber state machine, PID-1 reaper, or signal forwarder.
- Minimal surface: `New(*api.TerminalSpec, *slog.Logger)`, `Serve(ctx, net.Listener)`, `Stop(error)`, `Metadata()`. PTY is spawned on `Serve` so callers can claim the socket inode before paying the fork+exec cost.
- Implementation is a thin wrapper around `internal/terminal/terminalrunner.Exec` + `internal/terminal/terminalrpc.TerminalControllerRPC`. No behavior change to the wire protocol or to in-tree consumers; an unexported `rpcAdapter` bridges the facade's `Stop(error)` to the `*api.StopArgs` shape `TerminalControllerRPC.Stop` expects.
- Adds a single new method to the unexported `TerminalRunner` interface — `UseListener(net.Listener)` — so the facade can hand a pre-bound listener to the runner instead of going through `OpenSocketCtrl`. The runner's existing `Close` continues to own listener teardown for the Serve lifetime; the caller transfers ownership.

## Test plan
- [x] `make sbsh-sb` builds the canonical executable (verified `file ./sbsh` reports ELF).
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (CI suite — cmd, internal/terminal, internal/client, integration tag, e2e).
- [x] `make test-race` (includes `./pkg/...` and the new package).
- [x] `pre-commit run --files <touched>` (golangci-lint clean).
- [x] New unit test `pkg/terminal/server.TestServer_PingWriteSubscribeStop` boots a server, dials via `pkg/rpcclient/terminal`, exercises `Ping` (round-trip), `Write` (PTY input), `Subscribe` (FD-passing + multiwriter fan-out), and a graceful `Stop`, against the public surface only.

## Notes for reviewer
- Why a facade instead of promoting `internal/terminal.Controller`? Controller owns the listener via `OpenSocketCtrl` and bundles a wider set of in-tree concerns (manager, supervisor wiring). The facade goes directly to `terminalrunner.Exec` + `terminalrpc.TerminalControllerRPC` so the surface stays minimal and the wider runner is free to evolve.
- `UseListener` was added to the runner interface (and the `Test` fake) rather than parameterizing `OpenSocketCtrl`, so existing in-tree call sites are unchanged.
- `Status.SocketFile` is left blank in the facade's metadata write (the runner's existing `OpenSocketCtrl` is what populates it). Acceptable for the kuketty use case — the caller owns the socket path and writes its own metadata file. Happy to add a setter if the reviewer prefers populating it from `listener.Addr()`.
- Tag as v0.11.0 once merged, per the issue.

Closes #202

### DIFF
--- a/internal/terminal/terminalrunner/terminal_runner.go
+++ b/internal/terminal/terminalrunner/terminal_runner.go
@@ -18,6 +18,7 @@ package terminalrunner
 
 import (
 	"context"
+	"net"
 
 	"github.com/eminwux/sbsh/internal/terminal/terminalrpc"
 	"github.com/eminwux/sbsh/pkg/api"
@@ -25,6 +26,11 @@ import (
 
 type TerminalRunner interface {
 	OpenSocketCtrl() error
+	// UseListener injects a pre-bound control-socket listener so callers
+	// that need to claim the socket inode before forking (e.g.
+	// pkg/terminal/server) can hand it over without going through
+	// OpenSocketCtrl. Must be called before StartServer.
+	UseListener(ln net.Listener)
 	StartServer(ctx context.Context, sc *terminalrpc.TerminalControllerRPC, readyCh chan error, doneCh chan error)
 	StartTerminal(evCh chan<- Event) error
 	ID() api.ID
--- a/internal/terminal/terminalrunner/terminal_runner_exec.go
+++ b/internal/terminal/terminalrunner/terminal_runner_exec.go
@@ -195,3 +195,12 @@ func (sr *Exec) childDone() bool {
 func (sr *Exec) ID() api.ID {
 	return sr.id
 }
+
+// UseListener installs an externally-bound control-socket listener so
+// that OpenSocketCtrl is not the only path that can prime the accept
+// loop. Intended for the public pkg/terminal/server facade where the
+// caller owns the inode (e.g. kuketty claims it before fork+exec).
+// Must be called before StartServer.
+func (sr *Exec) UseListener(ln net.Listener) {
+	sr.lnCtrl = ln
+}
--- a/internal/terminal/terminalrunner/terminal_runner_fake.go
+++ b/internal/terminal/terminalrunner/terminal_runner_fake.go
@@ -18,6 +18,7 @@ package terminalrunner
 
 import (
 	"context"
+	"net"
 	"sync"
 
 	"github.com/eminwux/sbsh/internal/errdefs"
@@ -28,6 +29,7 @@ import (
 type Test struct {
 	mu                  sync.RWMutex // protects function pointer fields
 	OpenSocketCtrlFunc  func() error
+	UseListenerFunc     func(ln net.Listener)
 	StartServerFunc     func(ctx context.Context, sc *terminalrpc.TerminalControllerRPC, readyCh chan error, doneCh chan error)
 	StartTerminalFunc   func(evCh chan<- Event) error
 	CloseFunc           func(reason error) error
@@ -53,6 +55,12 @@ func (sr *Test) OpenSocketCtrl() error {
 		return sr.OpenSocketCtrlFunc()
 	}
 	return nil
+}
+
+func (sr *Test) UseListener(ln net.Listener) {
+	if sr.UseListenerFunc != nil {
+		sr.UseListenerFunc(ln)
+	}
 }
 
 func (sr *Test) StartServer(
--- a/pkg/terminal/server/adapter.go
+++ b/pkg/terminal/server/adapter.go
@@ -1,0 +1,119 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package server
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/eminwux/sbsh/pkg/api"
+)
+
+// rpcAdapter satisfies api.TerminalController for the RPC layer by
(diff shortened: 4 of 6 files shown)


eminwux:
**PR #203 Review — `feat(terminal): expose RPC server as public package pkg/terminal/server`**

Clean facade: minimal surface, wraps `terminalrunner.Exec` + `terminalrpc.TerminalControllerRPC` without changing the wire protocol or in-tree callers. Listener ownership is correctly transferred to the runner (closed by `StartServer` defer + lifecycle), `UseListener` is consumed via the existing `sr.lnCtrl`, `eventBufferSize` matches `controller.go:72`'s 32, and `rpcAdapter.Stop` mirrors `Controller.Stop`'s async semantics. Stub methods on `rpcAdapter` (`Run`/`WaitReady`/`WaitClose`/`Close`) exist only to satisfy `api.TerminalController` — RPC dispatch never reaches them. No secrets or PII in error strings. Commit signed.

LGTM.

PR #208, opened 2026-05-11
fix(terminal): apply socket mode/gid on UseListener path

### LINKED ISSUE #205: bug: pkg/terminal/server UseListener bypasses OpenSocketCtrl chmod+chown of the control socket
## Context

\`pkg/terminal/server.Server\` accepts a pre-bound listener via \`runner.UseListener(ln)\` (the documented escape hatch for callers that need to claim the inode before fork+exec — e.g., kuketty). When that path is used, \`Spec.SocketMode\` and \`Spec.SocketGID\` are silently ignored: the inode keeps the umask-clipped default (typically 0o755 for unix sockets), so callers that asked for group-scoped access (e.g., 0o660 / kukeon group) end up with the wrong perms on disk.

In the OpenSocketCtrl path the chmod and chown happen right after \`Listen\` (\`internal/terminal/terminalrunner/sockets.go:125-156\`). When \`UseListener\` is used instead, that code is bypassed entirely — \`UseListener\` just stores the listener (\`internal/terminal/terminalrunner/terminal_runner_exec.go:204-206\`) and \`StartServer\` (\`internal/terminal/terminalrunner/control_server.go\`) goes straight to \`Accept\` on it.

## Reproducer

\`\`\`go
spec := &api.TerminalSpec{
    SocketFile:  sockPath,
    SocketMode:  0o660,
    SocketGID:   &kukeonGID,
    Command:     \"/bin/sh\",
    // …
}
ln, _ := net.Listen(\"unix\", sockPath)
srv, _ := server.New(spec, logger)
go srv.Serve(ctx, ln)
// stat sockPath:
//   want: srw-rw---- root:kukeon
//   got:  srwxr-xr-x root:root   ← umask-clipped Listen default; chmod/chown skipped
\`\`\`

The Spec round-trips correctly (mode and GID survive marshalling, the metadata.json on disk shows the requested values) — the gap is purely on the runtime inode.

## Where it lives

- \`internal/terminal/terminalrunner/terminal_runner_exec.go:199-206\` — \`UseListener(ln)\` only sets \`sr.lnCtrl = ln\` and does not apply perms.
- \`internal/terminal/terminalrunner/sockets.go:80-170\` — \`OpenSocketCtrl\` is the only call site that chmods + chowns the socket; it is the alternative to \`UseListener\` (\`internal/terminal/terminalrunner/terminal_runner.go:28-34\`).
- \`pkg/terminal/server/server.go:108-156\` (\`bringUp\`) — calls \`UseListener\` and then \`StartServer\`. No code path between those two ever applies \`Spec.SocketMode\` / \`Spec.SocketGID\` to the inode.
- Spec fields: \`pkg/api/types.go\` defines \`SocketMode os.FileMode\` and \`SocketGID *int\` and the builder accepts them via \`WithSocketMode\` / \`WithSocketGID\` (\`pkg/builder/terminal.go:154-168\`) — the public surface asserts these are configurable.

## Proposal

Pick one; both are small:

1. **Apply perms inside \`UseListener\`.** Move the chmod+chown lines from \`OpenSocketCtrl\` (sockets.go:125-156) into a helper, call it from both \`OpenSocketCtrl\` (right after \`Listen\`) and \`UseListener\` (right after storing the listener; resolve the inode from the listener's \`Addr().String()\`). Locks behavior in one place; out-of-tree callers don't need to know about it. Slight wrinkle: \`UseListener\` is currently a void; either return an error from it or fold it into \`StartServer\`. The current \`void\` signature plus the \"caller may listen however they want\" framing argues for the next option.

2. **Document the caller responsibility and expose a helper.** Keep \`UseListener\` minimal; add \`pkg/terminal/server.ApplySocketPerms(path, mode, gid)\` so callers replicate the OpenSocketCtrl perm dance with one call. Update \`UseListener\`'s docstring (\`terminal_runner_exec.go:199-203\`) to say \"caller is responsible for the inode's mode and group; see ApplySocketPerms.\" Smaller behavior change; keeps the \"caller owns the inode\" framing explicit.

## Acceptance criteria

- [ ] An out-of-tree caller that constructs a listener via \`net.Listen\`, hands it to \`server.New(spec, logger).Serve(ctx, ln)\` with \`spec.SocketMode = 0o660\` and \`spec.SocketGID = &gid\`, ends up with a socket inode at \`spec.SocketMode\` / \`spec.SocketGID\` on disk — without having to chmod/chown manually.
- [ ] \`pkg/terminal/server/server_test.go\` covers the case (mode + gid both set, verify on-disk perms after \`waitReady\`).
- [ ] OpenSocketCtrl path keeps current behavior — fixture for both paths.

## Out of scope

- Capture / log file perms (those use \`applyArtifactPerms\` from the OpenFile path; correct for capture, separate gap for log → tracked separately).
- Owner uid: \`Chown(-1, gid)\` only — owner is the listening uid, by design. Don't widen.

## References

- Workaround landed in kukeon: https://github.com/eminwux/kukeon/pull/413 — \`cmd/kuketty/main.go:applySocketPerms\` chmods+chowns the inode after \`claimSocketListener\` and before \`server.New(…).Serve(…)\`, with a code comment pointing at this issue.
- sbsh code locations: \`internal/terminal/terminalrunner/terminal_runner_exec.go:199-206\`, \`internal/terminal/terminalrunner/sockets.go:80-170\`, \`pkg/terminal/server/server.go:108-156\`, \`pkg/builder/terminal.go:144-168\`.

eminwux:
## Summary
- Routes both `OpenSocketCtrl` and `UseListener` through a shared `applySocketPerms` helper so the on-disk control-socket permissions are identical regardless of which path bound the listener — out-of-tree callers (e.g. kuketty) of `pkg/terminal/server` no longer have to replicate the chmod/chown dance after `net.Listen`.
- `UseListener` now returns `error`. The runner stores `lnCtrl` first and resolves the inode from `ln.Addr().String()`, so a perm-apply failure leaves the listener owned by the runner (Close tears it down) — chosen over folding into `StartServer` because it keeps the `OpenSocketCtrl` vs `UseListener` symmetry the interface already exposes.
- Picked proposal 1 from #205 over proposal 2 to keep the black-box contract: `Spec.SocketMode` / `Spec.SocketGID` always apply to the resulting inode without callers needing to know how the listener got bound.

## Test plan
- [x] `make sbsh-sb` — produced ELF executables for `sbsh` and `sb`.
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (all packages green, including the new `pkg/terminal/server.TestServer_UseListener_AppliesSocketPerms` that stats the inode after `waitReady` and asserts `Mode().Perm()` matches `0o660` and `Stat_t.Gid` matches `Spec.SocketGID`; existing `OpenSocketCtrl`-path coverage continues to pass).
- [x] `pre-commit run --files <changed files>` (trailing-whitespace, EOF, go-fmt, go-mod-tidy, golangci-lint, addlicense all pass).

Closes #205

### DIFF
--- a/internal/terminal/terminalrunner/sockets.go
+++ b/internal/terminal/terminalrunner/sockets.go
@@ -38,11 +38,48 @@ const defaultSocketMode os.FileMode = 0o600
 // defaultSocketMode for the control socket.
 const defaultArtifactMode os.FileMode = 0o600
 
+// applySocketPerms chmods (and optionally chowns the group of) the
+// control-socket inode after a listener has been bound to it. Called
+// from both OpenSocketCtrl (runner binds the socket itself) and
+// UseListener (caller hands in a pre-bound listener — pkg/terminal/server
+// facade case), so both paths land at the same on-disk permissions.
+//
+// mode == 0 falls back to defaultSocketMode so callers can pass the
+// raw spec field through without a guard. gid == nil leaves the group
+// unchanged. The chown form is Chown(path, -1, gid) so the listener's
+// uid stays the owner; only the group is rewritten.
+func (sr *Exec) applySocketPerms(path string, mode os.FileMode, gid *int) error {
+	if mode == 0 {
+		mode = defaultSocketMode
+	}
+	if errChmod := os.Chmod(path, mode); errChmod != nil {
+		sr.logger.Error(
+			"applySocketPerms: chmod failed",
+			"socket", path,
+			"mode", fmt.Sprintf("0o%o", mode),
+			"error", errChmod,
+		)
+		return fmt.Errorf("chmod socket: %w", errChmod)
+	}
+	if gid != nil {
+		if errChown := os.Chown(path, -1, *gid); errChown != nil {
+			sr.logger.Error(
+				"applySocketPerms: chown failed",
+				"socket", path,
+				"gid", *gid,
+				"error", errChown,
+			)
+			return fmt.Errorf("chown socket: %w", errChown)
+		}
+	}
+	return nil
+}
+
 // applyArtifactPerms chmods (and optionally chowns the group of) a
 // per-terminal artifact path that was just opened. Used for the capture
-// transcript and the log file — the control socket has its own helper in
-// OpenSocketCtrl because the path needs to be removed first and the
-// listener has to be torn down on failure.
+// transcript and the log file — the control socket goes through
+// applySocketPerms instead because the listener has to be torn down on
+// failure.
 //
 // mode == 0 falls back to defaultArtifactMode so callers can pass the
 // raw spec field through without a guard. gid == nil leaves the group
@@ -89,10 +126,6 @@ func (sr *Exec) OpenSocketCtrl() error {
 	sr.metadata.Status.SocketFile = socketFile
 	sr.metadataMu.Unlock()
 
-	if socketMode == 0 {
-		socketMode = defaultSocketMode
-	}
-
 	sr.logger.Debug("OpenSocketCtrl: preparing to listen", "socket", socketFile)
 	errMetadata := sr.updateMetadata()
 	if errMetadata != nil {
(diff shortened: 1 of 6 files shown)


eminwux:
**PR #208 Review — `fix(terminal): apply socket mode/gid on UseListener path`**

Tight fix for #205. The shared `applySocketPerms` collapses both bind paths to the same on-disk perm contract; out-of-tree callers no longer have to replicate the chmod/chown dance. Error path is clean (`UseListener` stores `lnCtrl` then applies perms, so `runner.Close` tears down on failure), `bringUp` properly wraps and closes on the new error, fake/interface signatures match, and the new test stats the inode after `waitReady` to lock in mode + gid against the SocketGID/SocketMode contract. No sensitive-data leaks; commit is signed; CI green.

LGTM.

PR #210, opened 2026-05-14
feat(builder): profile-free TerminalSpec construction

### LINKED ISSUE #209: builder: profile-free TerminalSpec construction (WithStages, WithPrompt, split BuildTerminalSpec/FromProfile)
## Context

`pkg/builder.BuildTerminalSpec` is the public entry point out-of-tree consumers (kukeon's `kuketty`, future SDK users) use to render an `api.TerminalSpec` they hand to `pkg/terminal/server.New(...).Serve(...)`. Today the only path that populates `Spec.Stages.OnInit`, `Spec.Stages.PostAttach`, `Spec.Prompt`, `Spec.EnvInherit`, and the shell `Cmd` / `Env` defaults is a `TerminalProfile` YAML loaded from disk via `discovery.FindProfileByNameInDir`. There is no inline option for any of those shell-shaped fields; `terminalConfig` (`pkg/builder/terminal.go:37-57`) and `BuildTerminalSpecParams` (`internal/profile/profile.go:188-226`) lack the matching plumbing.

The control flow is also coupled to the profile concept even when the caller doesn't want one:

- Empty `ProfileName` is rewritten to `"default"` at `internal/profile/profile.go:298-300`.
- `discovery.FindProfileByNameInDir` runs unconditionally at `profile.go:313`.
- On not-found, `GetDefaultHardcodedProfile` (`profile.go:435-462`) synthesizes a hardcoded `TerminalProfileDoc` with `Cmd=/bin/bash`, `Prompt="\"(sbsh-$SBSH_TERM_ID) $PS1\""`, `Stages: api.StagesSpec{}`.
- A consumer that wants neither a YAML profile nor sbsh's hardcoded default has to call `WithDisableSetPrompt(true)` purely as a safety belt to keep the hardcoded PS1 off the wire (kukeon does exactly this at `internal/controller/runner/attachable.go:213-215`).

Net: an SDK user can't construct a `TerminalSpec` from in-memory fields. They have to either write a YAML to disk and load it, or accept the hardcoded shell defaults and hide them behind `DisableSetPrompt`. This is the friction that surfaced during the kukeon Phase 4 (#290) follow-up: kukeon ends up coupled to the `TerminalProfile` concept even though all it actually needs is to stamp inline values from cell config.

## Proposal

Three changes in `pkg/builder` (and the matching plumbing in `internal/profile`):

### 1. Inline `Stages` options

Add option helpers that stamp `Spec.Stages` directly:

```go
// WithStages overlays the full StagesSpec onto the result.
func WithStages(s api.StagesSpec) TerminalOption

// Convenience for the common case — overlays into Spec.Stages.OnInit.
func WithOnInit(steps []api.ExecStep) TerminalOption

// Convenience for the symmetric case — overlays into Spec.Stages.PostAttach.
func WithPostAttach(steps []api.ExecStep) TerminalOption
```

Plumb through `BuildTerminalSpecParams.Stages` (single field; the bare `WithOnInit` / `WithPostAttach` helpers populate sub-fields of one `StagesSpec` in the config accumulator). Apply as a final overlay onto `terminalSpec.Stages` after `CreateTerminalFromProfile` + `addInputValuesToTerminal` so the option works both standalone (no profile) and as an override on top of a profile.

### 2. Inline shell-field options

Add the remaining `ShellSpec`-shaped options so a caller can build a spec from in-memory fields:

```go
func WithPrompt(prompt string) TerminalOption  // sets Spec.Prompt
func WithEnvInherit(b bool) TerminalOption     // sets Spec.EnvInherit
```

`WithCommand` / `WithCwd` / `WithEnv` / `WithDisableSetPrompt` already cover the rest of `ShellSpec`. `Spec.Kind` is hardcoded to `TerminalLocal` in `CreateTerminalFromProfile` regardless of `RunTarget`, so no new option needed there until a non-local target lands.

Same overlay semantics as Stages: apply after the profile-derived `terminalSpec` is built so the option works in both lanes.

### 3. Split `BuildTerminalSpec` into two entry points (cleaner separation)

Today's single `BuildTerminalSpec` conflates two lanes. Replace with:

```go
// BuildTerminalSpec constructs a TerminalSpec from inline options only.
// No profile YAML is loaded; no hardcoded \"default\" profile fallback runs.
// Stages, Prompt, Cmd, Env, etc. come from the With* options or stay zero.
// runPath is still required.
func BuildTerminalSpec(
    ctx context.Context, logger *slog.Logger, runPath string,
    opts ...TerminalOption,
) (*api.TerminalSpec, error)

// BuildTerminalSpecFromProfile preserves today's profile-loader behavior:
// resolves ProfileName (defaulting to \"default\"), runs disk discovery via
// pkg/discovery, falls back to GetDefaultHardcodedProfile on not-found for
// the \"default\" name. Inline options overlay onto the profile-derived spec.
func BuildTerminalSpecFromProfile(
    ctx context.Context, logger *slog.Logger, runPath string,
    opts ...TerminalOption,
) (*api.TerminalSpec, error)
```

`WithProfile` / `WithProfilesDir` move to options only the `FromProfile` variant honors (a panic / errdefs.InvalidOption from the inline `BuildTerminalSpec` if a caller passes them by mistake — fail loud rather than silently ignore). The internal `profile.BuildTerminalSpec` keeps both code paths but exits before discovery for the inline lane.

This is the breaking-change piece of the issue:

- In-tree callers: `cmd/sbsh/...`, profile-driven test fixtures, anything that builds a spec with no explicit profile name today (relies on the implicit `"default"` resolution) → migrate to `BuildTerminalSpecFromProfile(...)`.
- Out-of-tree callers (kukeon's `kuketty`): switch to the new inline `BuildTerminalSpec(...)`, drop the `WithDisableSetPrompt` safety belt, drop `WithProfile` / `WithProfilesDir`.

A short migration note in the release notes covers both audiences. The split is one commit; in-tree migrations are mechanical.

## Acceptance criteria

- [ ] `pkg/builder` exports `WithStages(api.StagesSpec)`, `WithOnInit([]api.ExecStep)`, `WithPostAttach([]api.ExecStep)`, `WithPrompt(string)`, `WithEnvInherit(bool)`. Each is plumbed end-to-end into the rendered `api.TerminalSpec` (Stages / Prompt / EnvInherit fields).
- [ ] `BuildTerminalSpec(ctx, logger, runPath, opts...)` builds a spec from inline options only. No `pkg/discovery` call, no `GetDefaultHardcodedProfile` fallback, no implicit `\"default\"` profile resolution. Returns a spec with zero-valued Stages/Prompt/Cmd/etc. when no corresponding `With*` option is passed (except `Cmd`/`CmdArgs` which can default to `/bin/bash -i` to preserve a sensible runner default, or stay zero — design call within the PR).
- [ ] `BuildTerminalSpecFromProfile(ctx, logger, runPath, opts...)` retains today's behavior: resolves `ProfileName` (defaulting to `\"default\"`), runs disk discovery, falls back to `GetDefaultHardcodedProfile` on not-found for the `\"default\"` name only. Inline options overlay onto the profile-derived spec (matching today's `addInputValuesToTerminal` precedence).
- [ ] Passing `WithProfile` / `WithProfilesDir` to the inline `BuildTerminalSpec` returns an `errdefs`-class error (or panics if the convention is that builder misuse is a programmer bug). Fail loud, do not silently ignore.
- [ ] In-tree call sites (`cmd/sbsh`, tests, anything in `internal/` that invokes the builder without `WithProfile`) are migrated to whichever variant matches their intent. CI green.
- [ ] Release notes / `CHANGELOG.md` entry calling out the split, the new options, and the migration path for both lanes (`s/BuildTerminalSpec/BuildTerminalSpecFromProfile/` for in-tree profile users; new inline lane for out-of-tree SDK users).
- [ ] Unit tests in `pkg/builder/` (and/or `internal/profile/`) cover: (a) each new `With*` option lands in `Spec` for the inline lane; (b) each new `With*` option overlays correctly on top of a profile-derived spec in the `FromProfile` lane; (c) the inline lane never touches disk; (d) the inline lane rejects `WithProfile` / `WithProfilesDir`.

## Out of scope

- Reworking `TerminalProfile` YAML schema or `pkg/discovery` loader — they stay as-is; only the *use* of them in the builder split changes.
- Adding non-local `Spec.Kind` values (a `WithKind(api.TerminalKind)` option) — `CreateTerminalFromProfile` hardcodes `TerminalLocal` today and changing that is a separate concern.
- Changing `pkg/terminal/server` consumption — it already reads whatever `Spec.Stages.OnInit` / `Spec.Prompt` arrive; no server-side change needed.
- The kukeon-side renderer fix that uses the new options (separate kukeon issue once these land; tracked alongside eminwux/kukeon#493 for the test-coverage gap).

## References

- Friction surfaced while investigating kukeon's `kuketty` OnInit drop: eminwux/kukeon#493 (test-coverage gap that let it ship; renderer fix is a follow-up there).
- Renderer that today is forced through the profile lane: `internal/controller/runner/attachable.go::writeKukettyMetadata` in eminwux/kukeon, specifically the `WithDisableSetPrompt(true)` safety belt at lines 213-215.
- Builder option set today: \`pkg/builder/terminal.go:30-215\`.
- Profile-coupled flow today: \`internal/profile/profile.go:238-354\` (defaults `ProfileName="default"` at 298-300; discovery at 313; hardcoded fallback at 336).
- Stages overlay point in today's flow: \`internal/profile/profile.go:344-348\` — `CreateTerminalFromProfile` is the only writer of `terminalSpec.Stages`.

eminwux:
## Summary

Splits `pkg/builder.BuildTerminalSpec` into two entry points so SDK consumers can build a `TerminalSpec` from in-memory fields, without writing a YAML profile to disk or accepting the hardcoded `"default"` profile fallback.

- `BuildTerminalSpec(ctx, logger, runPath, opts...)` — **inline-only**. No `pkg/discovery`, no implicit `"default"` resolution, no `GetDefaultHardcodedProfile`. Stages / Prompt / EnvInherit stay zero unless the matching `With*` option is passed; Cmd / CmdArgs default to `/bin/bash -i` to preserve a sensible runner default.
- `BuildTerminalSpecFromProfile(ctx, logger, runPath, opts...)` — legacy profile path. Empty `WithProfile` resolves to `"default"`, runs disk discovery, falls back to `GetDefaultHardcodedProfile` for the `"default"` name only. Inline `With*` options overlay onto the profile-derived spec.
- New options: `WithStages`, `WithOnInit`, `WithPostAttach`, `WithPrompt`, `WithEnvInherit`. All five are plumbed end-to-end into `Spec.Stages` / `Spec.Prompt` / `Spec.EnvInherit`, on both lanes.
- Passing `WithProfile` or `WithProfilesDir` to the inline lane returns `errdefs.ErrInvalidOption` — fail loud, never silently ignore.

`cmd/sbsh` and `cmd/sbsh/terminal` migrated to `BuildTerminalSpecFromProfile`. CHANGELOG.md created with migration notes for both audiences.

## Test plan

- [x] `make sbsh-sb` — produces ELF binaries `sbsh` and `sb`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full unit suite (mirrors CI)
- [x] `cd docs/examples/library-consumer && go build ./... && go vet ./...` — out-of-tree consumer compiles unchanged on the inline lane
- [x] `pre-commit run --files ...` — fmt / golangci-lint / addlicense / prettier all clean
- [ ] `make test-e2e` — not run; e2e suite excluded by both `make test` and CI's unit job (`grep -v '/e2e$'`), and exercises the same CLI entry points covered by the migrated unit tests

Closes #209

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,46 @@
+# Changelog
+
+All notable user-visible changes to sbsh are recorded here. The format
+loosely follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+## Unreleased
+
+### Added
+
+- `pkg/builder` gained five new `TerminalOption` helpers so SDK
+  consumers can populate every shell-shaped field of a `TerminalSpec`
+  without writing a YAML profile to disk:
+  - `WithStages(api.StagesSpec)` overlays the full lifecycle stages.
+  - `WithOnInit([]api.ExecStep)` and `WithPostAttach([]api.ExecStep)`
+    overlay only their respective sub-fields.
+  - `WithPrompt(string)` stamps `Spec.Prompt`.
+  - `WithEnvInherit(bool)` stamps `Spec.EnvInherit` (set-sentinel
+    discipline; `false` is a valid distinct value from "unset").
+- `pkg/builder.BuildTerminalSpecFromProfile` preserves the legacy
+  profile-driven build path (empty `WithProfile` resolves to
+  `"default"`, `pkg/discovery` lookup, hardcoded-default fallback for
+  the `"default"` name only).
+
+### Changed
+
+- **Breaking:** `pkg/builder.BuildTerminalSpec` now builds a
+  `TerminalSpec` from inline options only. It no longer calls
+  `pkg/discovery`, no longer resolves an implicit `"default"`
+  profile, and no longer falls back to the hardcoded default profile.
+  Stages / Prompt / EnvInherit stay zero unless the matching `With*`
+  option is passed; Cmd / CmdArgs default to `/bin/bash -i` so the
+  runner has a sensible shell.
+- **Breaking:** Passing `WithProfile` or `WithProfilesDir` to the
+  inline `BuildTerminalSpec` returns `errdefs.ErrInvalidOption`. Use
+  `BuildTerminalSpecFromProfile` instead.
+
+### Migration
+
+- In-tree CLI / test callers that previously relied on the implicit
+  profile resolution should switch to `BuildTerminalSpecFromProfile`.
+  The `cmd/sbsh` entry points have already been migrated.
+- Out-of-tree SDK callers (e.g., kukeon's `kuketty`) should keep using
+  `BuildTerminalSpec` for the inline lane and adopt `WithPrompt` /
+  `WithStages` / `WithOnInit` / `WithPostAttach` / `WithEnvInherit`
+  instead of relying on `WithDisableSetPrompt(true)` as a safety belt
+  against the old hardcoded prompt.
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -152,7 +152,7 @@ You can also use sbsh with parameters. For example:
 				)
 			}
 
-			terminalSpec, buildErr := profile.BuildTerminalSpec(
+			terminalSpec, buildErr := profile.BuildTerminalSpecFromProfile(
 				cmd.Context(),
 				logger,
 				&profile.BuildTerminalSpecParams{
--- a/cmd/sbsh/terminal/terminal.go
+++ b/cmd/sbsh/terminal/terminal.go
@@ -140,7 +140,7 @@ func buildTerminalSpecFromFlags(
 		terminalCmdArgs = workload[1:]
 	}
 
-	spec, buildErr := profile.BuildTerminalSpec(
+	spec, buildErr := profile.BuildTerminalSpecFromProfile(
 		cmd.Context(),
 		logger,
 		&profile.BuildTerminalSpecParams{
--- a/docs/site/reference/library.md
+++ b/docs/site/reference/library.md
@@ -12,15 +12,15 @@ A complete, build-tested example lives at
 
 The public library lives under [`github.com/eminwux/sbsh/pkg`](https://pkg.go.dev/github.com/eminwux/sbsh/pkg):
 
-| Package                  | Purpose                                                                                                                                                                                                                                                 |
-| ------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `pkg/api`                | Declarative document types (`TerminalDoc`, `TerminalSpec`, `ClientDoc`, `TerminalProfileDoc`, …) and RPC payload types (`PingMessage`, `WriteRequest`, `SubscribeRequest`, `StopArgs`, …). Counterpart of the YAML/JSON manifests.                      |
-| `pkg/builder`            | Construct a `TerminalSpec` or `ClientDoc` from a profile name plus inline overrides (`WithProfile`, `WithProfilesDir`, `WithCommand`, `WithEnv`, `WithClientMode`, …). No YAML round-trip.                                                              |
-| `pkg/spawn`              | Launch detached `Terminal` and `Client` subprocesses (`NewTerminal`, `NewClient`). Returns `TerminalHandle` / `ClientHandle` with `PID`, `SocketPath`, `WaitReady`, `WaitClose`, `Close`.                                                               |
-| `pkg/discovery`          | Enumerate live terminals/clients or look them up by ID/Name under a run path (`ScanTerminals`, `FindTerminalByID`, `FindTerminalByName`, `ScanClients`, `FindClientByName`). Load profiles from disk (`LoadProfilesFromDir`, `FindProfileByNameInDir`). |
-| `pkg/rpcclient/terminal` | JSON-RPC client for the `TerminalController` socket (`Ping`, `Resize`, `Attach`, `Detach`, `Metadata`, `State`, `Stop`, `Write`, `Subscribe`).                                                                                                          |
-| `pkg/rpcclient/client`   | JSON-RPC client for the `ClientController` socket (`Ping`, `Metadata`, `State`, `Stop`, `Detach`).                                                                                                                                                      |
-| `pkg/errors`             | Curated error sentinels re-exported from `internal/errdefs`. Use `errors.Is` to branch on well-known failure modes without importing internal packages.                                                                                                 |
+| Package                  | Purpose                                                                                                                                                                                                                                                                                                                                                            |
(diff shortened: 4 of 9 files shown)


eminwux:
**PR #210 Review — `feat(builder): profile-free TerminalSpec construction`**

Clean lane split: `BuildTerminalSpec` is inline-only, `BuildTerminalSpecFromProfile` retains the legacy profile lookup, and the five new `With*` options thread end-to-end via a single `paramsFromConfig` mapping. The Overlay/Set-sentinel discipline is correct, `WithOnInit`/`WithPostAttach` intentionally clear `stagesOverlay` to avoid silently wiping the sibling sub-field, and `WithProfile`/`WithProfilesDir` rejection wraps `errdefs.ErrInvalidOption` so `errors.Is` works. No secret/PII leaks. Tests cover both lanes including the precedence between `WithStages` and `WithOnInit`/`WithPostAttach`, the inline `EnvInherit=false` default (called out as a behavior change in the CHANGELOG migration note), and the `ErrInvalidOption` rejection path. Single commit, verified signature.

One non-blocking item:

- **Doc staleness** `docs/site/reference/library.md:18`: the `pkg/builder` row still describes `WithProfile`/`WithProfilesDir` as inline overrides on `BuildTerminalSpec`. Post-PR those are rejected on that lane; the row should mention `BuildTerminalSpecFromProfile` and the new `WithStages`/`WithOnInit`/`WithPostAttach`/`WithPrompt`/`WithEnvInherit` options. Will file as a followup-nit.

LGTM.

eminwux:
Addressed in dac4854:
- Updated `docs/site/reference/library.md` so the `pkg/builder` row names both `BuildTerminalSpec` (inline-only) and `BuildTerminalSpecFromProfile`, and the new `WithStages` / `WithOnInit` / `WithPostAttach` / `WithPrompt` / `WithEnvInherit` options. Also reworked Walkthrough §1 to show both lanes with separate examples, since it had the same staleness.

eminwux:
Checked dac4854 (delta from a9e5d6d):
- Doc staleness in `docs/site/reference/library.md` (non-blocking nit from prior review): addressed — the `pkg/builder` row now splits the lane description, the walk-through section adds a `BuildTerminalSpecFromProfile` example, and all five new options (`WithStages`/`WithOnInit`/`WithPostAttach`/`WithPrompt`/`WithEnvInherit`) are named. New commit is signed.

Still LGTM.

PR #213, opened 2026-05-17
fix(terminalrpc): close sender-side fds after SCM_RIGHTS send

### LINKED ISSUE #212: kuketty fanout deadlocks after a few attach/detach cycles: cliFD leaked on SCM_RIGHTS send
## Summary

`internal/terminal/terminalrpc/codec.go:WriteResponse` ships per-client file descriptors via `SCM_RIGHTS` using `WriteMsgUnix`, but **never closes the sender's copy** after the send. The fd it sends is one end of a `socketpair(2)` whose other end is wrapped in the `net.Conn` the PTY fanout writes to. Every successful `Attach` (and every `Subscribe`) leaks one local fd into the terminal process. Once the kernel send buffer for that leaked socketpair fills (~208 KB), the fanout producer blocks, every live attacher stops receiving output, and `metadata.json.status.attachers` accumulates ghost entries that are never cleaned up.

In practice the terminal becomes unresponsive after a small number of attach/detach cycles. Reproduces against `sbsh` standalone and against kukeon cells (which embed kuketty as PID 1 of the work container).

## Repro

1. Start a terminal (any sbsh / kuketty instance with an interactive program in the work shell).
2. `kuke attach <cell>` (or `sb attach …`), then kill the attach client without a clean detach — `kill -9` the attach pid, or close the host pts. Repeat 2–4 times.
3. New attaches connect, register in `attachers`, and receive **nothing**. The terminal looks frozen but the PTY process is still alive.

## Diagnostic evidence (from a wedged kukeon cell)

The terminal process's open sockets, inspected from inside its netns:

```
ss -xnp
u_str ESTAB 0      80102   * 640795 * 640794  users:(("kuketty",pid=865699,fd=13))
u_str ESTAB 0      215296  * 640794 * 640795  users:(("kuketty",pid=865699,fd=14))
u_str ESTAB 0      0       * 923852 * 0       users:(("kuketty",pid=865699,fd=21))
u_str ESTAB 0      0       * 918010 * 0       users:(("kuketty",pid=865699,fd=24))
u_str ESTAB 0      0       * 1005805 * 1005804 users:(("kuke",pid=…,fd=7),("kuketty",pid=865699,fd=25))
```

- `fd 13 ↔ fd 14` — leaked socketpair: **both ends in kuketty**, Send-Q saturated in both directions (~80 KB / ~215 KB).
- `fd 21`, `fd 24` — sockets whose remote peer fd is gone (peer inode = 0) but kuketty still holds its local side. These correspond 1:1 with the ghost entries in `metadata.json.status.attachers`.
- `fd 25` — current live `kuke attach`. Send-Q is 0 because the fanout producer is already wedged upstream and never reaches this sink.

`metadata.json.status.attachers` for the wedged terminal kept rotating only its newest slot — older entries (`08a1e112`, `4c04b3b6` in this trace) stuck forever because `handleClient` never observes EOF on the conn (the leaked local cliFD prevents it).

## Root cause

`internal/terminal/terminalrpc/codec.go` lines 158–178:

```go
if r, ok := body.(*api.ResponseWithFD); ok && len(r.FDs) > 0 {
    ...
    rights := unix.UnixRights(r.FDs...)
    if _, _, err := c.uc.WriteMsgUnix(buf.Bytes(), rights, nil); err != nil {
        return err
    }
    return nil    // ← r.FDs are still open in this process
}
```

`unix.UnixRights` + `WriteMsgUnix` causes the kernel to **dup** the fd into the receiver's fd table. The sender's fd is unchanged — the SCM_RIGHTS contract is that the sender must `close(2)` its copy. Nothing here does.

Producer (`Attach` / `Subscribe`) creates a `socketpair`:
- `srvFD` → wrapped in `net.FileConn` → used by the fanout writer / dualcopier
- `cliFD` → returned in `response.FDs` for SCM_RIGHTS shipping → **leaked**

Consequences:
1. PTY output flows into the conn wrapping `srvFD`; the kernel queues bytes toward the peer (the leaked `cliFD`), which nothing in the terminal process ever reads. Buffer fills, `Write` blocks, the fanout multiwriter stalls.
2. When the remote `kuke attach`/`sb attach` dies, the receiver's dup is reaped but kuketty's local `cliFD` is still open. The conn never sees EOF, so the dualcopier never returns, so `cleanupClient` never runs, so `updateTerminalAttachers` never removes the dead client — ghost attachers persist in `metadata.json` until the terminal process exits.

`internal/terminal/terminalrunner/subscribe.go` closes `cliFD` only on the two error paths (lines 84 and 100) — the success path at line 124 has the same leak.

## Suggested fix

Close the local fds (success **and** failure) in the codec — one fix covers `Attach` and `Subscribe`:

```go
// internal/terminal/terminalrpc/codec.go, around line 158
if r, ok := body.(*api.ResponseWithFD); ok && len(r.FDs) > 0 {
    fds := r.FDs
    r.FDs = nil // make idempotent
    defer func() {
        for _, fd := range fds {
            _ = unix.Close(fd)
        }
    }()
    ...
    rights := unix.UnixRights(fds...)
    if _, _, err := c.uc.WriteMsgUnix(buf.Bytes(), rights, nil); err != nil {
        return err
    }
    return nil
}
```

The `defer` matters — if `json.Encode` errors before the send, today's code would also leak.

## Regression test

A unit/integration test that does N attach+detach cycles against a live terminal and asserts:
- `metadata.json.status.attachers` ends empty,
- `len(sr.clients) == 0`,
- no fds in the terminal process have only the local side of a `(AF_UNIX, SOCK_STREAM)` socketpair held.

The existing e2e suite attaches once per case, so the leak survives every CI run.

## Adjacent finding (not strictly this bug, worth same PR)

`subscribe.go` lines 91–93 do `uc.CloseRead()` to keep a stuck subscriber from filling the reverse-direction kernel buffer; the forward direction is then drained by `subscriberWriter`'s bounded ring + per-write `SetWriteDeadline`. The interactive `Attach` path in `connections.go` has **neither**: `dualcopier.RunCopier(client.pipeOutR, client.conn, …)` writes raw to `client.conn` with no deadline. After the SCM_RIGHTS fix, a single live attacher whose receiver is genuinely paused (suspended SSH, blocked terminal emulator, etc.) will still head-of-line-block the fanout. Bringing the interactive path up to feature parity with the subscriber path would close the remaining hazard.

eminwux:
## Summary

- `internal/terminal/terminalrpc/codec.go` ships per-client fds via `WriteMsgUnix`/`UnixRights` but never closed the sender's local copy. SCM_RIGHTS dups the fd into the receiver's table; the sender owns its originals and must close them, or the kernel keeps the underlying socketpair half open forever. Without the close, every successful `Attach` (and every `Subscribe`) leaks one fd in the terminal process, which eventually wedges the PTY fanout once the leaked send buffer fills.
- Fix is the one suggested in issue #212: close the local fds in a `defer` so success and pre-send marshal errors both clean up, and nil out `r.FDs` for idempotency. One change covers both `Attach` and `Subscribe`.
- Regression test: `pkg/terminal/server/server_attach_fdleak_linux_test.go` drives 20 `Attach`+`Detach` cycles through the public facade and asserts that the count of socket-typed entries in `/proc/self/fd` does not grow (verified to fail with +20 sockets when the fix is reverted). Counts sockets only because the runner has unrelated pipe churn per Attach that would otherwise dominate the signal (see follow-up below).
- Scope notes:
  - The issue body's "adjacent finding" (no `SetWriteDeadline` on the interactive `Attach` path) is intentionally out of scope here: fixing it would change semantics for `dualcopier.RunCopier`'s second caller (`internal/client/clientrunner`), so it deserves its own design. Filed as a follow-up before this PR merges.
  - During investigation I discovered a separate pre-existing leak: `cleanupClient` never closes the `pipeOutR`/`pipeOutW` pair from `setupPipes`, so each `Attach`/`Detach` cycle leaks two pipe fds. Distinct from the SCM_RIGHTS bug and outside the test's signal — filed as a follow-up before this PR merges.

## Test plan

- [x] `make sbsh-sb` (canonical build produces ELF for both `sbsh` and `sb`)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full project test target — cmd, internal, e2e)
- [x] `go test ./pkg/...` (includes the new `TestServer_AttachDetachCycles_NoFDLeak`)
- [x] Regression test verified to fail with the codec fix reverted (+20 socket fds over 20 cycles)
- [x] `pre-commit run --files` on changed files

Closes #212

### DIFF
--- a/internal/terminal/terminalrpc/codec.go
+++ b/internal/terminal/terminalrpc/codec.go
@@ -155,9 +155,21 @@ func (c *unixJSONServerCodec) WriteResponse(resp *rpc.Response, body interface{}
 		}
 	}
 
-	// FD-passing path
+	// FD-passing path. SCM_RIGHTS dup's each fd into the receiver's table;
+	// the sender owns its originals and must close them or the kernel keeps
+	// the underlying socketpair half open forever. Close on every exit path
+	// (success or pre-send marshal error) and clear r.FDs so a retry can't
+	// double-close.
 	if r, ok := body.(*api.ResponseWithFD); ok && len(r.FDs) > 0 {
-		c.logger.Info("Attach codec response", "json_payload", r.JSON, "fds", r.FDs)
+		fds := r.FDs
+		r.FDs = nil
+		defer func() {
+			for _, fd := range fds {
+				_ = unix.Close(fd)
+			}
+		}()
+
+		c.logger.Info("Attach codec response", "json_payload", r.JSON, "fds", fds)
 
 		wire := struct {
 			ID     any         `json:"id"`
@@ -170,7 +182,7 @@ func (c *unixJSONServerCodec) WriteResponse(resp *rpc.Response, body interface{}
 			return err
 		}
 
-		rights := unix.UnixRights(r.FDs...)
+		rights := unix.UnixRights(fds...)
 		if _, _, err := c.uc.WriteMsgUnix(buf.Bytes(), rights, nil); err != nil {
 			return err
 		}
--- a/pkg/terminal/server/server_attach_fdleak_linux_test.go
+++ b/pkg/terminal/server/server_attach_fdleak_linux_test.go
@@ -1,0 +1,142 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build linux
+
+package server_test
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/eminwux/sbsh/pkg/api"
+	rpcclient "github.com/eminwux/sbsh/pkg/rpcclient/terminal"
+)
+
+// TestServer_AttachDetachCycles_NoFDLeak locks in the fix for issue #212:
+// the JSON-RPC codec used to leak the sender-side end of the SCM_RIGHTS
+// socketpair on every Attach (and Subscribe). Each leaked fd kept the
+// peer's send buffer alive in the terminal process, eventually wedging
+// the PTY fanout once the buffer filled (~208 KiB).
+//
+// The test counts only socket-typed fds (`/proc/self/fd` entries whose
+// readlink starts with `socket:`) — the runner has unrelated pipe churn
+// per Attach that would otherwise dwarf the signal. Without the codec
+// fix every Attach leaks one socketpair fd; with it the count is flat
+// across cycles.
+func TestServer_AttachDetachCycles_NoFDLeak(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+	defer cancel()
+
+	h := startTestServer(ctx, t)
+	defer h.cleanup()
+
+	client := rpcclient.NewUnix(h.socketPath, h.logger)
+	defer client.Close()
+
+	// One warm-up cycle to surface any one-shot allocations (lazy
+	// pipes, goroutines holding read-side fds) before the baseline.
+	if err := attachDetachOnce(ctx, client, "warmup"); err != nil {
+		t.Fatalf("warmup attach/detach: %v", err)
+	}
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #213 Review — `fix(terminalrpc): close sender-side fds after SCM_RIGHTS send`**

LGTM. The fix is minimal and correctly placed: `defer` fires after `WriteMsgUnix` returns, so the kernel has either already dup'd the fds (success) or never did (error) — closing the sender's originals is safe on both paths. Nil-ing `r.FDs` before the defer makes a hypothetical retry idempotent. The WHY comment explains the SCM_RIGHTS dup semantics, which is the non-obvious bit. Regression test verified to fail with the fix reverted (+20 sockets/20 cycles) and counts only socket-typed `/proc/self/fd` entries to dodge the runner's pipe churn — sound test design. No sensitive-data regressions (the `fds` log line existed pre-PR). All four commits GPG-verified; CI all green.

Two adjacent leaks (no `SetWriteDeadline` on interactive `Attach`; `cleanupClient` not closing `pipeOutR`/`pipeOutW`) are out of scope and already filed as follow-ups per the PR body.

LGTM to merge.

PR #216, opened 2026-05-17
fix(terminalrunner): close client pipe fds on cleanup and detach

### LINKED ISSUE #215: Pipe fds leaked on every Attach/Detach cycle — setupPipes creates pipe never closed in cleanupClient
## Context

Distinct from issue #212 (the SCM_RIGHTS socket leak fixed by PR #213). Discovered while writing the regression test for #212.

`internal/terminal/terminalrunner/io.go:23-36` (`setupPipes`) calls `os.Pipe()` to allocate the `pipeOutR`/`pipeOutW` pair stored on the `ioClient` and added to `multiOutW`:

```go
client.pipeOutR, client.pipeOutW, err = os.Pipe()
...
multiOutW.Add(client.pipeOutW)
```

`internal/terminal/terminalrunner/connections.go:26-48` (`cleanupClient`) removes `client.pipeOutW` from the multiwriter, closes `client.conn`, removes the client from the map — but **never closes `client.pipeOutR` or `client.pipeOutW`**. Both ends of the pipe stay open in the terminal process for the lifetime of the process.

This is observable as +2 fds per `Attach`/`Detach` cycle, on top of (and unrelated to) the SCM_RIGHTS leak fixed in PR #213. PR #213's regression test (`pkg/terminal/server/server_attach_fdleak_linux_test.go`) explicitly counts only socket-typed fds for this reason — without that filter the pipe leak would dwarf the SCM_RIGHTS signal it was measuring.

Repro (against PR #213's tree, so the SCM_RIGHTS leak is already gone):

```
TestServer_AttachDetachCycles_NoFDLeak with counting changed back to "all fds":
  open fd count grew by 40 over 20 attach/detach cycles
  → 2 pipe fds per cycle, all `pipe:[<inode>]` in /proc/self/fd
```

## Proposal

Close both pipe ends in `cleanupClient` after the pipe is removed from `multiOutW`. Order matters: `multiOutW.Remove(client.pipeOutW)` first so the fan-out writer goroutine doesn't try to write to a closed pipe end (the existing teardown order already gets this right for `client.conn`).

```go
// cleanupClient, after multiOutW.Remove(client.pipeOutW):
if client.pipeOutW != nil {
    _ = client.pipeOutW.Close()
}
if client.pipeOutR != nil {
    _ = client.pipeOutR.Close()
}
```

The dualcopier writer goroutine in `handleClient` does `RunCopier(client.pipeOutR, client.conn, …)`. Closing `pipeOutR` will unblock its `Read` with an error and let the goroutine exit cleanly. `pipeOutW` close has no live writer once `multiOutW.Remove` returned, so it's a clean release.

## Acceptance criteria

- [ ] `cleanupClient` closes both `client.pipeOutR` and `client.pipeOutW`.
- [ ] Repeated `Attach`/`Detach` cycles do not grow the pipe-fd count in the terminal process (a regression test like `pkg/terminal/server/server_attach_fdleak_linux_test.go` but counting all fds, or pipe-typed fds specifically).
- [ ] No regressions in existing tests (`make test`, `go test ./pkg/...`).
- [ ] The dualcopier writer goroutine for the detached client terminates after `cleanupClient` runs (no leaked goroutines on the multiwriter side).

## Out of scope

- The SCM_RIGHTS leak (fixed in PR #213).
- The interactive-Attach head-of-line blocking adjacent finding (#214).
- Any refactor of `ioClient` to RAII-style ownership of its fds.

## References

- PR #213 — closes #212 (the SCM_RIGHTS socket leak), references this issue in its body.
- `internal/terminal/terminalrunner/io.go:23-36` (`setupPipes`) — pipe allocation.
- `internal/terminal/terminalrunner/connections.go:26-48` (`cleanupClient`) — site of the missing closes.
- `pkg/terminal/server/server_attach_fdleak_linux_test.go` — existing harness that can be adapted for the regression test.

eminwux:
## Summary
- `setupPipes` allocates an `os.Pipe` pair per Attach and stashes it on the `ioClient`; neither end was ever closed, so each Attach/Detach cycle leaked two pipe fds in the runner process (issue #215). PR #213's SCM_RIGHTS regression test counted only socket-typed fds for exactly this reason — without that filter the pipe leak would have dwarfed the socket signal it was measuring.
- Add the close to `cleanupClient` (matching the issue's proposed location) **and** to `Detach`. The issue's proposal targets `cleanupClient` alone, but in the current code path `CopierManager` only invokes its `finish()` callback on `ctx.Done()` (terminal shutdown) — so `cleanupClient` never fires on a normal `Detach` and a `cleanupClient`-only fix would not close the per-cycle leak the issue measures or unblock the writer-copier goroutine on the multiwriter side (AC #4). Treating this as the "premise rotted but intent inferable" case per `dev/CLAUDE.md`: proceeding with the broader scope and documenting it here. Reviewer push-back welcome if the preferred fix is instead to make `Detach` route through `cleanupClient`, or to fix `CopierManager`'s errgroup branch to also call `finish()`.
- Order in both call sites matches the proposal: `multiOutW.Remove(pipeOutW)` first, then close `pipeOutW` (let any buffered bytes drain to the writer copier), then `pipeOutR` (unblocks the writer's `Read` with EOF so its goroutine exits).

## Test plan
- [x] `make sbsh-sb` — produces ELF `sbsh` and hardlinks `sb`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full CI unit-test surface
- [x] `make test` — includes integration + e2e
- [x] `pre-commit run --files <changed files>` — go-fmt, golangci-lint, addlicense all pass
- [x] New regression test `TestServer_AttachDetachCycles_NoPipeFDLeak` in `pkg/terminal/server/server_attach_pipefdleak_linux_test.go` counts pipe-typed fds across 20 Attach/Detach cycles; flat with the fix, +40 without (matches the issue's repro number)
- [x] Existing `TestServer_AttachDetachCycles_NoFDLeak` (SCM_RIGHTS regression) still passes — the two tests stay independent (socket-typed vs pipe-typed fd counting)

Closes #215

### DIFF
--- a/internal/terminal/terminalrunner/connections.go
+++ b/internal/terminal/terminalrunner/connections.go
@@ -38,6 +38,21 @@ func (sr *Exec) cleanupClient(client *ioClient) {
 		multiOutW.Remove(client.pipeOutW)
 	}
 
+	// Release the per-client pipe fds; otherwise each Attach/Detach cycle
+	// leaks both ends in the runner process. Close pipeOutW first so any
+	// bytes still buffered in the pipe can drain to the writer copier,
+	// then pipeOutR so its Read unblocks with EOF and the goroutine exits.
+	if client.pipeOutW != nil {
+		if perr := client.pipeOutW.Close(); perr != nil {
+			sr.logger.Debug("error closing client pipeOutW", "err", perr, "client", client.id)
+		}
+	}
+	if client.pipeOutR != nil {
+		if perr := client.pipeOutR.Close(); perr != nil {
+			sr.logger.Debug("error closing client pipeOutR", "err", perr, "client", client.id)
+		}
+	}
+
 	if cerr := client.conn.Close(); cerr != nil {
 		sr.logger.Warn("error closing client connection", "err", cerr, "client", client.id)
 	}
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -332,6 +332,23 @@ func (sr *Exec) Detach(id *api.ID) error {
 	// TODO: remove from stdin fan-in if you have that side too
 	// sr.ptyPipes.multiInR.Remove(ioClient.pipeInR)
 
+	// 2a) Release the per-client pipe fds. cleanupClient (which would
+	// otherwise close them) never fires on Detach because CopierManager
+	// only invokes finish() on ctx.Done(), so without this every
+	// Attach/Detach cycle leaks both pipe ends in the runner process.
+	// Closing pipeOutR also unblocks the writer copier's Read so its
+	// goroutine exits instead of parking until terminal shutdown.
+	if ioClient.pipeOutW != nil {
+		if perr := ioClient.pipeOutW.Close(); perr != nil {
+			sr.logger.Debug("error closing client pipeOutW", "err", perr, "client", ioClient.id)
+		}
+	}
+	if ioClient.pipeOutR != nil {
+		if perr := ioClient.pipeOutR.Close(); perr != nil {
+			sr.logger.Debug("error closing client pipeOutR", "err", perr, "client", ioClient.id)
+		}
+	}
+
 	// 3) Drop from the registry so it’s no longer discoverable
 	sr.removeClient(ioClient)
 
--- a/pkg/terminal/server/server_attach_pipefdleak_linux_test.go
+++ b/pkg/terminal/server/server_attach_pipefdleak_linux_test.go
@@ -1,0 +1,121 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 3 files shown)


eminwux:
**PR #216 Review — `fix(terminalrunner): close client pipe fds on cleanup and detach`**

Tight bug fix for #215. Close ordering is right in both sites (Remove from `multiOutW` → close `pipeOutW` so buffered bytes drain → close `pipeOutR` so the writer copier's `Read` unblocks with EOF), and the new test counts pipe-typed fds independently of the SCM_RIGHTS regression so the two detectors stay decoupled. The duplicated close block across `cleanupClient` and `Detach` is the right call here — `Detach`'s async-conn-close-with-grace behavior is load-bearing and can't be folded into `cleanupClient`.

A couple of small non-blocking notes:

- **Polish** `internal/terminal/terminalrunner/connections.go:45-54` and `internal/terminal/terminalrunner/lifecycle.go:341-350`: a tiny `(*Exec).closeClientPipes(*ioClient)` helper would keep the two sites in lockstep if a future change touches one. Below the rule-of-three threshold so purely a suggestion.
- **Latent log noise** `internal/terminal/terminalrunner/connections.go:56`: on terminal shutdown `CopierManager`'s `ctx.Done` callback still fires `cleanupClient` for already-`Detach`'d clients, so the new pipe closes now generate Debug log per detached client on shutdown, on top of the pre-existing `conn.Close` Warn. Nilling `client.pipeOut{R,W}`/`conn` after the close in `Detach` would suppress both. Pre-existing-ish — not introduced here, just slightly louder.

LGTM.

PR #217, opened 2026-05-17
fix(terminalrunner): bound interactive attacher writer to unblock fan-out

### LINKED ISSUE #214: Interactive Attach path lacks SetWriteDeadline — single paused attacher head-of-line blocks the PTY fanout
## Context

`internal/terminal/terminalrunner/connections.go:50` (`handleClient`) drives the interactive Attach output path with `dualcopier.RunCopier(client.pipeOutR, client.conn, …)`. `dualcopier.runReadWriter` (`internal/dualcopier/copier.go:108`) does a plain `w.Write(toWrite[total:])` in a loop with **no `SetWriteDeadline` on `client.conn`** and no bounded ring between the PTY reader and the conn.

The subscriber path took the opposite approach (`internal/terminal/terminalrunner/subscriber.go:121-156`): `subscriberWriter.Run` wraps every `conn.Write` in `SetWriteDeadline(subscriberWriteTimeout)` and absorbs PTY output into a bounded ring so a slow consumer either gets the `[sbsh: subscriber lagged]` sentinel and is detached, or is bounded by the per-write deadline.

This was explicitly flagged as the "Adjacent finding (not strictly this bug, worth same PR)" in issue #212. It was deferred from PR #213 because changing `dualcopier.RunCopier` semantics would also affect its second caller (`internal/client/clientrunner/io.go`), and the safer scoped fix (per-write deadline on `client.conn` in `handleClient` only) deserves its own design pass.

After the SCM_RIGHTS fix in PR #213 the cycle-leak deadlock from issue #212 no longer reproduces, but the underlying single-paused-attacher hazard remains: any live attacher whose receiver is genuinely paused (suspended SSH session, blocked terminal emulator, paused tmux pane) will still head-of-line block the fanout `multiOutW` once the socketpair send buffer fills. PTY output stops reaching all other attachers, and the capture sink is gated behind the same multiwriter.

## Proposal

Bring the interactive Attach path up to feature parity with the subscriber path:

- Option A (preferred): replace the direct `dualcopier.RunCopier(client.pipeOutR, client.conn, …)` writer goroutine with a `subscriberWriter`-style wrapper around `client.conn` (bounded ring + per-write `SetWriteDeadline` + lagged-detach). This keeps `dualcopier` semantics unchanged for `clientrunner` and isolates the bounded-output policy to the server side of the interactive attach path.
- Option B: keep `dualcopier`, but add a per-write deadline on the output direction only when the writer is a `*net.UnixConn`. Smaller diff, but couples the deadline policy to `dualcopier` itself.

Either way, the lagged-detach behavior must funnel through `cleanupClient` so `metadata.Status.Attachers` and the multiwriter membership stay consistent.

## Acceptance criteria

- [ ] A single attacher whose receiver is paused (test fixture: stop reading from the client-side conn) does not stall PTY output for other attachers or the capture file beyond a bounded wait (`subscriberWriteTimeout` analogue).
- [ ] The paused attacher is detached after the bounded wait — `metadata.Status.Attachers` no longer lists it, `multiOutW` no longer fans out to its writer, and `cleanupClient` ran exactly once.
- [ ] No semantics change for `internal/client/clientrunner/io.go`'s use of `dualcopier.RunCopier` (verified by existing `internal/client/clientrunner` tests still passing).
- [ ] Regression test along the lines of `pkg/terminal/server/server_attach_fdleak_linux_test.go` from PR #213, but for the paused-attacher signal: assert other attachers keep receiving output and the paused one is reaped.

## Out of scope

- Subscriber path (already has the right shape).
- `dualcopier` refactor for non-output deadlines.
- Reworking the multiwriter API beyond what's needed to swap one writer implementation.

## References

- Issue #212 "Adjacent finding" section.
- PR #213 (the SCM_RIGHTS fd-leak fix).
- `internal/terminal/terminalrunner/subscriber.go:121-156` for the reference shape.
- `internal/terminal/terminalrunner/connections.go:50-99` for the call site to modify.
- `internal/dualcopier/copier.go:48-131` for the unchanged dualcopier semantics.

eminwux:
## Summary

- The interactive Attach path (`internal/terminal/terminalrunner/connections.go`) used to write PTY output to `client.conn` via `dualcopier.RunCopier` with no `SetWriteDeadline` and no bounded ring between the PTY reader and the conn. A single attacher whose receiver was paused (suspended SSH session, blocked emulator, paused tmux pane) would fill the socketpair send buffer and head-of-line block the shared fan-out — every other attacher and the capture sink stopped receiving PTY output until the paused peer drained or the runner died. Issue #214 flagged this as the adjacent finding deferred from PR #213's SCM_RIGHTS fix.
- Per the issue's preferred Option A, the writer-side `dualcopier.RunCopier` goroutine is replaced with a `subscriberWriter`-style wrapper around `client.conn`: bounded in-memory ring (1 MiB default), per-write `SetWriteDeadline` (5 s default), and lagged-detach funnelled through `cleanupClient`. `dualcopier` semantics are unchanged — the reader-side dualcopier and the `internal/client/clientrunner/io.go` consumer keep their existing behavior, so this PR isolates the bounded-output policy to the server side of the interactive attach path.
- The per-attach `os.Pipe` pair allocated by `setupPipes` is no longer needed and is removed; `cleanupClient`, `Detach`, and `ioClient` are simplified accordingly. `cleanupClient` now also calls `removeClient` before `updateTerminalAttachers` so the detached attacher actually drops out of `metadata.Status.Attachers` in a single metadata write (the previous order produced a transient stale snapshot that was only fixed by a follow-up call in the old `handleClient`).
- Reader-side errors (abrupt client disconnect) now feed into the same `detachOnce` funnel via the dualcopier's `errFunc`, so the drain goroutine is woken even when no further PTY bytes arrive — closing a pre-existing latent gap in the cleanup path.
- `pkg/terminal/server/server_attach_pipefdleak_linux_test.go`'s docstring is updated: the test still asserts the same invariant (no per-attach pipe-fd leak) but the underlying architecture changed — the runner allocates no per-attach pipes at all now.

## Test plan

- [x] `make sbsh-sb` (canonical build) + `file ./sbsh` returns ELF executable.
- [x] `go build ./...`, `go vet ./...`.
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full unit test suite passes, including:
  - `internal/client/clientrunner/...` (the other `dualcopier.RunCopier` consumer; semantics unchanged).
  - `internal/terminal/terminalrunner/...` (existing subscriber and runner tests).
  - `pkg/terminal/server/...` (existing fdleak and pipefdleak regression tests still pass + new regression below).
- [x] New regression: `pkg/terminal/server/server_attach_pausedattacher_linux_test.go` exercises the paused-attacher signal end to end. It attaches a paused client (never reads) and a drainer (reads continuously), pumps ~2 MiB through the PTY, and asserts that (a) the drainer keeps receiving past 1.5 MiB, and (b) the paused attacher is reaped from `metadata.Status.Attachers` within a bounded wait while the drainer remains listed.
- [x] `go test -tags=integration ./cmd/sb/get/...` (CI integration step).
- [x] `cd docs/examples/library-consumer && go build ./... && go vet ./...` (CI library-consumer step).
- [x] `pre-commit run` on the staged files — clean (no `golangci-lint` issues).

Closes #214

### DIFF
--- a/internal/terminal/terminalrunner/connections.go
+++ b/internal/terminal/terminalrunner/connections.go
@@ -18,57 +18,74 @@ package terminalrunner
 
 import (
 	"net"
+	"sync"
 
 	"github.com/eminwux/sbsh/internal/dualcopier"
 	"github.com/eminwux/sbsh/pkg/api"
 )
 
+// cleanupClient is the converging cleanup path for an interactive attach.
+// It is invoked exactly once per client (guarded by detachOnce at the call
+// site): on graceful Detach, on reader-side error from the dualcopier,
+// and on bounded-writer lagged-detach. It is responsible for removing the
+// attacher from the PTY fan-out, signalling the drain goroutine to exit
+// (the drain owns the conn close on its defer), and updating
+// metadata.Status.Attachers.
 func (sr *Exec) cleanupClient(client *ioClient) {
 	sr.logger.Info("client connection handler exiting", "client", client.id)
 
-	// Drop this client's pipe from the fan-out *before* closing the conn so
-	// no further PTY output targets a dying writer. Without this, an abrupt
-	// disconnect (socket EOF, attacher kill, network drop) leaves the dead
-	// pipeOutW in multiOutW; a later broken pipe used to cascade into a
-	// fatal terminal event and reap the child process.
+	// Drop this client's writer from the fan-out *before* closing it so no
+	// further PTY output targets a dying ring.
 	sr.ptyPipesMu.RLock()
 	multiOutW := sr.ptyPipes.multiOutW
 	sr.ptyPipesMu.RUnlock()
-	if multiOutW != nil && client.pipeOutW != nil {
-		multiOutW.Remove(client.pipeOutW)
+	if multiOutW != nil && client.outWriter != nil {
+		multiOutW.Remove(client.outWriter)
 	}
 
-	// Release the per-client pipe fds; otherwise each Attach/Detach cycle
-	// leaks both ends in the runner process. Close pipeOutW first so any
-	// bytes still buffered in the pipe can drain to the writer copier,
-	// then pipeOutR so its Read unblocks with EOF and the goroutine exits.
-	if client.pipeOutW != nil {
-		if perr := client.pipeOutW.Close(); perr != nil {
-			sr.logger.Debug("error closing client pipeOutW", "err", perr, "client", client.id)
-		}
-	}
-	if client.pipeOutR != nil {
-		if perr := client.pipeOutR.Close(); perr != nil {
-			sr.logger.Debug("error closing client pipeOutR", "err", perr, "client", client.id)
+	// Closing outWriter signals its drain goroutine to flush any pending
+	// bytes and exit; the drain's deferred conn.Close releases the
+	// socketpair fd. We do not close conn directly here so the drain
+	// remains the sole owner of the lifecycle — double-close on a
+	// *net.UnixConn just logs and is harmless, but a single owner keeps
+	// the path easy to reason about.
+	if client.outWriter != nil {
+		_ = client.outWriter.Close()
+	} else if client.conn != nil {
+		// Early-failure fallback: handleClient never wired the writer,
+		// so there is no drain goroutine to take the conn with it.
+		if cerr := client.conn.Close(); cerr != nil {
+			sr.logger.Warn("error closing client connection", "err", cerr, "client", client.id)
 		}
 	}
 
-	if cerr := client.conn.Close(); cerr != nil {
-		sr.logger.Warn("error closing client connection", "err", cerr, "client", client.id)
-	}
+	// removeClient must happen before updateTerminalAttachers so the
+	// metadata snapshot reflects the just-detached client's absence.
+	// Inverting this order leaves the attacher visible until the next
+	// metadata write.
+	sr.removeClient(client)
 	if err := sr.updateTerminalAttachers(); err != nil {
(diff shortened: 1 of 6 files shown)


eminwux:
**PR #217 Review — `fix(terminalrunner): bound interactive attacher writer to unblock fan-out`**

Correct fix for #214. Swapping the writer-side `dualcopier.RunCopier` for a `subscriberWriter` wrap around `client.conn` is the right shape: it reuses the same bounded-ring + per-write-deadline + lagged-detach primitive the Subscribe path already follows (`internal/terminal/terminalrunner/subscribe.go:104-117`), so this PR extends an existing mechanism rather than introducing a parallel one. `detachOnce` funnelling reader-error / ctx-cancel / drain-lagged into a single `cleanupClient` is well-reasoned, the cleanup-ordering rationale (multiOutW.Remove → outWriter.Close → drain owns conn.Close; removeClient before updateTerminalAttachers) is documented inline, and the new reader-side `detach()` closes the latent gap the old code had when the conn died mid-stream. Test exercises the exact signal: ~2 MiB pumped, drainer asserted past 1.5 MiB, paused attacher reaped within bounded wait while drainer stays present. Commit is verified.

One small item:

- **Stale comment** `internal/terminal/terminalrunner/terminal.go:217`: `// ATTACHED: stream to client (pipeOutW) AND log file` still names `pipeOutW`, which this PR removes — the fan-out target is now a per-attacher `subscriberWriter`. Non-blocking; just a comment refresh.

LGTM.

PR #218, opened 2026-05-17
test(e2e): use t.TempDir for run-path scratch

### LINKED ISSUE #207: test(e2e): runs leave untracked e2e/tmp/ directory in the source tree
## Context

A dev session noticed that running the Go test suite (`go test ./...`) leaves an untracked `e2e/tmp/` directory in the working tree after every run. dev/CLAUDE.md requires a clean tree at commit time, so every session has to `rm -rf e2e/tmp/` before staging — friction that's mechanical but recurring.

## Repro

From a clean working tree on `main`:

```bash
go test ./...
git status
```

Observed:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	e2e/tmp/
```

Contents look like per-run scratch (random IDs as subdirectory names):

```
$ ls e2e/tmp/
0f6127ae  234b27fc  2b0c03e1  3534cb5e  ...
```

## Root cause

`e2e/e2e_test.go:232-237` defines `getRandomRunPath` as `path.Join("tmp", rndDir)`, and `mkdirRunPath` (`e2e_test.go:220-230`) resolves it under `os.Getwd()` and `MkdirAll`s into the source tree:

```go
func getRandomRunPath(t *testing.T) string {
    t.Helper()
    rndDir := naming.RandomID()
    fullDir := path.Join("tmp", rndDir)
    return fullDir
}

func mkdirRunPath(t *testing.T, fullDir string) {
    ...
    fullDir = filepath.Join(cwd, fullDir)
    if err = os.MkdirAll(fullDir, 0o755); err != nil { ... }
}
```

Nothing in `e2e_test.go` registers a `t.Cleanup` to remove `fullDir` (or its `tmp/` parent), and `.gitignore` does not list `e2e/tmp/`, so the artifacts surface as untracked on every dev session.

## Suggested fix

Pick one:

1. Move the e2e scratch into `t.TempDir()` so Go's test runtime cleans up — preferred, since the random-ID directory naming is then redundant and the source tree stays pristine even on `go test -run` failures. Touches `getRandomRunPath` / `mkdirRunPath` and any callers.
2. Add a `t.Cleanup(func() { _ = os.RemoveAll(fullDir) })` inside `mkdirRunPath` so each test cleans its own subdir on success or failure.
3. Add `e2e/tmp/` to `.gitignore` — lowest-effort, but leaves stale scratch accreting on disk session-to-session.

Option 1 also has the benefit that parallel `go test` invocations from different terminals won't collide on `e2e/tmp/`.

## Out of scope

- Refactoring the e2e harness beyond the scratch-directory location.
- Reviewing other tests for similar non-`TempDir` artifacts (separate sweep if anyone wants it).

eminwux:
## Summary
- E2E tests previously wrote scratch dirs into `e2e/tmp/<rnd>` under the source tree (via `path.Join("tmp", rndDir)` resolved against `os.Getwd()` and `os.MkdirAll`), with no cleanup registered. Every `go test ./...` run left untracked `e2e/tmp/` paths, forcing manual `rm -rf` before any commit.
- Collapsed the `getRandomRunPath` + `mkdirRunPath` pair into a single `newRunPath(t)` that returns `t.TempDir()`. The Go test runtime creates the directory and registers a `t.Cleanup` to remove it on completion, so the source tree stays pristine even when tests fail or are interrupted by `go test -run`.
- All eight call sites in `e2e/e2e_sb_test.go`, `e2e/e2e_sb_write_read_test.go`, and `e2e/e2e_sbsh_test.go` updated; `path` and `internal/naming` imports dropped from `e2e_test.go` (no other consumers in that file).
- Chose option 1 from the issue's "Suggested fix" list (preferred there); options 2 and 3 are intentionally not pursued — `t.TempDir` is strictly stronger than the manual `t.Cleanup` variant (handles `t.Run` subtests and panics for free) and avoids the disk-accretion downside of a `.gitignore` entry.

## Test plan
- [x] `make sbsh-sb` — canonical build produces ELF (`file ./sbsh` = `ELF 64-bit LSB executable`).
- [x] `go vet ./...` — clean.
- [x] `go test ./...` — all packages pass, including `github.com/eminwux/sbsh/e2e` (2.713s).
- [x] `git status` after the full test run shows only the intended source edits — **no `e2e/tmp/` untracked path** (the bug repro from the issue body no longer reproduces).
- [x] `pre-commit run --files <changed>` — all hooks pass (`go fmt`, `golangci-lint`, `go-mod-tidy`, `addlicense`).

Closes #207

### DIFF
--- a/e2e/e2e_sb_test.go
+++ b/e2e/e2e_sb_test.go
@@ -42,8 +42,7 @@ func TestSb_NoTerminals(t *testing.T) {
 		{"g", "t"},
 	}
 	for _, args := range tests {
-		runPath := getRandomRunPath(t)
-		mkdirRunPath(t, runPath)
+		runPath := newRunPath(t)
 		runPathEnv := buildSbRunPathEnv(t, runPath)
 		out = runReturningBinary(t, []string{runPathEnv}, sb, args...)
 		expected = discovery.NoTerminalsString
@@ -63,8 +62,7 @@ func TestSb_NoClients(t *testing.T) {
 		{"get", "clients"},
 	}
 	for _, args := range tests {
-		runPath := getRandomRunPath(t)
-		mkdirRunPath(t, runPath)
+		runPath := newRunPath(t)
 		runPathEnv := buildSbRunPathEnv(t, runPath)
 		out = runReturningBinary(t, []string{runPathEnv}, sb, args...)
 		expected = discovery.NoClientsString
--- a/e2e/e2e_sb_write_read_test.go
+++ b/e2e/e2e_sb_write_read_test.go
@@ -32,8 +32,7 @@ import (
 // command's output appears in both the capture file (`sb read`) and the
 // live stream (`sb read -f`).
 func TestSb_WriteRead_Roundtrip(t *testing.T) {
-	runPath := getRandomRunPath(t)
-	mkdirRunPath(t, runPath)
+	runPath := newRunPath(t)
 
 	termName := "e2e-writeread"
 	runPathEnv := buildSbRunPathEnv(t, runPath)
@@ -113,8 +112,7 @@ func TestSb_WriteRead_Roundtrip(t *testing.T) {
 // sending bogus bytes.
 func TestSb_Write_CaretParsingErrors(t *testing.T) {
 	t.Parallel()
-	runPath := getRandomRunPath(t)
-	mkdirRunPath(t, runPath)
+	runPath := newRunPath(t)
 	runPathEnv := buildSbRunPathEnv(t, runPath)
 
 	// No terminal needs to be running — parsing happens before we dial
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #218 Review — `test(e2e): use t.TempDir for run-path scratch`**

Clean, minimal fix for #207. Collapses the `getRandomRunPath` + `mkdirRunPath` pair into `newRunPath(t)` returning `t.TempDir()`, which delegates cleanup to the test runtime and works correctly even when tests panic or are interrupted. All 7 call sites converted; `path` and `internal/naming` imports correctly dropped from `e2e_test.go` (no remaining references). Subprocess receives an absolute `SB_RUN_PATH` instead of a cwd-relative one — functionally equivalent (the prior `mkdirRunPath` mkdir'd under cwd and the subprocess inherited cwd, so both paths resolved to the same dir) and strictly cleaner. No correctness, leak, dead-code, or convention concerns; commits verified; bug repro from the issue confirmed no longer reproduces.

LGTM.

Issue #226, opened 2026-05-18
bug(terminalrunner): Subscribe/Close race orphans subscriber drain goroutine and fd


eminwux:
## Context

Surfaced by audit #219.

`internal/terminal/terminalrunner/lifecycle.go:243` calls \`closeAllSubscribers()\`, which snapshots \`sr.subscribers\` under \`subsMu\`, drops the lock, then iterates calling \`s.Close()\` on each. After that, \`Close()\` cancels the ctx and closes the PTY.

`internal/terminal/terminalrunner/subscribe.go:65-127` \`Subscribe\` checks \`multiOutW != nil\` (which Close never nils), then \`addSubscriber\` → \`multiOutW.Add\` → \`go sub.Run()\`.

Race window: after \`closeAllSubscribers\` releases \`subsMu\` but before the runner exits, a concurrent \`Subscribe\` RPC (a \`srv.ServeCodec\` goroutine still draining a pre-accepted control connection) lands. The new subscriber registers itself in \`sr.subscribers\` and starts a drain goroutine that parks on \`s.cond.Wait()\` — \`terminalManagerReader\` is already gone, so no \`Write\` ever wakes it, no \`Close()\` from \`closeAllSubscribers\` (already executed) ever fires.

The orphaned subscriber holds:
- A goroutine parked on \`cond.Wait\` indefinitely
- The socketpair server-side fd (held via \`s.conn\`)
- A reference in \`multiOutW.writers\` and in \`sr.subscribers\` (until those collections are GC'd)

Under repeated runner restart in tests this accumulates.

## Proposal

Same family as the \`Attach\` race tracked in the panic issue:

1. Cheapest: have \`Subscribe\` check ctx.Done() (or a shutdown flag) before \`addSubscriber\`, and again after \`go sub.Run()\` — if shutdown was observed at any point, immediately call \`detach()\` so the drain wakes and the conn closes.
2. Architectural: add the \`shuttingDown\` atomic flag (same as the panic issue) and short-circuit Subscribe with a typed error post-Close.

(2) collapses both races into one fix; (1) is a localized patch.

## Acceptance criteria

- [ ] Concurrent \`Close()\` + \`Subscribe()\` does not leave a subscriber whose drain goroutine is unreachable
- [ ] Regression test: with race detector, run N concurrent Subscribe calls racing one Close, then assert all drain goroutines have exited (e.g., via a wait-group hook or by polling \`runtime.NumGoroutine\` to settle)
[...]

PR #233, opened 2026-05-18
fix(terminalrunner): clear clients map in Close to prevent nil-map panic

### LINKED ISSUE #225: bug(terminalrunner): Close() panics on concurrent Attach via nil-map write
## Context

Surfaced by audit #219.

`internal/terminal/terminalrunner/lifecycle.go:232-240`:

\`\`\`go
sr.clientsMu.Lock()
for _, c := range sr.clients {
    if err := c.conn.Close(); err != nil {
        sr.logger.Warn(\"could not close client connection\", ...)
    }
}

sr.clients = nil
sr.clientsMu.Unlock()
\`\`\`

`sockets.go:CreateNewClient` → `connections.go:addClient` does:

\`\`\`go
func (sr *Exec) addClient(c *ioClient) {
    sr.clientsMu.Lock()
    sr.clients[*c.id] = c   // <-- write to nil map → PANIC
    sr.clientsMu.Unlock()
}
\`\`\`

Race window: \`Close()\` releases \`clientsMu\` after setting \`sr.clients = nil\`. A concurrent RPC \`Attach\` (any \`srv.ServeCodec\` goroutine in-flight, since per-conn handlers outlive the accept loop's ctx-cancel exit) acquires the mutex after Close and writes to the nil map, panicking the runner process.

This is the most severe finding from the audit sweep. The other registries (\`subscribers\`) are not nil'd in Close, so they only race on iteration, not on write.

## Proposal

Pick one of:

1. Drop the \`sr.clients = nil\` assignment — emptying the loop above is already sufficient to release client refs; the map can persist until GC.
2. Replace with \`clear(sr.clients)\` (Go 1.21+) so the map stays addressable.
3. Add a \`shuttingDown\` atomic flag (matches the pattern from PR #185 for \`Controller.Close\`) and have \`CreateNewClient\` / \`Subscribe\` short-circuit to an error once it's set, so post-Close attach attempts get a clean rejection instead of a races-or-panics path.

(3) is the architecturally cleanest — it also closes the parallel \`Subscribe\`/\`Close\` race tracked separately — but (1) or (2) is enough to remove the panic. The choice depends on whether post-Close \`Attach\` should fail loudly (return error) or quietly (succeed-but-fail-on-first-IO).

## Acceptance criteria

- [ ] \`Exec.Close()\` no longer leaves \`sr.clients\` in a state where a subsequent \`addClient\` panics
- [ ] Regression test: concurrent \`Close()\` + \`CreateNewClient()\` from multiple goroutines completes without panic across N>=100 iterations (race detector enabled)
- [ ] If option (3) is chosen, \`Subscribe\` and \`CreateNewClient\` both check the shutdown flag and return a typed error post-Close

## Out of scope

- The parallel \`Subscribe\`/\`Close\` orphan-subscriber leak — tracked separately.
- The downstream behavior question of what an RPC client should see when it calls Attach against a closing runner — covered by acceptance criteria above but the exact error surface is a design call.

## References

- Audit issue: #219
- Sibling pattern: PR #185 (\`Controller.Close\` atomic shutdown gate via CAS)

eminwux:
## Summary

- Fixes #225 — the priority-A panic where `Close()` set `sr.clients = nil` and a concurrent `addClient` from a per-conn RPC handler still in flight (handlers outlive the accept loop's ctx-cancel exit) wrote to a nil map and crashed the runner process.
- Replaces `sr.clients = nil` with `clear(sr.clients)` (Option 2 from the issue body) — the lowest-blast-radius fix among the three proposals. The map stays addressable, so a late `addClient` is a benign stale-entry write instead of a panic. Per CLAUDE.md "Decide-or-stall": preferred the minimal one-line change over the broader atomic-shutdown-flag refactor (Option 3); the architectural shutdown-gate work that would also close the sibling `Subscribe`/`Close` race (#226) can land separately without rebasing on this.
- Option 3's `Subscribe`/`CreateNewClient` shutdown-flag short-circuit is not implemented here — that's the post-Close error-surface design call the issue explicitly defers; AC bullet 3 is conditional on choosing Option 3, so it does not apply.
- Adds two regression tests in `lifecycle_close_clients_race_test.go`:
  - `TestExecClose_PostCloseAddClientDoesNotPanic` is the deterministic one — verified to fail with the exact `assignment to entry in nil map` panic when the fix is reverted.
  - `TestExecClose_NoNilMapOnConcurrentAddClient` races 200 concurrent `addClient` calls against `Close()` under `-race`; double-duty as a lock-discipline check.

## Test plan

- [x] `make sbsh-sb` — produces ELF binary
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full CI test target (cmd/sb, cmd/sbsh, internal/terminal, internal/client, integration, e2e)
- [x] `go test -race -count=3 ./internal/terminal/terminalrunner/`
- [x] Verified the deterministic test fails without the fix and passes with it
- [x] pre-commit hooks pass (golangci-lint, go fmt, addlicense, etc.)

Closes #225

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -229,14 +229,19 @@ func (sr *Exec) Close(reason error) error {
 	}
 
 	// close clients
+	//
+	// clear() instead of `sr.clients = nil`: a concurrent addClient call
+	// can race past Close's unlock (per-conn RPC handlers outlive the
+	// accept loop's ctx-cancel exit), and writing to a nil map panics.
+	// Emptying the map keeps it addressable so the late write is a benign
+	// stale-entry no-op shape. See #225.
 	sr.clientsMu.Lock()
 	for _, c := range sr.clients {
 		if err := c.conn.Close(); err != nil {
 			sr.logger.Warn("could not close client connection", "id", sr.id, "client", c.id, "err", err)
 		}
 	}
-
-	sr.clients = nil
+	clear(sr.clients)
 	sr.clientsMu.Unlock()
 
 	// close all output subscribers
--- a/internal/terminal/terminalrunner/lifecycle_close_clients_race_test.go
+++ b/internal/terminal/terminalrunner/lifecycle_close_clients_race_test.go
@@ -1,0 +1,143 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package terminalrunner
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+	"net"
+	"runtime/debug"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/eminwux/sbsh/pkg/api"
+)
+
+// newDummyClient mints an ioClient with a real net.Conn so Close's
+// per-client conn.Close call has something to close. net.Pipe returns two
+// connected in-memory conns; we keep just one — the other is GC'd, which
+// is fine for a teardown-only test.
+func newDummyClient(idx int) *ioClient {
+	id := api.ID(fmt.Sprintf("client-%d", idx))
+	c, _ := net.Pipe()
+	return &ioClient{id: &id, conn: c}
+}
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #233 Review — `fix(terminalrunner): clear clients map in Close to prevent nil-map panic`**

Mechanically minimal fix — `clear(sr.clients)` keeps the map addressable so a late `addClient` (per-conn RPC handler outliving the accept loop's ctx-cancel) writes a benign stale entry instead of panicking. Lock discipline on `clientsMu` is consistent across `addClient`/`removeClient`/`getClient`/`Close`/`sockets.go`'s broadcast. The decision to land Option 2 here and defer the shutdown-gate work (Option 3 / #226) is documented inline at `lifecycle.go:230-235` and in the PR body, which is the right call to keep blast radius small.

Regression coverage is solid: `TestExecClose_PostCloseAddClientDoesNotPanic` is the always-fires deterministic proof, and `TestExecClose_NoNilMapOnConcurrentAddClient` races 200 goroutines under `-race` for lock-discipline double-duty. `newCloseRaceExec` mirrors `newShutdownTestExec`'s shape and the struct fields line up with the real `Exec` definition. License header on the new test file is correct.

No blocking items. No sensitive-data concerns. No dead symbols.

LGTM.

PR #234, opened 2026-05-18
fix(terminalrunner): propagate real shell exit code/signal in non-init watchChildExit

### LINKED ISSUE #227: bug(terminalrunner): non-init-mode watchChildExit discards real shell exit code
## Context

Surfaced by audit #219.

`internal/terminal/terminalrunner/lifecycle.go:120-154` \`watchChildExit\`:

\`\`\`go
if sr.initMode && sr.reaper != nil {
    code, ok := <-sr.reaper.TrackedExitCh()
    if ok {
        exitErr = fmt.Errorf(\"shell process exited: code=%d\", code)
    } else {
        exitErr = errors.New(\"shell process exited\")
    }
    ...
} else {
    _ = sr.cmd.Wait()
    exitErr = errors.New(\"the shell process has exited\")
}
\`\`\`

The init-mode branch captures the actual exit code (\`code=%d\`). The non-init-mode branch discards \`cmd.Wait()\`'s return — which carries the \`*exec.ExitError\` with the real exit status and signal info — and substitutes a fixed-string error.

This loses signal: \`waitOnTerminal\` then sends this opaque string on \`closeReqCh\`, which becomes the \`EvCmdExited.Err\` event, which the controller logs and surfaces to callers. Outside container init mode (the default!), users cannot distinguish \"shell exited 0\" from \"shell killed by SIGSEGV\".

## Proposal

\`\`\`go
} else {
    err := sr.cmd.Wait()
    if err != nil {
        exitErr = fmt.Errorf(\"shell process exited: %w\", err)
    } else {
        exitErr = errors.New(\"shell process exited (code 0)\")
    }
}
\`\`\`

If callers downstream switch on the literal \"the shell process has exited\" string, they need updating — that's a separate concern but worth grepping before flipping.

## Acceptance criteria

- [ ] Non-init-mode \`watchChildExit\` propagates \`cmd.Wait()\` error (\`*exec.ExitError\` exposes code + signal)
- [ ] Test: spawn a child that exits with code 42, assert the \`EvCmdExited\` event's error carries 42
- [ ] Test: spawn a child killed by SIGSEGV, assert the event carries signal info
- [ ] Grep ensures no caller switches on the old fixed string

## Out of scope

- The init-mode branch already captures the code via reaper.TrackedExitCh; not touched here.

## References

- Audit issue: #219

eminwux:
## Summary
- The non-init-mode branch of `watchChildExit` discarded `cmd.Wait()`'s `*exec.ExitError`, substituting a fixed string. Outside container init mode (the default), `EvCmdExited.Err` could not distinguish a clean shell exit from `exit 42` or a SIGSEGV death.
- Wrap `cmd.Wait()`'s error verbatim so consumers can `errors.As` it to `*exec.ExitError` and read both `ExitCode()` and the `syscall.WaitStatus` signal info.
- Clean exits (`cmd.Wait() == nil`) now report `"shell process exited (code 0)"` so the close-path event is still non-nil for log consumers.

## Test plan
- [x] `make test` (full repo target — covers `./cmd/sb...`, `./cmd/sbsh...`, `./internal/terminal...`, `./internal/client...`, integration, e2e)
- [x] `make sbsh-sb` produces an ELF executable (`file ./sbsh`)
- [x] `go test -race ./internal/terminal/terminalrunner/...` — three new regression tests pass:
  - `TestWatchChildExit_NonInitMode_PropagatesExitCode` — child exits 42, asserts wrapped `ExitError.ExitCode() == 42`
  - `TestWatchChildExit_NonInitMode_PropagatesSignal` — child killed by SIGSEGV, asserts `WaitStatus.Signaled() && WaitStatus.Signal() == SIGSEGV`
  - `TestWatchChildExit_NonInitMode_CleanExit` — exit 0 path returns a non-nil sentinel that is *not* a wrapped `ExitError`
- [x] `pre-commit run --files ...` clean
- [x] Grep for the prior literal `"the shell process has exited"` confirms no callers switch on it (only site was the line being changed)

## Notes
- Init-mode branch (already correct via `reaper.TrackedExitCh`) intentionally untouched per issue AC.

Closes #227

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -139,8 +139,14 @@ func (sr *Exec) watchChildExit() {
 			_ = sr.cmd.Process.Release()
 		}
 	} else {
-		_ = sr.cmd.Wait()
-		exitErr = errors.New("the shell process has exited")
+		// cmd.Wait()'s *exec.ExitError carries the real exit code (and signal
+		// info on signaled deaths) — propagate it so EvCmdExited consumers can
+		// distinguish "shell exited 0" from "shell killed by SIGSEGV".
+		if werr := sr.cmd.Wait(); werr != nil {
+			exitErr = fmt.Errorf("shell process exited: %w", werr)
+		} else {
+			exitErr = errors.New("shell process exited (code 0)")
+		}
 	}
 
 	sr.logger.Info("child process has exited", "parent_pid", os.Getpid(), "child_pid", pid)
--- a/internal/terminal/terminalrunner/lifecycle_watch_child_exit_test.go
+++ b/internal/terminal/terminalrunner/lifecycle_watch_child_exit_test.go
@@ -1,0 +1,159 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package terminalrunner
+
+import (
+	"context"
+	"errors"
+	"io"
+	"log/slog"
+	"os/exec"
+	"sync"
+	"syscall"
+	"testing"
+	"time"
+)
+
+// newWatchChildExitTestExec builds a minimal non-init-mode Exec suitable for
+// exercising watchChildExit's cmd.Wait() branch in isolation — no PTY, no RPC,
+// no metadata. The caller is responsible for cmd.Start before calling
+// watchChildExit; watchChildExit then Waits the child and publishes the
+// real exit error on closeReqCh.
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #234 Review — `fix(terminalrunner): propagate real shell exit code/signal in non-init watchChildExit`**

Clean targeted bugfix. The non-init branch now propagates `cmd.Wait()`'s `*exec.ExitError` verbatim via `%w`, so `errors.As(EvCmdExited.Err, &exitErr)` recovers both `ExitCode()` and the `syscall.WaitStatus` signal info; the clean-exit path keeps a non-nil sentinel so consumers still see a close event. Implementation matches issue #227's proposed patch one-for-one, no scope creep. Three regression tests in `lifecycle_watch_child_exit_test.go` cover exit-42, SIGSEGV, and exit-0, isolating `watchChildExit` via the same minimal-Exec helper pattern as `lifecycle_shutdown_test.go`. Grep across `pr-234` confirms no consumer switches on the old literal `"the shell process has exited"` and the `EvCmdExited.Err` consumers (`pkg/terminal/server/server.go:176`, `internal/client/controller.go:404`, `internal/terminal/controller.go:241`) just log/propagate, so the format change is safe. No sensitive data, no dead symbols, all commits signature-verified.

A non-blocking observation:

- **Init-mode parity (informational)** `internal/terminal/terminalrunner/lifecycle.go:128-141`: the init-mode branch still surfaces exit info as a `code=%d` string (with `128+sig` encoding from the reaper), so `errors.As` only recovers `*exec.ExitError` in non-init mode. PR explicitly defers this per issue AC and the "Out of scope" note, and no current consumer cares — flagging for awareness only, not a follow-up issue.

LGTM.

PR #237, opened 2026-05-18
fix(terminalrunner): close Subscribe/Close race that orphans drain goroutines

### LINKED ISSUE #226: bug(terminalrunner): Subscribe/Close race orphans subscriber drain goroutine and fd

eminwux:
## Summary
- Adds two ctx-check gates inside `Subscribe`: an *early* check (before any registry mutation) that rejects late RPCs cheaply once shutdown is visible, and a *late* check (after `addSubscriber` + `multiOutW.Add` + `go sub.Run()`) that closes the racing window where `closeAllSubscribers` already snapshotted `sr.subscribers` before the subscriber was registered. The late check calls `sub.Close()` so the drain wakes, runs its deferred `onDetach` (`multiOutW.Remove` + `removeSubscriber`), and closes `conn` — both `detach()` and `sub.Close()` are idempotent so co-occurring with `Close`'s own iteration is harmless.
- New `errTerminalClosing` sentinel in `errors.go` lets callers match on the post-Close rejection. Package-private; callers across the RPC boundary see the string.
- Adopts option (1) from the issue (localized two-check patch) rather than option (2) (shuttingDown atomic flag). #225's parallel `Attach` panic was already fixed in #233 via `clear(sr.clients)`, so the architectural shutdown gate no longer needs to subsume both — keeping the change narrow.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make sbsh-sb` produces ELF executables (`file ./sbsh ./sb` confirms `ELF 64-bit LSB executable`)
- [x] `go test -race -count=100 -timeout=300s -run '^TestSubscribe_' ./internal/terminal/terminalrunner/` — 100/100 race-clean
- [x] New `TestSubscribe_NoOrphanDrainOnConcurrentClose` regression test: confirmed it *fails* against the pre-fix code (residual subscribers from orphaned drain goroutines) and passes against this PR
- [x] New `TestSubscribe_PostCloseRejectsCleanly` deterministic regression for the post-Close single-thread path
- [x] `make test` (unit + integration + e2e)
- [x] `make test-race`

Closes #226

### DIFF
--- a/internal/terminal/terminalrunner/errors.go
+++ b/internal/terminal/terminalrunner/errors.go
@@ -23,3 +23,9 @@ var (
 	ErrPipeWrite error = errors.New("could not write pty->pipe")
 	ErrPipeRead  error = errors.New("could not read pipe->pipe")
 )
+
+// errTerminalClosing is returned by Subscribe (and any future post-Close
+// RPC handler) when the runner has observed ctx cancel and refuses to
+// register a new attacher. Package-private — callers across the RPC
+// boundary see the string form.
+var errTerminalClosing = errors.New("terminal closing")
--- a/internal/terminal/terminalrunner/subscribe.go
+++ b/internal/terminal/terminalrunner/subscribe.go
@@ -101,6 +101,16 @@ func (sr *Exec) Subscribe(req *api.SubscribeRequest, response *api.ResponseWithF
 		return fmt.Errorf("Subscribe: terminal not running")
 	}
 
+	// Early shutdown check: avoid registry mutation when Close has already
+	// flipped the context. Cheaper than the late check below and rejects
+	// late RPCs that arrive on a per-conn handler still draining after the
+	// accept loop's ctx-cancel exit. See #226.
+	if sr.ctx.Err() != nil {
+		_ = conn.Close()
+		_ = unix.Close(cliFD)
+		return errTerminalClosing
+	}
+
 	var sub *subscriberWriter
 	var detachOnce sync.Once
 	detach := func() {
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #237 Review — `fix(terminalrunner): close Subscribe/Close race that orphans drain goroutines`**

Correct, focused fix for #226. The early/late ctx-check pair closes the snapshot-vs-addSubscriber window cleanly, idempotency on `detach()` + `sub.Close()` keeps it safe under either ordering with `closeAllSubscribers`, and the race regression test (100/100 race-clean per the test plan) exercises the actual window rather than mocking it. No correctness, secret-leak, or dead-code blockers; all commits are signed; CI is green.

One non-blocking nit:

- **Dead test signal** `internal/terminal/terminalrunner/subscribe_close_race_test.go:115`: `subscribeErrCount` is incremented at line 135 but never read. The comment claims it \"surfaces if every single subscriber was rejected (would indicate the race window never opened),\" but no assertion checks that — silently dead. Either assert (e.g. \`if subscribeErrCount.Load() == concurrentSubscribers { t.Fatal(\"race window never opened\") }\`) or drop the variable and the comment.

LGTM.

Issue #232, opened 2026-05-18
bug(terminal/controller): runLoop checks wrong variable, silently drops Close errors on rpcDoneCh path


eminwux:
## Repro

While auditing terminalrunner shutdown (#219) and writing the misleading-shutdown-error follow-up (#228), I read `internal/terminal/controller.go:218-224`:

\`\`\`go
case err := <-c.rpcDoneCh:
    c.logger.ErrorContext(c.ctx, \"rpc server has failed\", \"err\", err)
    if errC := c.Close(err); err != nil {
        c.logger.ErrorContext(c.ctx, \"error during Close after rpc server failure\", \"err\", errC)
        err = fmt.Errorf(\"%w: %w: %w\", err, errdefs.ErrOnClose, errC)
    }
    return fmt.Errorf(\"%w: %w\", errdefs.ErrRPCServerExited, err)
\`\`\`

The conditional reads `if errC := c.Close(err); err != nil`. The assignment is to `errC` (Close's return) but the gate checks `err` (the value received from `rpcDoneCh`). When `err == nil` (which is the *normal* graceful-shutdown path after #228 lands, or any time \`rpcDoneCh\` delivers a nil), `errC` from Close is never inspected — any Close error is silently dropped.

When `err != nil` *and* Close also failed, the wrapping currently works by accident (errC is the latest assignment, used in the body). But the inverse case — graceful rpcDoneCh + Close failure — loses the Close error entirely with no log line and no contribution to the returned error.

Not reproduced with a test; spotted by code inspection. Reproducible by injecting a Close that returns a non-nil error and triggering the graceful rpcDoneCh path.

## Observed

Source code; not a runtime trace. The shape:

- `c.Close(err)` returns a non-nil `errC` — for example, the future graceful-path nil-on-rpcDoneCh from #228 combined with any Close-side cleanup failure (socket removal warning escalated to error, metadata write rejected, etc.).
- The condition `err != nil` evaluates to false because `err == nil`.
- The if-body is skipped — `errC` is shadowed and lost.
- The function returns `fmt.Errorf(\"%w: %w\", errdefs.ErrRPCServerExited, err)` with `err == nil`, which renders as `\"ErrRPCServerExited: %!w(<nil>)\"` (the latter pathology also tracked in #162, but that's a different code site).

## Root cause

Typo in the if-condition — the variable assigned to should be the one being checked. The pattern that works:

\`\`\`go
if errC := c.Close(err); errC != nil {
    ...
}
[...]

Issue #219, opened 2026-05-18
audit(terminalrunner): lifecycle, cleanup, and fd-ownership sweep


eminwux:
## Context
The last three merged fixes in `internal/terminal/terminalrunner/` all
addressed lifecycle or fd-ownership bugs:
- #217 — bound interactive attacher writer to unblock fan-out
- #216 — close client pipe fds on cleanup and detach
- #185 — atomic shutdown gate in `Controller.Close` prevents deadlock
- #184 — prune dead writers instead of killing the PTY

That pattern density says adjacent code in the same package is likely to
have siblings of the same bug classes. Sweep with a bug-hunting lens now,
rather than waiting for the next regression.

## Proposal
Read every `.go` file under `internal/terminal/terminalrunner/` with focus on:
- file-descriptor ownership across attach / detach / cleanup paths
- goroutine lifetimes and shutdown ordering
- writer fan-out blocking on slow or dead consumers
- error paths that skip cleanup
- signal handling and reaper interaction

For each defect found, file a new `bug`-labeled issue linking back to this
audit. Land any small obvious fixes (one-line nil check, missing close,
typo'd defer) in a single PR against this issue. Larger fixes belong in
the per-bug follow-ups.

If nothing actionable surfaces, close this issue with an audit-report
comment summarising the files read and the invariants checked.

## Acceptance criteria
- [ ] Every `.go` file under `internal/terminal/terminalrunner/` has been read
- [ ] fd-ownership across attach/detach/cleanup verified end-to-end
- [ ] goroutine lifetimes vs. controller shutdown verified
- [ ] writer fan-out behavior under slow/dead consumers verified
- [ ] Each finding filed as a separate `bug` issue linked here
- [ ] This issue closed with either a fix PR or an audit-report comment

## Notes
Recent neighbors: #217, #216, #185, #184, #195, #167.

PR #239, opened 2026-05-18
fix(terminalrunner): send nil on doneCh for graceful ctx-cancel shutdown

### LINKED ISSUE #228: bug(terminalrunner): StartServer reports misleading 'unknown rpc server error' on graceful ctx-cancel shutdown
## Context

Surfaced by audit #219.

`internal/terminal/terminalrunner/control_server.go:68-79`:

\`\`\`go
conn, err := sr.lnCtrl.Accept()
if err != nil {
    // Normal path: listener closed by ctx cancel
    if errors.Is(err, net.ErrClosed) || sr.ctx.Err() != nil {
        select {
        case doneCh <- errors.New(\"unknown rpc server error\"):
        default:
        }
        close(doneCh)
        return
    }
    ...
}
\`\`\`

The comment on the surrounding block says \"Normal path: listener closed by ctx cancel\" — but the value sent on \`doneCh\` is a non-nil error labelled \`unknown rpc server error\`. Consumers (\`internal/terminal/controller.go:218\` logs it as \`\"rpc server has failed\"\`; \`pkg/terminal/server/server.go:182-186\` returns it as the runLoop cause) treat any non-nil err as abnormal.

Sibling \`internal/client/clientrunner/control_server.go:76-82\` correctly sends \`nil\` on this path. The terminalrunner version diverged.

## Proposal

Send \`nil\` on the graceful path, matching the sibling:

\`\`\`go
if errors.Is(err, net.ErrClosed) || sr.ctx.Err() != nil {
    select {
    case doneCh <- nil:
    default:
    }
    close(doneCh)
    return
}
\`\`\`

Consumer impact — \`internal/terminal/controller.go:218-224\` currently does:

\`\`\`go
case err := <-c.rpcDoneCh:
    c.logger.ErrorContext(c.ctx, \"rpc server has failed\", \"err\", err)
    if errC := c.Close(err); err != nil { ... }   // also: condition checks 'err' not 'errC' — separate bug
    return fmt.Errorf(\"%w: %w\", errdefs.ErrRPCServerExited, err)
\`\`\`

With \`err = nil\`, the wrapper \`fmt.Errorf(\"%w: %w\", errdefs.ErrRPCServerExited, nil)\` produces \`\"ErrRPCServerExited: %!w(<nil>)\"\`. So the consumer needs a small companion fix to handle the nil-doneCh case cleanly (drop the log or distinguish graceful vs abnormal). Bundle or split as the implementer prefers.

Also flagged in passing: the controller's \`if errC := c.Close(err); err != nil\` checks the wrong variable — assignment is to \`errC\` but condition gates on \`err\`. Worth filing separately if a follow-up touches this code.

## Acceptance criteria

- [ ] Graceful ctx-cancel shutdown delivers \`nil\` on \`doneCh\`, matching clientrunner
- [ ] Consumers (\`internal/terminal/controller.go\`, \`pkg/terminal/server/server.go\`) handle the nil-from-graceful case without emitting bogus error logs
- [ ] Regression test: start runner, cancel ctx, assert no spurious \"rpc server has failed\" log line

## Out of scope

- The wrong-variable check in \`internal/terminal/controller.go:220\` — file separately if not bundled.

## References

- Audit issue: #219
- Sibling reference: \`internal/client/clientrunner/control_server.go:76-82\` (correct pattern)

eminwux:
## Summary

- StartServer (`internal/terminal/terminalrunner/control_server.go`) now sends `nil` on `doneCh` when the listener is closed by ctx cancel, matching the clientrunner sibling. Previously it sent `errors.New("unknown rpc server error")`, causing consumers to treat a clean shutdown as an abnormal failure.
- Controller's `rpcDoneCh` consumer (`internal/terminal/controller.go`) splits the graceful (`err == nil`) path: it logs at Info, calls `Close(nil)`, and returns the bare `errdefs.ErrRPCServerExited` sentinel — no more bogus "rpc server has failed" log line, and no more `%!w(<nil>)` rendering from `fmt.Errorf("%w: %w", ErrRPCServerExited, nil)`.
- Server facade's `runLoop` (`pkg/terminal/server/server.go`) mirrors the ctx.Done branch when `rpcDoneCh` delivers `nil`: returns `ctx.Err()` so callers see the cancellation cause rather than a synthetic "server: rpc server exited" string.

## Scope notes

- The wrong-variable check at `internal/terminal/controller.go:220` (`if errC := c.Close(err); err != nil` — assignment is to `errC`, condition gates on `err`) is left as-is per the issue's "Out of scope" note. Tracked separately in #232.
- Out-of-band: this is part of the audit cluster around #219.

## Test plan

- [x] `make sbsh-sb` + `file ./sbsh` (ELF 64-bit LSB executable)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `E2E_BIN_DIR=$(pwd) go test ./e2e`
- [x] `go test -run TestStartServer_GracefulCtxCancelSendsNil -race ./internal/terminal/terminalrunner/`
- [x] `pre-commit run --files ...`

Closes #228

### DIFF
--- a/internal/terminal/controller.go
+++ b/internal/terminal/controller.go
@@ -216,6 +216,14 @@ func (c *Controller) Run(spec *api.TerminalSpec) error {
 			c.handleEvent(ev)
 
 		case err := <-c.rpcDoneCh:
+			if err == nil {
+				c.logger.InfoContext(c.ctx, "rpc server exited gracefully")
+				if errC := c.Close(nil); errC != nil {
+					c.logger.ErrorContext(c.ctx, "error during Close after rpc server graceful exit", "err", errC)
+					return fmt.Errorf("%w: %w", errdefs.ErrRPCServerExited, errC)
+				}
+				return errdefs.ErrRPCServerExited
+			}
 			c.logger.ErrorContext(c.ctx, "rpc server has failed", "err", err)
 			if errC := c.Close(err); err != nil {
 				c.logger.ErrorContext(c.ctx, "error during Close after rpc server failure", "err", errC)
--- a/internal/terminal/terminalrunner/control_server.go
+++ b/internal/terminal/terminalrunner/control_server.go
@@ -71,7 +71,7 @@ func (sr *Exec) StartServer(
 			// Normal path: listener closed by ctx cancel
 			if errors.Is(err, net.ErrClosed) || sr.ctx.Err() != nil {
 				select {
-				case doneCh <- errors.New("unknown rpc server error"):
+				case doneCh <- nil:
 				default:
 				}
 				close(doneCh)
--- a/internal/terminal/terminalrunner/control_server_graceful_test.go
+++ b/internal/terminal/terminalrunner/control_server_graceful_test.go
@@ -1,0 +1,114 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package terminalrunner
+
+import (
+	"bytes"
+	"context"
+	"log/slog"
+	"net"
+	"path/filepath"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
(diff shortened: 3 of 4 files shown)


eminwux:
**PR #239 Review — `fix(terminalrunner): send nil on doneCh for graceful ctx-cancel shutdown`**

Clean, focused fix for #228. `StartServer` now sends `nil` on the graceful ctx-cancel path, matching the clientrunner sibling at `internal/client/clientrunner/control_server.go:78`. The controller and server-facade consumers both split the `err == nil` branch correctly: the controller returns the bare `errdefs.ErrRPCServerExited` sentinel (so `errors.Is` consumers in `cmd/sbsh/sbsh_test.go` and `internal/terminal/controller_test.go:420` keep matching, and the prior `%!w(<nil>)` rendering goes away), and `pkg/terminal/server/server.go:186-189` propagates `ctx.Err()` instead of the synthetic `"server: rpc server exited"` string. New regression test exercises both the nil delivery and the absence of the spurious `"rpc server has failed"` log line. All commits verified, no sensitive-data leaks.

LGTM to proceed — no blocking items.

PR #240, opened 2026-05-18
fix(terminalrunner): close capture file fd in Close to prevent leak

### LINKED ISSUE #229: bug(terminalrunner): Close() leaks capture-file fd (logf is never closed)
## Context

Surfaced by audit #219.

`internal/terminal/terminalrunner/terminal.go:194-218` (\`startPty\`) opens the capture file:

\`\`\`go
logf, errO := os.OpenFile(captureFile, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o600)
if errO != nil {
    return fmt.Errorf(\"open log file: %w\", errO)
}
if errPerm := sr.applyArtifactPerms(\"capture\", captureFile, captureMode, captureGID); errPerm != nil {
    _ = logf.Close()
    return errPerm
}
...
multiOutW := NewDynamicMultiWriter(sr.logger, logf)
\`\`\`

\`logf\` is passed into \`DynamicMultiWriter\` as the always-on capture writer. \`lifecycle.go:198-280\` (\`Close()\`) tears down ptmx, listener, clients, subscribers — but never closes \`logf\`. The fd outlives the runner instance.

Process-exit reclaims it, so production single-shot runners are fine. Tests that construct and \`Close\` multiple runners in-process — or any future server that recycles runner instances — accumulate one leaked fd per cycle.

Companion observation: \`logf\` is held only as an opaque \`io.Writer\` inside \`DynamicMultiWriter\`, so a fix needs to either (a) keep an \`*os.File\` reference on \`Exec\` for Close to call, or (b) teach \`DynamicMultiWriter\` (or a typed sub-API) to close writers that implement \`io.Closer\`.

## Proposal

Option A — simplest, narrow blast radius:

\`\`\`go
// terminal_runner_exec.go: Exec
captureFile *os.File // owned by runner, closed in Close
\`\`\`

Assign in \`startPty\` after \`OpenFile\` succeeds; in \`Close()\`, call \`sr.captureFile.Close()\` *after* \`closeAllSubscribers\` and ptmx teardown (so any in-flight \`multiOutW.Write\` has settled).

Option B — generalize \`DynamicMultiWriter\` to also support closeable writers — broader API change; defer unless a second leak surfaces.

## Acceptance criteria

- [ ] \`Exec.Close()\` closes the capture file descriptor exactly once
- [ ] Repeated NewTerminalRunnerExec → Start → Close in a loop does not accumulate fds for the capture file (verify via \`/proc/self/fd\` count delta)
- [ ] Double-close is harmless (idempotency)

## Out of scope

- Generalizing fd-ownership for \`DynamicMultiWriter\` writers (Option B); revisit if a second leak appears.
- Log-file (not capture-file) ownership; \`SetupFileLogger\` creates that handle in the parent process and the runner does not own it.

## References

- Audit issue: #219
- Related pipe fd leak fix: PR #216 (same audit lens)

eminwux:
## Summary

- `Exec.Close()` now closes the capture file fd opened in `startPty`. Pre-fix, `logf` was handed to `DynamicMultiWriter` as an opaque `io.Writer` and the runner never closed it — process-exit reclaimed the fd in production single-shot runners, but in-process recycle (tests, future server) accumulated one leaked fd per cycle.
- Implements **Option A** from the issue (narrow blast radius): `*os.File` field on `Exec`, idempotent close via `*sync.Once`. Option B (teaching `DynamicMultiWriter` about `io.Closer` writers) is deferred per the issue's "Out of scope" note.
- Close sequencing matches the issue's guidance — capture fd closes **after** `closeAllSubscribers` and ptmx teardown, so any in-flight `multiOutW.Write` from the PTY reader has settled before the fd disappears.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make sbsh-sb` + `file ./sbsh` (ELF executable confirmed)
- [x] `make e2e`
- [x] `pre-commit run --files <changed>`
- [x] New regression tests in `lifecycle_capture_fd_test.go`:
  - `TestExecClose_CapturesFileFD` — direct close verification (second `*os.File.Close()` returns `os.ErrClosed`)
  - `TestExecClose_CaptureFileDoubleCloseIsHarmless` — `sync.Once` guards double-close
  - `TestExecClose_CaptureFileFDNoAccumulation` — loop 50× and assert `/proc/self/fd` delta stays under noise threshold (Linux-only; skips elsewhere)

Closes #229

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -263,6 +263,18 @@ func (sr *Exec) Close(reason error) error {
 		}
 	}
 
+	// Close the capture file fd. Sequenced after closeAllSubscribers and
+	// ptmx teardown so the PTY reader has stopped writing to multiOutW
+	// (which holds captureFile as its always-on writer). closeCapture
+	// guards against double-Close per #229's idempotency AC.
+	if sr.captureFile != nil && sr.closeCapture != nil {
+		sr.closeCapture.Do(func() {
+			if err := sr.captureFile.Close(); err != nil {
+				sr.logger.Warn("could not close capture file", "id", sr.id, "err", err)
+			}
+		})
+	}
+
 	// remove Ctrl socket
 	sr.metadataMu.RLock()
 	socketFile := sr.metadata.Status.SocketFile
--- a/internal/terminal/terminalrunner/lifecycle_capture_fd_test.go
+++ b/internal/terminal/terminalrunner/lifecycle_capture_fd_test.go
@@ -1,0 +1,158 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package terminalrunner
+
+import (
+	"context"
+	"errors"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"runtime"
+	"sync"
+	"testing"
+
+	"github.com/eminwux/sbsh/pkg/api"
+)
+
+// newCaptureFDExec builds a minimal Exec wired with a real *os.File as the
+// owned capture-file handle. Mirrors newCloseRaceExec in shape; the only
+// extra step is opening the temp capture file and assigning it through the
+// same path startPty does after applyArtifactPerms succeeds.
+func newCaptureFDExec(t *testing.T) (*Exec, *os.File) {
+	t.Helper()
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #240 Review — `fix(terminalrunner): close capture file fd in Close to prevent leak`**

Tight, well-scoped fix that matches issue #229's Option A. Close-time sequencing (`closeAllSubscribers` → `closePTY.Do` → `closeCapture.Do`) is correct, the `sync.Once` guards the AC's idempotency requirement, and the three-test regression set covers direct close, double-close, and the fd-accumulation loop. No secret/PII surface. Per-commit signatures verified. CI green.

Two non-blocking observations:

- **Pre-existing leak** `internal/terminal/terminalrunner/terminal.go:206-219`: after `sr.captureFile = logf` (line 206), error returns from `updateMetadata()` (line 209) or `os.Pipe()` (line 215) still leak the fd because the caller never reaches `Close`. Pre-existing — predates this PR, which only widens the happy-path fix — but worth closing via `sr.closeCapture.Do(...)` on those returns now that the fd is struct-owned.
- **Latent `closedCh` double-close panic** `internal/terminal/terminalrunner/lifecycle.go:301`: `close(sr.closedCh)` runs unguarded on every `Close`, so a real second `Close` panics. The test at `lifecycle_capture_fd_test.go:92-95` honestly flags it as "orthogonal" and resets `closedCh` to keep the fd-idempotency assertion clean — acceptable given the PR's narrow scope, but the broader runner-`Close` idempotency stays unfixed.

LGTM.

PR #243, opened 2026-05-18
fix(terminalrunner): tear down attacher outWriters in Close to drain goroutines

### LINKED ISSUE #230: bug(terminalrunner): Close() does not explicitly tear down attacher outWriters in multiOutW (latent race)
## Context

Surfaced by audit #219.

`internal/terminal/terminalrunner/lifecycle.go:232-240` (\`Close()\`):

\`\`\`go
sr.clientsMu.Lock()
for _, c := range sr.clients {
    if err := c.conn.Close(); err != nil { ... }
}

sr.clients = nil   // see also: panic issue tracked separately
sr.clientsMu.Unlock()
\`\`\`

The teardown closes each client's \`conn\` but does *not*:
- \`sr.ptyPipes.multiOutW.Remove(c.outWriter)\`
- \`c.outWriter.Close()\` (wakes the drain goroutine)

These are the exact two steps \`cleanupClient\` (\`connections.go:34-70\`) and the \`Detach\` RPC path (\`lifecycle.go:320-362\`) perform on every other detach path. Close skips them.

Today this still cleans up, *indirectly*: the dualcopier reader-side goroutine in \`handleClient\` notices the conn close, fires its finish callback, which fires \`detach()\` → \`cleanupClient\` → \`multiOutW.Remove\` + \`outWriter.Close\`. So the drain goroutines eventually wake and exit.

Why it's worth filing anyway:
1. The indirection makes the shutdown ordering hard to reason about — three goroutines (dualcopier reader, CopierManager, the drain) all need to converge through \`detachOnce\` before the client is fully gone, and they race the parallel \`Close()\` paths (\`closeAllSubscribers\`, ptmx close, ctx cancel).
2. \`cleanupClient\` calls \`updateTerminalAttachers()\` — a metadata write — which now happens *during* runner shutdown, racing \`Close()\`'s own metadata edits.
3. If a future refactor changes the dualcopier finish-func semantics or removes the reader-side conn-close detection, the drain goroutines silently start leaking with no test to flag it.

## Proposal

Make Close's client teardown symmetric with the per-client paths:

\`\`\`go
sr.clientsMu.Lock()
for _, c := range sr.clients {
    if c.outWriter != nil {
        sr.ptyPipes.multiOutW.Remove(c.outWriter)  // single-owner of fan-out membership
        _ = c.outWriter.Close()                    // wake drain
    }
    _ = c.conn.Close()
}
sr.clientsMu.Unlock()
\`\`\`

Note: this interacts with the separate \`sr.clients = nil\` panic issue — bundle the two fixes if convenient (both touch the same block).

## Acceptance criteria

- [ ] \`Close()\` explicitly removes each client's outWriter from \`multiOutW\` and calls \`Close()\` on it
- [ ] Drain goroutines have observably exited by the time \`Close()\` returns (or shortly after, via the existing detachOnce gating)
- [ ] Regression test: runner with N attachers, call Close, then assert goroutine count returns to baseline
- [ ] Metadata writes from \`cleanupClient\` and \`Close\` don't race-clobber each other

## Out of scope

- The \`sr.clients = nil\` panic — tracked separately.
- The subscriber-side equivalent (\`closeAllSubscribers\` already handles it correctly).

## References

- Audit issue: #219
- Existing per-client teardown: \`connections.go:cleanupClient\` and \`lifecycle.go:Detach\`

eminwux:
## Summary

- `Close()` (`internal/terminal/terminalrunner/lifecycle.go`) previously closed only `c.conn` per client and left each attacher's `subscriberWriter` registered in `multiOutW` with its drain goroutine parked on `cond.Wait`. The drain only exited indirectly via the dualcopier reader-side `onClose` → `detach` → `cleanupClient` → `aw.Close` path. If a future refactor changed that indirection's semantics, drains would silently leak with no test to flag it.
- Fix makes Close's teardown symmetric with `cleanupClient` and the `Detach` RPC path: drop the attacher's bounded writer from the fan-out, `Close()` it to wake the drain, then close the conn.
- `multiOutW` is snapshotted under `ptyPipesMu.RLock()` *outside* `clientsMu` because `cleanupClient` takes `ptyPipesMu` then `clientsMu` (via `removeClient`); acquiring them in the opposite order in `Close` would invert the lock graph.
- The `clear(sr.clients)` from #225 is preserved.

## Notes for reviewer

- **AC #4 ("Metadata writes from `cleanupClient` and `Close` don't race-clobber each other").** The fix does not introduce a new metadata write per client. The drain goroutine's `defer onDetach()` still runs `cleanupClient` post-Close — same as today's indirect path — and is gated by the existing `detachOnce`; the second `Remove`/`Close` calls are no-ops, `removeClient` finds the map already cleared, and `updateTerminalAttachers` fires exactly once per client (same as pre-fix). The race exists in main and is unchanged by this PR; out of scope here.
- The fix follows the issue's literal Proposal block. The bundled-with-#225-panic option mentioned in the issue is already in main, so this PR only adds the symmetric teardown.

## Test plan

- [x] `make test` (passes; includes the new `TestExecClose_TearsDownAttacherOutWriters`)
- [x] `go test -race ./internal/terminal/terminalrunner/...`
- [x] `make e2e` (passes)
- [x] Regression verified: temporarily reverted the `lifecycle.go` change and re-ran the new test — fails with `baseline=2, got=7` (5 leaked drain goroutines). Restored, test passes.
- [x] `pre-commit run --files <changed>` — all hooks pass

Closes #230

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -236,13 +236,36 @@ func (sr *Exec) Close(reason error) error {
 
 	// close clients
 	//
+	// Snapshot multiOutW outside clientsMu: cleanupClient takes
+	// ptyPipesMu then clientsMu (via removeClient), so acquiring them in
+	// the opposite order here would invert the lock graph.
+	//
+	// Per-client teardown mirrors cleanupClient (connections.go) and the
+	// Detach RPC path: drop the attacher's bounded writer from the
+	// fan-out, Close it to wake its drain goroutine, then close the conn.
+	// Pre-fix Close only closed conn and relied on the dualcopier
+	// reader-side onClose firing detach to indirectly drain the
+	// goroutine; that path still runs (idempotently, gated by
+	// detachOnce + clear() below), but the indirect dependency is no
+	// longer load-bearing for goroutine teardown. See #230.
+	//
 	// clear() instead of `sr.clients = nil`: a concurrent addClient call
 	// can race past Close's unlock (per-conn RPC handlers outlive the
 	// accept loop's ctx-cancel exit), and writing to a nil map panics.
 	// Emptying the map keeps it addressable so the late write is a benign
 	// stale-entry no-op shape. See #225.
+	sr.ptyPipesMu.RLock()
+	multiOutW := sr.ptyPipes.multiOutW
+	sr.ptyPipesMu.RUnlock()
+
 	sr.clientsMu.Lock()
 	for _, c := range sr.clients {
+		if c.outWriter != nil {
+			if multiOutW != nil {
+				multiOutW.Remove(c.outWriter)
+			}
+			_ = c.outWriter.Close()
+		}
 		if err := c.conn.Close(); err != nil {
 			sr.logger.Warn("could not close client connection", "id", sr.id, "client", c.id, "err", err)
 		}
--- a/internal/terminal/terminalrunner/lifecycle_close_drain_test.go
+++ b/internal/terminal/terminalrunner/lifecycle_close_drain_test.go
@@ -1,0 +1,95 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #243 Review — `fix(terminalrunner): tear down attacher outWriters in Close to drain goroutines`**

Clean, focused fix for the drain-goroutine leak in #230. The new per-client teardown mirrors `Detach` (`lifecycle.go:374-383`) and `cleanupClient` (`connections.go:36-60`), so all three paths now share the same Remove → outWriter.Close → conn handoff. Lock-order claim verified: `cleanupClient` takes `ptyPipesMu` (RLock/RUnlock at `connections.go:39-41`) before `clientsMu` (via `removeClient` at `connections.go:160`); the new `Close` snapshots `multiOutW` outside `clientsMu` the same way, so no inversion. The regression test (`lifecycle_close_drain_test.go`) explicitly probes the parked-drain failure mode — `time.Sleep(20ms)` ensures each `Run` reaches `cond.Wait` before Close, and the baseline+1 goroutine assertion is what would have caught the leak before #230. Commits verified, all checks green, no secrets in new log lines.

LGTM.

PR #245, opened 2026-05-18
fix(terminalrpc): consume params on nil-body to prevent cross-talk

### LINKED ISSUE #220: audit(terminalrpc): SCM_RIGHTS and fd-transfer sweep
## Context
#213 fixed a sender-side fd leak after `SCM_RIGHTS` send in
`internal/terminal/terminalrpc/`. The bug pattern — "sender keeps its copy
after the kernel duplicates the fd to the receiver" — is easy to repeat
across every code path that calls `sendmsg` with `SCM_RIGHTS`. Worth a
sweep of the package before the next bug lands the same way.

## Proposal
Read `internal/terminal/terminalrpc/codec.go` and `rpc_server.go` end-to-end
with focus on:
- every `SCM_RIGHTS` send site: does the sender close its copy?
- every `SCM_RIGHTS` recv site: does the receiver own and eventually close?
- partial-send / partial-recv error paths: do fds leak on failure?
- codec edge cases: oversize messages, truncated reads, EOF mid-frame
- server-side handler lifetimes and concurrent-request handling

For each defect: file a separate `bug` issue, link back here. Small
obvious fixes land in one PR against this audit. Larger fixes route through
the per-bug follow-ups.

## Acceptance criteria
- [ ] All `SCM_RIGHTS` send/recv sites in the package have been verified
- [ ] Error / partial-IO paths verified for fd-leak safety
- [ ] Concurrent-request handling in `rpc_server.go` verified

## Notes
Recent neighbor: #213.

eminwux:
## Summary

Audit of `internal/terminal/terminalrpc/` for the SCM_RIGHTS fd-leak pattern that landed in #213, plus a small obvious fix for the one defect found.

- **The SCM_RIGHTS send site (only one) is sound.** `codec.go`'s `WriteResponse` defer-closes the local fds on every exit path (success, marshal error, `WriteMsgUnix` error) and nils out `r.FDs` for idempotency. `UnixRights` does not consume fds, so no leak there. There are no SCM_RIGHTS recv sites in the package (server-side codec only reads JSON; client-side recv lives in `pkg/rpcclient/terminal/`, out of scope for this issue).
- **`rpc_server.go` is pure delegation** — no fd handling, no shared state, no concurrency primitives. Concurrent-request handling is owned by `net/rpc`'s per-call goroutine dispatch.
- **The codec carries an unrelated defect** unrelated to SCM_RIGHTS but in the "codec edge cases" scope of #220: `ReadRequestBody(nil)` returned without consuming the entry queued in `paramsBySeq` by the matching `ReadRequestHeader`. `net/rpc` calls `ReadRequestBody(nil)` to discard the body of an unknown method, so every unknown-method call with non-empty params (a) leaked one buffer for the connection's lifetime and (b) caused **cross-talk** on the next call: `ReadRequestBody` picks the smallest-seq entry, which would be the stale prior-request one, so a later legitimate request's handler would receive the prior request's params. In practice clients and servers know the same method set so this is rarely triggered, but the leak/cross-talk is real on any version skew or malicious client.

Fix: move the `paramsBySeq` consume block above the `body == nil` check so the entry is always drained.

Regression test: `internal/terminal/terminalrpc/codec_test.go` drives two requests through the codec, the first an "unknown" method with `ReadRequestBody(nil)`, the second a known method with distinct params; asserts the second handler receives its own params (verified to fail with `body.X = 111; want 222` and `paramsBySeq has 1 leaked entries; want 0` when the fix is reverted).

## Test plan

- [x] `make sbsh-sb` (canonical build produces ELF for both `sbsh` and `sb`)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full project test target — cmd, internal, e2e)
- [x] `go test ./internal/terminal/terminalrpc/...` (includes the two new regression tests)
- [x] Regression tests verified to fail with the codec fix reverted (cross-talk on test 1, leak on test 2)
- [x] `pre-commit run --files` on changed files

Closes #220

### DIFF
--- a/internal/terminal/terminalrpc/codec.go
+++ b/internal/terminal/terminalrpc/codec.go
@@ -88,13 +88,14 @@ func (c *unixJSONServerCodec) ReadRequestHeader(r *rpc.Request) error {
 
 // ReadRequestBody unmarshals the first element of the params array into body (like std jsonrpc).
 func (c *unixJSONServerCodec) ReadRequestBody(body interface{}) error {
-	if body == nil {
-		return nil
-	}
 	// We don't get seq here, so we rely on json.Decoder’s sequencing; net/rpc calls
 	// ReadRequestBody immediately after ReadRequestHeader on the same goroutine,
 	// so the most recent seq is r.Seq. To keep this simple, we pop the smallest seq
-	// that still has params (the one we just set).
+	// that still has params (the one we just set). We must consume the entry even
+	// when body is nil — net/rpc passes nil to discard the body of an unknown
+	// method, and leaving the entry behind would both leak the buffer for the
+	// connection's lifetime and cause cross-talk on the next call (smallest-seq
+	// pick would return the stale prior-request params).
 	c.seqMu.Lock()
 	var chosenSeq uint64
 	for s := range c.paramsBySeq {
@@ -105,6 +106,10 @@ func (c *unixJSONServerCodec) ReadRequestBody(body interface{}) error {
 	params := c.paramsBySeq[chosenSeq]
 	delete(c.paramsBySeq, chosenSeq)
 	c.seqMu.Unlock()
+
+	if body == nil {
+		return nil
+	}
 
 	if len(params) == 0 || string(params) == "null" {
 		return nil
--- a/internal/terminal/terminalrpc/codec_test.go
+++ b/internal/terminal/terminalrpc/codec_test.go
@@ -1,0 +1,132 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #245 Review — `fix(terminalrpc): consume params on nil-body to prevent cross-talk`**

Targeted fix backed by sharp tests. The reorder — drain `paramsBySeq` before the `body == nil` early-return — is the right shape: every `ReadRequestBody` now owns the entry its matching `ReadRequestHeader` enqueued, and the smallest-seq pop becomes safe again because no stale predecessor can sit ahead of the next request's entry. Cross-checked the single production codec caller (`internal/terminal/terminalrunner/control_server.go:101`) — signature unchanged, no public API churn. Tests reproduce both the cross-talk (`body.X` must be 222) and the leak (`paramsBySeq` must be 0), and the PR confirms both fail when the fix is reverted. Audit conclusions on the SCM_RIGHTS send site (`codec.go` `WriteResponse` defer-closes + nils `r.FDs`) and the pure-delegation `rpc_server.go` match what I see.

No blocking items.

LGTM.

PR #247, opened 2026-05-18
fix(pkg/attach): return ErrContextDone deterministically on ctx-cancel

### LINKED ISSUE #221: audit(pkg/attach): multi-attacher fan-out, slow-consumer, detach-race sweep
## Context
`pkg/attach/` is the public attach entry point and pairs tightly with the
fan-out writer that #217 just fixed in `terminalrunner`. #193 added
`ErrDetached` / `ErrPeerClosed` sentinels so callers can disambiguate
clean detach from peer hangup. The combination — multiple concurrent
attachers, midstream detach, slow consumers — is a classic place for
goroutine leaks, dropped output, and detach races.

## Proposal
Read `pkg/attach/attach.go` and `pkg/attach/errors.go` (with the test files
for context) with focus on:
- multiple concurrent attachers: shared state, fan-out correctness
- slow / blocked consumer behavior: does it back-pressure or drop?
- detach midstream: clean teardown vs. leaked goroutines or fds
- `ErrDetached` / `ErrPeerClosed` / `ErrSocketPathRequired`
  discrimination paths — are callers actually able to branch on them?
- error paths from `Run` that skip cleanup

For each defect: file a separate `bug` issue, link back here. Small fixes
in one PR. Larger ones get their own issues.

## Acceptance criteria
- [ ] All exported entry points in `pkg/attach/` read end-to-end
- [ ] Fan-out behavior with multiple attachers verified
- [ ] Slow-consumer and detach-midstream paths verified
- [ ] Sentinel-error discrimination verified from a caller's perspective

## Notes
Recent neighbors: #217 (fan-out fix), #193 (sentinels), #175 (sentinel
export), #173 (synthesize id for SocketFile-only attach).

### LINKED ISSUE #246: bug(pkg/attach): Run shutdown select drops ErrContextDone half the time on ctx cancellation
## Context

`pkg/attach.Run` (`pkg/attach/attach.go:153-173`) is the package's only entrypoint. Its docstring (lines 67-74) promises:

> Run … returns when the session ends. The loop terminates cleanly on:
>   - context cancellation (returns ctx.Err wrapped in errdefs.ErrContextDone),
>   - the remote terminal closing the connection,
>   - the embedded client's detach keystroke firing (when enabled),
>   - any unrecoverable error from the underlying controller …

Surfaced during the `pkg/attach` audit (#221).

## Defect

The shutdown select is:

```go
select {
case <-ctx.Done():
    if waitErr := ctrl.WaitClose(); waitErr != nil { … }
    <-errCh
    return fmt.Errorf("%w: %w", errdefs.ErrContextDone, ctx.Err())

case ctrlErr := <-errCh:
    if ctrlErr == nil || errors.Is(ctrlErr, context.Canceled) {
        return nil
    }
    _ = ctrl.WaitClose()
    return classifySessionEnd(ctrlErr)
}
```

When the embedder cancels `ctx`, the controller's own `s.ctx` (derived via `context.WithCancelCause` in `internal/client/controller.go:87`) is cancelled too. The controller observes that in its main loop (`internal/client/controller.go:270-289`), wraps the cause as `errdefs.ErrContextDone: <ctx.Err>`, and sends on `errCh`. Both select cases become ready at essentially the same moment, and Go's select chooses at random:

- **`<-ctx.Done()` wins (~50%):** returns `errdefs.ErrContextDone: <ctx.Err>` — the documented behavior.
- **`ctrlErr := <-errCh` wins (~50%):** falls into the `errors.Is(ctrlErr, context.Canceled)` short-circuit and returns **nil** — silently violating the docstring contract.

Embedders branching on `errors.Is(err, errdefs.ErrContextDone)` or `errors.Is(err, context.Canceled)` therefore get inconsistent results between runs of the same code path under the same input. Worse, the nil-return case makes a cancelled session look indistinguishable from a clean peer-closed session that returned nil too — wrong exit code, wrong telemetry, wrong UI message.

## Adjacent (same fix surface)

The same select block has a second weakness worth fixing in the same PR: when `ctrl.WaitReady()` returns non-nil (which only happens when `ctx.Done()` fires before `ctrlReadyCh` closes), the code does a bare `<-errCh` with no defensive `ctrl.Close(waitErr)` trigger. The downstream code currently bounds every setup-time RPC (`pkg/rpcclient/terminal/rpc_session.go:67-145` `callWithCodec` selects on ctx; `internal/client/clientrunner/io.go:50,172,201` use ctx-bound `WithTimeout`), so the bare receive doesn't deadlock today — but the contract `pkg/attach.Run` consumes from the controller is "Run will eventually exit on ctx-cancel," and that contract isn't actively enforced from this side. The same active-close treatment applies to the `ctx.Done` arm: `ctrl.WaitClose()` blocks on `closedCh` which is closed only by `Close(reason)`. Adding an explicit `ctrl.Close(reason)` before the wait makes both shutdown waits guaranteed-finite regardless of future setup-step refactors.

## Proposal

Restructure `pkg/attach.Run`'s shutdown handling to:

1. Drop the `errors.Is(ctrlErr, context.Canceled)` short-circuit in the `errCh` arm so a ctx-cancelled `ctrl.Run` returns its wrapped `errdefs.ErrContextDone` error through `classifySessionEnd` (which is a no-op for that sentinel — passes it through unchanged). Both select arms then return the same `errdefs.ErrContextDone: <ctx.Err>` shape.
2. On the `WaitReady`-fail and `ctx.Done` arms, call `ctrl.Close(reason)` explicitly before draining. `Close` is idempotent (`shuttingDown.CompareAndSwap`) and self-contained — it cancels the controller's ctx, closes `closedCh`, and lets the waits return without depending on downstream setup steps observing ctx.

## Acceptance criteria

- [ ] `pkg/attach.Run`'s ctx-cancellation path returns `errdefs.ErrContextDone` wrapping `ctx.Err()` deterministically, regardless of select race.
- [ ] Regression test: cancel `ctx` while `Run` is mid-session against the existing `fakeTerminalController` fixture; assert `errors.Is(err, errdefs.ErrContextDone)` and `errors.Is(err, context.Canceled)`. Run the test with `-count=100` (or comparable) to confirm non-flakiness.
- [ ] Existing tests in `pkg/attach/attach_test.go` continue to pass (sentinel discrimination, nil-logger, missing-socket, attach-failure).
- [ ] `WaitReady`-fail and `ctx.Done` arms actively trigger `ctrl.Close(reason)` before waiting, so the shutdown is finite even if a future setup-step refactor introduces a non-ctx-aware blocking call.

## Out of scope

- The `controller.WaitClose() error` signature in `internal/client/controller.go:448-454` is currently a single-case select that always returns nil — making its error return unreachable, which in turn makes `pkg/attach/attach.go:155-156` dead code. Worth a separate cleanup PR; not blocking this fix.
- Fan-out, slow-consumer, detach-midstream behaviors live in `internal/terminal/terminalrunner` (the package #217 hardened) and `internal/client/clientrunner/io.go`, not in `pkg/attach`. Out of scope for this issue.

## References

- Audit issue: #221.
- Adjacent fan-out hardening: #217 (referenced in #221's body).
- Sentinel-export work: #193, #175.
- `pkg/attach/attach.go:79-188` (the entrypoint).
- `internal/client/controller.go:124-321` (controller `Run` main loop) and `:426-454` (`Close`/`WaitClose`).
- `pkg/rpcclient/terminal/rpc_session.go:67-145` (downstream ctx-aware RPC path).

eminwux:
## Summary

- Addresses the `pkg/attach/` audit (#221) and the bug it surfaced (#246). `pkg/attach.Run`'s shutdown handling had two paths returning different sentinels for the same observable event (ctx cancellation) — the post-setup select arm short-circuited on `errors.Is(ctrlErr, context.Canceled)` and returned **nil** ~50% of the time, while the setup-phase WaitReady arm returned `errdefs.ErrWaitOnReady`. The docstring (lines 67-74) promises one sentinel — `errdefs.ErrContextDone` — for any ctx-cancel.
- Both arms now (a) actively call `ctrl.Close(reason)` so the shutdown is guaranteed-finite regardless of whether downstream setup steps observe ctx, and (b) return `errdefs.ErrContextDone` wrapping `ctx.Err()`. The `WaitClose` errors-out branch on the ctx-done arm was already unreachable (the controller's `WaitClose` is a single-case select on `closedCh`) — removed to drop dead code.
- Note for review: the public re-export `pkg/errors.ErrWaitOnReady` (`pkg/errors/errors.go:80`) is unchanged — that sentinel is still produced by the separate per-terminal wait paths in `cmd/sbsh/terminal/terminal.go:549` and `cmd/sbsh/sbsh.go:494`. Only the `pkg/attach.Run` arm that previously emitted it on ctx-cancel-during-setup has switched to `ErrContextDone` to satisfy the unified-sentinel docstring.

## Audit verdict for #221

`pkg/attach/` is a thin façade (4 files, ~220 LOC). End-to-end read confirms:

- **Exported entry points:** only `Run(ctx, Options) error`. Options struct + three sentinels (`ErrSocketPathRequired`, `ErrDetached`, `ErrPeerClosed`) — covered.
- **Fan-out behavior with multiple attachers:** not a `pkg/attach` concern. Each `Run` call creates its own private control socket under `os.TempDir()` and dials the target. Multi-attacher fan-out lives in `internal/terminal/terminalrunner` — the package PR #217 hardened. Documented correctly in `pkg/attach/doc.go` and `attach.go:75-78`.
- **Slow-consumer behavior:** not introduced at this layer either. `pkg/attach` passes the caller's `*os.File` stdout straight through to the controller; bounded-output policy + per-write deadlines live in `terminalrunner.subscriberWriter` and the server-side attach path (#217).
- **Detach-midstream:** the in-band ^]^] keystroke is the embedded client's responsibility (`internal/client/clientrunner/io.go:96-103`). `pkg/attach.classifySessionEnd` maps the resulting `errdefs.ErrClientDetached` (set by `controller.classifySessionEnd` once `detachRequested` flips) to `attach.ErrDetached`. Test coverage already in `attach_test.go`.
- **Sentinel-error discrimination:** `ErrSocketPathRequired`, `ErrAttach`, `ErrDetached`, `ErrPeerClosed`, and the setup-error pass-through are exercised; this PR adds the missing `ErrContextDone` deterministic-return test.
- **Defect found:** the shutdown-select race + setup-arm sentinel mismatch documented in #246. Filed as a separate `bug` per the audit body. Fix included in this PR.

Out-of-scope observation surfaced during the audit (not filed): `internal/client/controller.go:448-454` `WaitClose() error` is a single-case select that can only return nil — the `error` return is unreachable and the corresponding caller check at `pkg/attach/attach.go:155-156` (pre-fix) was dead code. Drive-by removed from this PR's diff; a follow-up cleanup PR could either tighten the signature to `WaitClose()` (no error) or implement the timeout-aware path the signature implies, but neither is blocking.

## Test plan

- [x] `file ./sbsh` — confirms ELF 64-bit LSB executable.
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full unit suite passes, including the new regression `TestRun_CtxCancelReturnsErrContextDone` in `pkg/attach/attach_test.go`.
- [x] `go test -count=10 -run '^TestRun_CtxCancel' ./pkg/attach/` — new regression non-flaky across 10 runs.
- [x] `go test -tags=integration ./cmd/sb/get/...` — integration step passes.
- [x] `cd docs/examples/library-consumer && go build ./... && go vet ./...` — library-consumer example builds and vets clean.
- [x] `make e2e` — full e2e suite passes.
- [x] `pre-commit run --files pkg/attach/attach.go pkg/attach/attach_test.go` — all hooks pass.

Closes #221
Closes #246

### DIFF
--- a/pkg/attach/attach.go
+++ b/pkg/attach/attach.go
@@ -144,29 +144,41 @@ func Run(ctx context.Context, opts Options) error {
 	}()
 
 	if waitErr := ctrl.WaitReady(); waitErr != nil {
-		// Drain ctrl.Run so the goroutine can exit; ignore its error
-		// (already surfaced via WaitReady or closing).
+		// WaitReady only returns non-nil when ctx fires before the
+		// controller signals readiness — i.e. ctx-cancel during setup.
+		// Actively close so the goroutine can exit even if a setup step
+		// is blocked on something that doesn't observe ctx; without
+		// this, <-errCh would have to trust every downstream RPC to
+		// honour ctx-cancel. Return the same ErrContextDone shape the
+		// post-setup path uses so callers branching on errors.Is see
+		// one sentinel for any ctx-cancel.
+		_ = ctrl.Close(waitErr)
 		<-errCh
-		return fmt.Errorf("%w: %w", errdefs.ErrWaitOnReady, waitErr)
+		return fmt.Errorf("%w: %w", errdefs.ErrContextDone, waitErr)
 	}
 
 	select {
 	case <-ctx.Done():
-		if waitErr := ctrl.WaitClose(); waitErr != nil {
-			return fmt.Errorf("%w: %w", errdefs.ErrWaitOnClose, waitErr)
-		}
+		// Force shutdown for the same reason as the WaitReady path, and
+		// so WaitClose (which blocks on closedCh, which only Close
+		// closes) doesn't depend on the controller reaching its own
+		// ctx.Done observer before exiting.
+		ctxErr := ctx.Err()
+		_ = ctrl.Close(ctxErr)
+		_ = ctrl.WaitClose()
 		<-errCh
-		return fmt.Errorf("%w: %w", errdefs.ErrContextDone, ctx.Err())
+		return fmt.Errorf("%w: %w", errdefs.ErrContextDone, ctxErr)
 
 	case ctrlErr := <-errCh:
-		if ctrlErr == nil || errors.Is(ctrlErr, context.Canceled) {
+		if ctrlErr == nil {
 			return nil
 		}
 		// Drain WaitClose to release internal resources, then map the
 		// controller-level session-end sentinel to the matching public
-		// sentinel so embedders can branch with errors.Is. Setup and
-		// infrastructure failures (no errdefs session-end sentinel
-		// chained) keep their identity.
+		// sentinel so embedders can branch with errors.Is. ctx-cancel
+		// surfaces here as errdefs.ErrContextDone chained with ctx.Err
+		// — pass it through unchanged so the two arms of this select
+		// return the same shape regardless of which one wins the race.
 		_ = ctrl.WaitClose()
 		return classifySessionEnd(ctrlErr)
 	}
--- a/pkg/attach/attach_test.go
+++ b/pkg/attach/attach_test.go
@@ -49,6 +49,11 @@ type fakeTerminalController struct {
 	attachCalls atomic.Int32
 	state       api.TerminalStatusMode
 	attachErr   error
+	// attachBlock, if non-nil, makes Attach block until the channel is
+	// closed. Used to pin Run inside the controller's setup phase so
+	// the ctx-cancel race in pkg/attach.Run's shutdown select is
+	// reachable from a test.
+	attachBlock <-chan struct{}
 }
 
 func (f *fakeTerminalController) Ping(in *api.PingMessage, out *api.PingMessage) error {
@@ -69,6 +74,9 @@ func (f *fakeTerminalController) State(_ *api.Empty, out *api.TerminalStatusMode
 
 func (f *fakeTerminalController) Attach(_ *api.ID, _ *api.Empty) error {
 	f.attachCalls.Add(1)
+	if f.attachBlock != nil {
+		<-f.attachBlock
+	}
 	return f.attachErr
 }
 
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #247 Review — `fix(pkg/attach): return ErrContextDone deterministically on ctx-cancel`**

Clean fix for the audit-surfaced race in #246: both ctx-cancel paths (WaitReady-during-setup and post-setup `ctx.Done`) now actively `ctrl.Close` and return `errdefs.ErrContextDone` wrapping `ctx.Err()`, matching the docstring at `pkg/attach/attach.go:67-74`. The errCh arm correctly drops the `errors.Is(ctrlErr, context.Canceled)` short-circuit — the controller's own `ctx.Done` branch already returns `ErrContextDone` chained with `ctx.Err()`, and the new `classifySessionEnd` default-passthrough preserves that sentinel identity. No public-API regressions: `pkg/attach.Run` signature unchanged, sentinels (`ErrSocketPathRequired`/`ErrDetached`/`ErrPeerClosed`) untouched, and `pkg/errors.ErrWaitOnReady` re-export stays valid for `cmd/sbsh/{sbsh,terminal}` which produce that sentinel from their own paths. The drive-by removal of the unreachable `WaitClose` error-return branch is right — `internal/client/controller.go:451-457` confirms it's a single-case select on `closedCh` that can only return nil.

`TestRun_CtxCancelReturnsErrContextDone` is well-targeted: blocking the fake's `Attach` RPC pins the controller in setup phase (between `WaitReady` and the main loop) — exactly the race window where the pre-fix code raced and could return nil. The 10x non-flaky run is meaningful evidence.

All 4 commits signed and verified. All CI green (Build & Test, E2E, CodeQL, Analyze ×2). Test plan fully checked, with both the unit suite and the host-runnable e2e/`make e2e` exercised locally.

Nothing blocking, nothing nit-worthy.

LGTM.

Issue #252, opened 2026-05-18
bug(sb stop): --force SIGKILLs only the controller; child shell pgroup can be orphaned


eminwux:
## Context

Observed during the \`sb stop\` audit (#222).

\`TerminalStatus.Pid\` records the controller's \`os.Getpid()\` (\`internal/terminal/terminalrunner/metadata.go:43\`). The child shell is spawned with \`Setsid: true\` (\`internal/terminal/terminalrunner/terminal.go:90\`), so it lives in its own session/pgroup (child PID == pgid, different from the controller's). The controller's graceful teardown signals the child pgroup explicitly: \`syscall.Kill(-pid, SIGTERM)\` then \`SIGKILL\` after grace (\`internal/terminal/terminalrunner/lifecycle.go:184,199\`).

\`sb stop --force\` sends \`SIGKILL\` to the controller PID directly. SIGKILL is uncatchable: the controller does not run its deferred \`shutdownChild\`. The child pgroup is in a different session, so the kill does not propagate.

Mitigation in practice: when the controller dies, its open PTY master FD closes. The kernel sends \`SIGHUP\` to the foreground process group of the slave. Most shells exit on \`SIGHUP\`.

But: a shell that traps \`SIGHUP\` (\`trap '' HUP\`), a long-running foreground program that ignores \`SIGHUP\` (\`nohup\`), or any process that re-attached the slave for its own reasons remains alive and reparented to init.

The same hole exists in the regular graceful path's \`--kill-after\` escalation in \`cmd/sb/stop/terminals.go:293\`: when the controller's grace expires and the operator escalates to SIGKILL, only the controller dies. (The controller's own \`shutdownChild\` SIGKILL on the pgroup never runs because the controller is dead.)

## Proposal

1. **Record the child pgroup PID in metadata.** Add e.g. \`ChildPgid int\` to \`TerminalStatus\`. The controller captures \`sr.cmd.Process.Pid\` at startup (already a pgid because of \`Setsid: true\`) and writes it.
2. **\`sb stop --force\` (and \`--kill-after\` escalation) signal both:** \`syscall.Kill(-childPgid, SIGKILL)\` first, then \`Kill(controllerPid, SIGKILL)\`. Same fix in both the \`--force\` and \`--kill-after\` branches.
3. **Document the change in \`sb stop --help\`** (e.g. \`--force\` now reaps the entire pgroup, not just the controller).

## Acceptance criteria

[...]

Issue #251, opened 2026-05-18
bug(sb stop): Stop RPC retries spend ~9.6s on a stale-socket controller before SIGTERM fallback


eminwux:
## Context

Observed during the \`sb stop\` audit (#222).

\`pkg/rpcclient/terminal/rpc_session.go:callWithCodec\` retries 3 attempts with delays \`[0, 200ms, 400ms]\`. Each attempt's dial is bounded by \`stopRPCDialTimeout = 3 * time.Second\` (set in \`cmd/sb/stop/terminals.go:46\`). Worst-case total before \`rc.Stop\` returns an error: ~3s + 0.2s + 3s + 0.4s + 3s ≈ **9.6s**.

When the controller socket is stale (controller crashed, socket file persists on disk but no listener), every \`sb stop\` burns the full ~9.6s on RPC retries before falling back to SIGTERM. The user's \`--timeout\` only bounds the post-signal \`waitForExit\`, not the RPC dial budget. Consequence:

- \`sb stop --timeout 1s\` against a stale-socket terminal still takes ~10.6s wall-clock.
- \`sb stop --all\` is linear in the number of stale-socket terminals: N × 9.6s of retry waste.
- The retries help for transient \`EAGAIN\` / partial-bring-up blips on long-poll RPCs, but \`Stop\` is binary (responsive or not) — retrying buys nothing here.

## Proposal

Pick one (in rough order of blast radius):

1. **Lower-risk:** add a per-call ctx with deadline \`= opts.timeout\` (or a smaller dedicated dial budget) in \`stopViaRPC\`. Retries then respect the user's bound and stale-socket targets return inside the user's timeout window.
2. **Targeted:** add a \`WithMaxAttempts(int)\` option to \`pkg/rpcclient/terminal\` and set it to 1 in the stop path. Other callers (\`io.go\`, \`write\`, \`read\`) keep the retry default.
3. **Larger:** skip the RPC entirely when an \`os.Stat\` on the socket path returns \`ENOENT\`, or when a fast \`syscall.Connect\` with a 50ms deadline returns \`ECONNREFUSED\`/\`ENOENT\` — fall through to SIGTERM immediately.

(1) is the least invasive and aligns with the user's expectation that \`--timeout\` bounds the whole stop.

## Acceptance criteria

- [ ] \`sb stop --timeout 1s\` against a stale-socket terminal returns within ~2s.
[...]

PR #254, opened 2026-05-18
fix(stop): reject recycled PIDs in reconcile and reconcile single-name targets

### LINKED ISSUE #222: audit(sb stop): lifecycle teardown, signal handling, pid-reuse paths
## Context
`sb stop terminal` is the user-facing teardown path. Two existing open
issues already flag thin coverage in this area:
- #107 — test: deeper coverage for `sb stop terminal` paths
- #109 — stop: consider `pidfd_open` for kernel-enforced PID-reuse safety

Worth a focused bug-hunt before doing either of those, so the test cases in
#107 are informed and any pid-reuse decision in #109 lands on a verified
baseline.

## Proposal
Read `cmd/sb/stop/stop.go` and `cmd/sb/stop/terminals.go` end-to-end with
focus on:
- signal ordering: SIGTERM → wait → SIGKILL semantics, timeouts, races
- pid-reuse window: from "resolve pid" to "send signal", what guards exist?
- partial-failure paths: what if one terminal in a batch stop fails?
- interaction with the client terminal store — does state stay consistent
  on half-completed stops?
- exit-code semantics surfaced to the operator

For each defect: file a separate `bug` issue, link here. Land small fixes
in one PR; route larger ones through per-bug follow-ups. Findings should
feed #107's test plan if they surface gaps.

## Acceptance criteria
- [ ] `cmd/sb/stop/` read end-to-end with the bug lens
- [ ] Signal-ordering and timeout invariants verified
- [ ] Pid-reuse window identified and either ruled safe or filed as bug
- [ ] Partial-failure semantics on batch stop verified

## Notes
Related: #107 (test coverage), #109 (pidfd_open).

### LINKED ISSUE #248: bug(discovery/reconcile): isProcessAlive ignores PidStart; recycled PIDs keep stale metadata 'alive'
## Context

`internal/discovery/reconcile.go:isProcessAlive` checks only `syscall.Signal(0)` liveness against the recorded PID. It does not consult `TerminalStatus.PidStart` / `ClientStatus.PidStart`, the per-process start-time token the codebase already captures for exactly this purpose (`internal/pidutil`).

When a recorded PID has been recycled to an unrelated process — the foot-gun \`internal/pidutil\` was introduced to close — \`reconcileTerminal\` / \`reconcileClient\` see "something is alive at PID X" and skip the state-flip to \`Exited\`. The on-disk metadata then stays in the prior live state (e.g. \`Ready\`) indefinitely.

Observed by inspection during the \`sb stop\` audit (#222). The \`cmd/sb/stop/terminals.go:processIsOurs\` helper *does* perform the \`pidutil.Match\` check and catches this at signal time, so a recycled-PID terminal still gets reported correctly by \`sb stop\` as \`already exited\` — but reconcile itself remains blind, so:

- \`sb get terminals\` continues to list the stale terminal/client as live.
- \`sb prune terminals\` will not prune it (it only prunes \`Exited\`).
- \`sb stop --all\` includes it in the active set returned by \`resolveTargets\` and only filters it out per-terminal inside \`stopOneTerminal\`.

## Proposal

Thread \`PidStart\` through the reconcile path so it rejects recycled PIDs the same way \`processIsOurs\` already does in \`cmd/sb/stop/terminals.go\`. Concretely: change the aliveness predicate in \`internal/discovery/reconcile.go\` to consult \`pidutil.Match\` when a token is recorded (zero token → fall back to liveness only, matching \`processIsOurs\`'s degrade-gracefully contract). Apply to both \`reconcileTerminal\` and \`reconcileClient\`.

## Acceptance criteria

- [ ] \`internal/discovery/reconcile.go\` aliveness check rejects recycled PIDs when \`PidStart\` is set.
- [ ] Behavior degrades to liveness-only when \`PidStart == 0\` (legacy metadata / non-Linux).
- [ ] Unit test in \`internal/discovery/\` writes a doc whose \`Pid\` is alive but whose \`PidStart\` doesn't match, and asserts reconcile flips \`state\` to \`Exited\`.
- [ ] No regression in existing reconcile tests.

## Out of scope

- \`pidfd_open\` adoption (already tracked in #109).
- Reconcile concurrency / locking changes.

## References

- Audit: #222
- PID-reuse foundation: \`internal/pidutil/\`, \`pkg/api/terminal.go:111-115\` (\`PidStart\` doc), \`cmd/sb/stop/terminals.go:325-355\` (\`sendSignal\` / \`processIsOurs\` precedent).

### LINKED ISSUE #249: bug(sb stop): single-name target path doesn't reconcile metadata; stale state survives 'already exited'
## Context

\`cmd/sb/stop/terminals.go:resolveTargets\` calls \`discovery.ReconcileTerminals\` only on the \`--all\` branch. The single-target name path returns the doc straight from \`FindTerminalByName\` without reconciling. Observed during the \`sb stop\` audit (#222).

When the recorded process is dead, \`stopOneTerminal\` correctly reports \"already exited\" via \`processIsOurs\` — but the on-disk metadata is never updated to \`state=Exited\`. Until another command (\`sb get terminals\`, \`sb stop --all\`, \`sb prune terminals\`) triggers a reconcile, the metadata.json still reads e.g. \`Ready\`.

Effect on operators:
- Two \`sb stop foo\` invocations against a dead terminal both print \`already exited\` while the JSON keeps saying \`Ready\`.
- External tooling that reads metadata.json directly between \`sb stop foo\` and the next reconciliation sees lying state.
- Inconsistent with \`--all\`, which both reports correctly and persists.

## Proposal

In \`cmd/sb/stop/terminals.go:resolveTargets\`, after \`FindTerminalByName\` returns a non-nil doc, call the discovery package's reconcile entry point on the single doc (mirror the \`--all\` branch). The reconcile helper already persists \`state=Exited\` when the process is gone.

\`\`\`go
doc, err := discovery.FindTerminalByName(ctx, logger, opts.runPath, opts.name)
...
discovery.ReconcileTerminals(ctx, logger, opts.runPath, []api.TerminalDoc{*doc})
return []api.TerminalDoc{*doc}, nil
\`\`\`

(or expose the package-private \`reconcileTerminal\` and call that directly to avoid the slice copy).

**Note for the implementer:** this fix is most useful *after* #248 lands. Without #248, a recycled-PID terminal still won't be reconciled to \`Exited\` here either, because the reconcile helper currently treats PID-only liveness as proof of survival. Land #248 first or in the same PR.

## Acceptance criteria

- [ ] \`resolveTargets\` reconciles the single-target doc the same way \`--all\` does.
- [ ] Unit/integration test: write metadata with \`state=Ready\` and a dead PID, run the single-name stop path, assert metadata.json gets \`state=Exited\`.
- [ ] No regression in existing \`cmd/sb/stop\` tests.

## Out of scope

- Reconcile semantics changes (covered in #248).
- The \`pidfd_open\` swap (#109).

## References

- Audit: #222
- Related: #248 (reconcile PidStart blind spot).

eminwux:
## Summary

- **Closes #248**: `internal/discovery/reconcile.go:isProcessAlive` was PID-only. A recycled PID kept stale metadata stuck at the prior live state (e.g. `Ready`) — `sb get`, `sb prune`, and `sb stop --all` all saw it as live. New `isInstanceAlive` consults `pidutil.Match` when a token is recorded; zero-token metadata still falls back to liveness-only (matches `pidutil.Match`'s contract for legacy/non-Linux callers). Applied to both `reconcileTerminal` and `reconcileClient`.
- **Closes #249**: `cmd/sb/stop/terminals.go:resolveTargets` only ran reconcile on the `--all` branch. The single-target path now reconciles the singleton doc too, so metadata stays consistent (`state=Exited` persists to disk) when the recorded process is gone, matching `--all` semantics.
- **Closes #222** (audit): see the audit-report comment on the issue for the full sweep. Three further findings (RPC-retry budget, controller-grace vs `--timeout` asymmetry, `--force` orphaning child pgroup) are filed as separate `bug` follow-ups for later prioritisation; see #251 and #252.

## Notes for the reviewer

- `isInstanceAlive` lives in `internal/discovery/reconcile.go` rather than pidutil because the contract is reconcile-specific: liveness-first, token-second, with a "transient /proc error keeps prior behavior" branch that pidutil itself deliberately does not impose. The existing `processIsOurs` in `cmd/sb/stop/terminals.go` is the opposite shape (token-first, liveness-second) — leaving each surface its own semantics avoids over-coupling the stop signal path to the reconcile-side defaults. Happy to consolidate if the reviewer prefers a single helper.
- The single-name reconcile uses the existing `discovery.ReconcileTerminals` slice entry point instead of the package-private `reconcileTerminal` to avoid expanding the discovery package's surface for a one-shot caller. Cost is a one-element slice allocation per `sb stop <name>`.
- The new `internal/discovery/reconcile_test.go::TestReconcileTerminals_RejectsRecycledPid` is Linux-gated because the PidStart token is `/proc`-derived (zero on other platforms). The zero-token-fallback case is covered cross-platform by `TestReconcileTerminals_ZeroTokenFallsBackToLiveness`.

## Test plan

- [x] `file ./sbsh` — confirms ELF 64-bit LSB executable.
- [x] `go test ./internal/discovery/... ./cmd/sb/stop/...` — passes (new tests + existing).
- [x] `go test -count=10 -run 'Reconcile|ResolveTargets' ./internal/discovery/... ./cmd/sb/stop/...` — non-flaky over 10 runs.
- [x] `make test` — Makefile test target passes (cmd/sb..., cmd/sbsh..., internal/terminal..., internal/client..., integration get/, e2e).
- [x] `pre-commit run --files internal/discovery/reconcile.go internal/discovery/reconcile_test.go cmd/sb/stop/terminals.go cmd/sb/stop/terminals_test.go` — all hooks pass.
- [ ] `make test-race` — not run; the make `test` target above does not include `-race`, and a full `-race` sweep across `./pkg/...` is broader than the diff. Reviewer can run separately if desired.

Closes #248
Closes #249
Closes #222

### DIFF
--- a/cmd/sb/stop/terminals.go
+++ b/cmd/sb/stop/terminals.go
@@ -224,7 +224,12 @@ func resolveTargets(
 	if doc == nil {
 		return nil, fmt.Errorf("%w: terminal %q", errdefs.ErrTerminalNotFound, opts.name)
 	}
-	return []api.TerminalDoc{*doc}, nil
+	// Persist Exited if the recorded process is gone (or its PID has been
+	// recycled to an unrelated instance). Mirrors the --all branch above so
+	// metadata stays consistent regardless of which target path was taken.
+	docs := []api.TerminalDoc{*doc}
+	discovery.ReconcileTerminals(ctx, logger, opts.runPath, docs)
+	return docs, nil
 }
 
 func stopOneTerminal(
--- a/cmd/sb/stop/terminals_test.go
+++ b/cmd/sb/stop/terminals_test.go
@@ -18,12 +18,18 @@ package stop
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
+	"io"
+	"log/slog"
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/eminwux/sbsh/cmd/config"
+	"github.com/eminwux/sbsh/internal/defaults"
 	"github.com/eminwux/sbsh/internal/errdefs"
+	"github.com/eminwux/sbsh/pkg/api"
 	"github.com/spf13/viper"
 )
 
@@ -134,5 +140,73 @@ func Test_ProcessIsOurs_ZeroTokenFallsBack(t *testing.T) {
 	}
 	if processIsOurs(0, 0) {
 		t.Fatal("expected processIsOurs(0, 0) to be false")
+	}
+}
+
+func writeStopTestTerminal(t *testing.T, runPath, id, name string, state api.TerminalStatusMode, pid int) {
+	t.Helper()
+	dir := filepath.Join(runPath, defaults.TerminalsRunPath, id)
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		t.Fatalf("mkdir %s: %v", dir, err)
+	}
+	doc := api.TerminalDoc{
+		APIVersion: api.APIVersionV1Beta1,
+		Kind:       api.KindTerminal,
+		Spec:       api.TerminalSpec{ID: api.ID(id), Name: name, RunPath: runPath},
+		Status: api.TerminalStatus{
+			State:           state,
+			Pid:             pid,
+			TerminalRunPath: dir,
+		},
+	}
+	b, err := json.Marshal(doc)
+	if err != nil {
+		t.Fatalf("marshal: %v", err)
+	}
+	if writeErr := os.WriteFile(filepath.Join(dir, "metadata.json"), b, 0o644); writeErr != nil {
+		t.Fatalf("write metadata.json: %v", writeErr)
+	}
+}
+
+func readStopTestTerminalState(t *testing.T, runPath, id string) api.TerminalStatusMode {
+	t.Helper()
+	b, err := os.ReadFile(filepath.Join(runPath, defaults.TerminalsRunPath, id, "metadata.json"))
+	if err != nil {
+		t.Fatalf("read metadata: %v", err)
+	}
+	var doc api.TerminalDoc
+	if unmarshalErr := json.Unmarshal(b, &doc); unmarshalErr != nil {
+		t.Fatalf("unmarshal: %v", unmarshalErr)
+	}
+	return doc.Status.State
+}
+
+// Test_ResolveTargets_SingleName_PersistsExited covers the previously-missing
+// reconcile step on the single-target name path: a stale terminal whose
+// recorded PID is dead should be persisted as Exited after resolveTargets
+// runs, matching what the --all branch already does.
+func Test_ResolveTargets_SingleName_PersistsExited(t *testing.T) {
+	const deadPid = 0x7fffffff // very-high PID unlikely to exist
+	runPath := t.TempDir()
+	writeStopTestTerminal(t, runPath, "id-dead", "alpha", api.Ready, deadPid)
+
+	opts := stopOpts{
+		runPath: runPath,
+		name:    "alpha",
+		timeout: defaultStopTimeout,
+	}
+	logger := slog.New(slog.NewTextHandler(io.Discard, nil))
+	docs, err := resolveTargets(context.Background(), logger, opts)
+	if err != nil {
+		t.Fatalf("resolveTargets: %v", err)
+	}
+	if len(docs) != 1 {
+		t.Fatalf("expected 1 doc, got %d", len(docs))
+	}
+	if docs[0].Status.State != api.Exited {
+		t.Fatalf("in-memory: expected Exited after reconcile, got %v", docs[0].Status.State)
+	}
+	if got := readStopTestTerminalState(t, runPath, "id-dead"); got != api.Exited {
+		t.Fatalf("persisted: expected Exited after reconcile, got %v", got)
 	}
 }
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #254 Review — `fix(stop): reject recycled PIDs in reconcile and reconcile single-name targets`**

Looks good. `isInstanceAlive` cleanly extends the liveness predicate with token-based PID-reuse rejection, correctly handles the `os.ErrNotExist` race against `/proc`, and preserves the zero-token-fallback contract `pidutil.Match` already documents. The single-name `resolveTargets` reconcile mirrors the `--all` intent for on-disk metadata consistency. Tests cover dead-PID, zero-token-fallback, and Linux-gated recycled-PID branches plus the new single-name persist case. All commits signed, all checks green. No secrets/PII or dead-symbol issues.

A few small items (all non-blocking nits):

- **Comment drift** `cmd/sb/stop/terminals.go:228`: the new comment says *"Mirrors the --all branch above"*, but the single-name path returns the Exited doc (so `stopOneTerminal` prints `already exited`) while `--all` filters Exited out before returning. The asymmetry is intentional and arguably better UX; the comment could say *"Persists Exited; the active-filter from --all is unnecessary here since stopOneTerminal handles the Exited case explicitly."*

LGTM to merge.

PR #256, opened 2026-05-18
fix(sb stop): --force reaps child shell pgroup before SIGKILL'ing controller

### LINKED ISSUE #252: bug(sb stop): --force SIGKILLs only the controller; child shell pgroup can be orphaned

eminwux:
## Summary

- Closes #252. `sb stop --force` (and the `--kill-after` SIGKILL escalation) now `syscall.Kill(-ChildPgid, SIGKILL)` the recorded child pgroup **before** SIGKILL'ing the controller, so a shell that traps SIGHUP (or ignores it via `nohup`) is taken down with the controller rather than reparented to init. Before this fix, only the controller PID was signaled — SIGKILL is uncatchable so its deferred `shutdownChild` never ran, and the child shell's separate session meant the SIGKILL didn't propagate.
- Adds `ChildPgid int` to `api.TerminalStatus` (omitempty for back-compat with on-disk metadata written by older controllers). The runner captures `cmd.Process.Pid` after PTY start — with `Setsid: true`, `pgid == pid`.
- New helper `sendPgroupKill(pgid int)` in `cmd/sb/stop/terminals.go`: refuses `pgid <= 0` (since `Kill(0, SIGKILL)` would signal the caller's own pgroup — a destructive surprise for terminals whose metadata predates this field), and suppresses ESRCH (already-drained pgroup) so a child that died between metadata-write and stop doesn't surface as an error.

## Notes for the reviewer

- **No PidStart-style recycle protection on `ChildPgid`.** The caller has already filtered with `processIsOurs(doc.Status.Pid, doc.Status.PidStart)` on the controller; per `watchChildExit`, the controller exits when the child does, so a live controller implies the child PID we recorded is still our child. Adding a per-pgid start-time token would have been new scope the issue didn't ask for. Happy to add one if the reviewer disagrees.
- **Back-compat with older metadata.** Terminals started by a controller built before this change have `ChildPgid == 0` on disk. `sendPgroupKill(0)` is a deliberate no-op (skipping the pgroup-kill silently rather than risking `Kill(0, ...)`). Those terminals lose the new protection until restarted — surfacing this only in `--help` and the field comment felt cleaner than synthesizing a fallback that could be wrong.
- **Ordering.** Pgroup-kill fires first, controller-kill second. The issue specifies this order; the rationale is to hit the pgroup while still well-formed (controller alive, PTY chain intact) and avoid any state where the controller is dead but the child group is in transition.
- **Help text.** Updated the `Long` description and the `--force` short help to mention pgroup reaping so operators reading `sb stop --help` see the new behavior.
- **Regression-check.** I temporarily commented out the `sendPgroupKill` call and re-ran `Test_StopOneTerminal_Force_ReapsChildPgroupOnSIGHUPTrap`; it failed with \`child shell pgroup still alive after sigkillGrace=2s; --force only reaped controller\` (the load-bearing assertion) — confirming the test catches the bug it's meant to catch. Then I restored the fix and re-ran clean.

## Test plan

- [x] \`make sbsh-sb\` — canonical build succeeds; \`file ./sbsh\` confirms ELF 64-bit LSB executable and \`./sb\` hardlink.
- [x] \`go build ./...\`, \`go vet ./...\` — clean.
- [x] \`go test -run 'Test_StopOneTerminal_Force|Test_SendPgroupSignal' -count=10 ./cmd/sb/stop/...\` — new tests non-flaky over 10 runs.
- [x] \`make test\` — full Makefile test target passes (\`./cmd/sb...\`, \`./cmd/sbsh...\`, \`./internal/terminal...\`, \`./internal/client...\`, integration \`./cmd/sb/get/...\`, \`./e2e\`).
- [x] \`pre-commit run --files cmd/sb/stop/terminals.go cmd/sb/stop/terminals_test.go internal/terminal/terminalrunner/terminal.go pkg/api/terminal.go\` — all hooks pass.
- [x] Mutation check: temporarily disabled the new \`sendPgroupKill\` call in the \`--force\` branch and confirmed the new test fails with the load-bearing assertion (\`child shell pgroup still alive after sigkillGrace=2s; --force only reaped controller\`); restored the fix and re-verified.
- [ ] \`make test-race\` — not run; the Makefile \`test\` target above doesn't include \`-race\`, and a tree-wide \`-race\` sweep is broader than this diff. Reviewer can run separately if desired.

Closes #252

### DIFF
--- a/cmd/sb/stop/terminals.go
+++ b/cmd/sb/stop/terminals.go
@@ -70,6 +70,10 @@ is unreachable, falls back to SIGTERM on the recorded PID. After signalling,
 waits up to --timeout for the process to exit. With --force, sends SIGKILL
 immediately. With --kill-after, escalates to SIGKILL when the timeout elapses.
 
+Both --force and the --kill-after escalation reap the child shell's entire
+process group (SIGKILL on the recorded child pgid) before SIGKILL'ing the
+controller, so a shell that traps or ignores SIGHUP is taken down too.
+
 Metadata is left in place so 'sb prune terminals' can clean it up.`,
 		SilenceUsage:      true,
 		Args:              cobra.MaximumNArgs(1),
@@ -101,7 +105,7 @@ func setupStopTerminalsCmd(c *cobra.Command) {
 	c.Flags().Bool("all", false, "Stop every active terminal")
 	_ = viper.BindPFlag(config.SB_STOP_ALL.ViperKey, c.Flags().Lookup("all"))
 
-	c.Flags().Bool("force", false, "Send SIGKILL immediately instead of a graceful stop")
+	c.Flags().Bool("force", false, "Send SIGKILL immediately (reaps the entire child pgroup, not just the controller)")
 	_ = viper.BindPFlag(config.SB_STOP_FORCE.ViperKey, c.Flags().Lookup("force"))
 
 	c.Flags().Duration("timeout", defaultStopTimeout, "How long to wait for the terminal to exit")
@@ -260,6 +264,14 @@ func stopOneTerminal(
 	}
 
 	if opts.force {
+		// Reap the child pgroup first: SIGKILL the controller alone
+		// leaves any pgroup member that traps SIGHUP (or ignores it via
+		// nohup) reparented to init. Pre-#252 metadata has no ChildPgid
+		// recorded; sendPgroupSignal treats zero as a no-op.
+		if err := sendPgroupKill(doc.Status.ChildPgid); err != nil {
+			fmt.Fprintf(stderr, "failed to SIGKILL terminal %q child pgroup: %v\n", name, err)
+			return resultFailed
+		}
 		if err := sendSignal(doc.Status.Pid, doc.Status.PidStart, syscall.SIGKILL); err != nil {
 			fmt.Fprintf(stderr, "failed to SIGKILL terminal %q: %v\n", name, err)
 			return resultFailed
@@ -295,6 +307,12 @@ func stopOneTerminal(
 	}
 
 	fmt.Fprintf(stdout, "terminal %q did not exit within %s; escalating to SIGKILL\n", name, opts.timeout)
+	// Mirror the --force branch: reap the child pgroup before SIGKILL'ing
+	// the controller so SIGHUP-trapping members don't survive.
+	if err := sendPgroupKill(doc.Status.ChildPgid); err != nil {
+		fmt.Fprintf(stderr, "failed to SIGKILL terminal %q child pgroup: %v\n", name, err)
+		return resultFailed
+	}
 	if err := sendSignal(doc.Status.Pid, doc.Status.PidStart, syscall.SIGKILL); err != nil {
 		fmt.Fprintf(stderr, "failed to SIGKILL terminal %q: %v\n", name, err)
 		return resultFailed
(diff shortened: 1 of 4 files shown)


eminwux:
**PR #256 Review — `fix(sb stop): --force reaps child shell pgroup before SIGKILL'ing controller`**

Targeted fix for #252: `sb stop --force` (and the `--kill-after` SIGKILL escalation) now SIGKILL the recorded child pgroup before SIGKILL'ing the controller, so SIGHUP-trapping shells go down with the controller instead of reparenting to init. `sendPgroupKill` correctly refuses `pgid <= 0` (the `Kill(0, SIGKILL)` self-signal trap) and suppresses ESRCH. The `processIsOurs(Pid, PidStart)` filter at `cmd/sb/stop/terminals.go:261` backs the comment's claim about the caller already verifying liveness. Sole commit is GPG-verified; no secret/PII leaks in new error messages.

One nit:

- **Name drift** `cmd/sb/stop/terminals.go:270`, `cmd/sb/stop/terminals_test.go:323,334`: the comment says *"sendPgroupSignal treats zero as a no-op"* and the two test functions are named `Test_SendPgroupSignal_*`, but the helper is `sendPgroupKill`. No `sendPgroupSignal` exists anywhere on the PR head — looks like leftover drift from an earlier rename. Rename for consistency (non-blocking; will file as followup-nit).

LGTM.

PR #262, opened 2026-05-18
fix(discovery): scanMetadataFiles logs and skips per-file failures instead of aborting

### LINKED ISSUE #258: bug(discovery): scanMetadataFiles aborts on first corrupt metadata.json, breaking all discovery
## Context

`pkg/discovery/scan.go:64-80` (`scanMetadataFiles`) iterates every `metadata.json` matched by `glob`. The very first read or unmarshal failure returns `(nil, error)`, aborting the scan.

Callers — `ScanTerminals`, `ScanClients`, and every downstream surface built on them (`sb ls`, `sb attach`, `sb stop`, `sb prune`, the reconcile pass in `ScanAndPrint*`) — propagate that error. **One corrupt or unreadable terminal/client metadata file therefore takes down the entire discovery layer for every terminal under the same `runPath`.**

Tests at `pkg/discovery/clients_test.go:92` (`TestScanClients_InvalidJSON`) and `pkg/discovery/terminals_test.go:98` (`TestScanTerminals_InvalidJSON`) actively lock in this hard-fail behavior — it's intentional. But the design is poorly aligned with the profile loader (`pkg/discovery/profiles.go`), which surfaces per-file failures as `ProfileWarning` and continues scanning. The mismatch is a real availability gap: a single bad terminal directory hides every other live terminal.

## Proposal

Make discovery resilient to per-file failures the way profile loading is. On a read or decode error for a single `metadata.json`, log + skip and continue scanning. Two equivalent shapes:

- Return `([]T, []DiscoveryWarning, nil)` mirroring `LoadProfilesFromDir`, or
- Keep the `([]T, error)` shape, route per-file failures to a logger.Warn, and reserve the error return for the unrecoverable glob-itself failure.

Either way, one corrupt directory must not be allowed to mask every other live terminal from `sb ls`, `sb attach`, `sb stop`.

Update the two `TestScan*_InvalidJSON` tests to assert the new contract (warnings/skips, not error).

## Acceptance criteria

- [ ] `scanMetadataFiles` no longer returns early on a single read/decode failure
- [ ] `sb ls` shows other terminals when one terminal's `metadata.json` is corrupt
- [ ] `sb attach <other-id>` succeeds when an unrelated terminal's `metadata.json` is corrupt
- [ ] `TestScanClients_InvalidJSON` and `TestScanTerminals_InvalidJSON` updated to assert the warnings/skip contract
- [ ] Existing happy-path tests still pass

## Out of scope

- Auto-pruning corrupt metadata directories (separate concern — operator decides whether to keep or remove).
- Profile loader changes — already does this correctly.

## References

- `pkg/discovery/scan.go:64-80`
- `pkg/discovery/clients_test.go:92`
- `pkg/discovery/terminals_test.go:98`
- `pkg/discovery/profiles.go:67-75` (warnings-not-errors contract for the comparable loader)
- Discovery audit: #224

eminwux:
## Summary

- One corrupt or unreadable `metadata.json` no longer aborts the whole discovery scan and masks every other live terminal/client from `sb ls`, `sb attach`, `sb stop`, `sb prune`.
- Per-file read or decode failures are logged at `Warn` (with file + error) and skipped; the scan continues. The unrecoverable `glob` failure and `ctx.Err()` paths still return an error.
- Docstring on `scanMetadataFiles` codifies the new warnings-not-errors contract.

## Implementation choice

The issue body offered two equivalent shapes:

1. Return `([]T, []DiscoveryWarning, nil)` mirroring `LoadProfilesFromDir`.
2. Keep the `([]T, error)` shape, route per-file failures to a `logger.Warn`, reserve `error` for the unrecoverable glob-itself failure.

I went with **shape 2**: it satisfies the AC's core goal verbatim ("one corrupt directory must not mask every other live terminal") with a far smaller blast radius — the existing `([]api.TerminalDoc, error)` / `([]api.ClientDoc, error)` signatures are preserved across every caller (`internal/discovery`, `cmd/sb/get`, `cmd/sb/stop`, `cmd/config/autocomplete.go`, `pkg/discovery` `Find*` helpers) without forcing a multi-package signature change. Both `_InvalidJSON` tests now assert the no-error/skip contract per AC bullet 4. Reviewer can push back if the structured-warning shape was actually preferred — that would be a straightforward follow-up.

## Test plan

- [x] `make sbsh-sb` + `file ./sbsh` (ELF binary verified)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=5m \$(go list ./... | grep -v '/e2e\$')` — all green
- [x] `make e2e` — green
- [x] `docs/examples/library-consumer` build + vet — green
- [x] `pre-commit run --files pkg/discovery/scan.go pkg/discovery/clients_test.go pkg/discovery/terminals_test.go` — all hooks pass
- [x] New test `TestScanTerminals_InvalidJSON_SiblingsStillLoad` — sibling good docs load when a third is corrupt (AC bullet 2)
- [x] New test `TestFindTerminalByID_SkipsCorruptSibling` — `sb attach <other-id>` path returns the live doc even when an unrelated metadata is corrupt (AC bullet 3)
- [x] New test `TestScanClients_InvalidJSON_SiblingsStillLoad` — mirror for clients

## Acceptance criteria

- [x] `scanMetadataFiles` no longer returns early on a single read/decode failure
- [x] `sb ls` shows other terminals when one terminal's `metadata.json` is corrupt (covered by the new sibling tests at the discovery layer; `sb ls` routes through `ScanAndPrintTerminals` → `ScanTerminals` → `scanMetadataFiles`)
- [x] `sb attach <other-id>` succeeds when an unrelated terminal's `metadata.json` is corrupt (covered by `TestFindTerminalByID_SkipsCorruptSibling`)
- [x] `TestScanClients_InvalidJSON` and `TestScanTerminals_InvalidJSON` updated to assert the warnings/skip contract
- [x] Existing happy-path tests still pass

Closes #258

### DIFF
--- a/pkg/discovery/clients_test.go
+++ b/pkg/discovery/clients_test.go
@@ -95,9 +95,39 @@ func TestScanClients_InvalidJSON(t *testing.T) {
 	runPath := t.TempDir()
 	writeInvalidClientJSON(t, runPath, "bad")
 
-	_, err := discovery.ScanClients(ctx, logger, runPath)
-	if err == nil {
-		t.Fatal("expected decode error, got nil")
+	got, err := discovery.ScanClients(ctx, logger, runPath)
+	if err != nil {
+		t.Fatalf("expected no error on invalid JSON (per-file failures are skipped), got: %v", err)
+	}
+	if len(got) != 0 {
+		t.Fatalf("expected 0 clients (only the corrupt one was present), got %d", len(got))
+	}
+}
+
+// TestScanClients_InvalidJSON_SiblingsStillLoad verifies the core contract
+// from #258: one corrupt metadata.json must not mask every other live
+// client from discovery.
+func TestScanClients_InvalidJSON_SiblingsStillLoad(t *testing.T) {
+	ctx := context.Background()
+	logger := testLogger()
+	runPath := t.TempDir()
+
+	writeClientDoc(t, runPath, "c-good-1", "alpha")
+	writeInvalidClientJSON(t, runPath, "c-bad")
+	writeClientDoc(t, runPath, "c-good-2", "beta")
+
+	got, err := discovery.ScanClients(ctx, logger, runPath)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	wantIDs := []string{"c-good-1", "c-good-2"}
+	if len(got) != len(wantIDs) {
+		t.Fatalf("want %d clients, got %d (%+v)", len(wantIDs), len(got), got)
+	}
+	for i, w := range wantIDs {
+		if string(got[i].Spec.ID) != w {
+			t.Fatalf("mismatch at %d: want %q, got %q", i, w, got[i].Spec.ID)
+		}
 	}
 }
 
--- a/pkg/discovery/scan.go
+++ b/pkg/discovery/scan.go
@@ -36,6 +36,11 @@ import (
 
 // scanMetadataFiles reads all metadata.json files matching the pattern and unmarshals them.
 // It returns a slice of unmarshaled metadata structs of type T.
+//
+// Per-file read or decode failures are logged at Warn and skipped — one
+// corrupt or unreadable metadata.json must not mask every other live
+// terminal/client from discovery. An error is returned only for
+// unrecoverable conditions (glob itself failing, ctx cancellation).
 func scanMetadataFiles[T any](
 	ctx context.Context,
 	logger *slog.Logger,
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #262 Review — `fix(discovery): scanMetadataFiles logs and skips per-file failures instead of aborting`**

LGTM — narrow, well-tested behavior fix; the new warn-and-skip path preserves the unrecoverable-error contract (glob/ctx) and the exported `ScanTerminals`/`ScanClients` signatures, and the sibling/`FindTerminalByID` tests directly cover the AC.

PR #263, opened 2026-05-18
fix(discovery): reconcile before pruning so stale terminals/clients get cleaned

### LINKED ISSUE #259: bug(discovery): ScanAndPrune{Clients,Terminals} doesn't reconcile, leaving stale entries unpruned
## Context

`internal/discovery/clients.go:78` (`ScanAndPruneClients`) and `internal/discovery/terminals.go:84` (`ScanAndPruneTerminals`) load metadata, then call `scanAndPruneMetadata` with the predicate `Status.State == Exited` (or `ClientExited`) **evaluated against the on-disk state, with no prior call to `ReconcileTerminals` / `ReconcileClients`.**

A terminal whose owning process is dead but whose metadata still claims `Ready` — typical when the user never ran `sb ls` between the crash and the prune, or the process exited without a clean state-flip — is invisible to prune. The user has to run `sb ls` first (which reconciles as a side effect) and *then* `sb prune` for the cleanup to actually catch stale terminals. That ordering is non-obvious and undocumented.

Compare to `ScanAndPrintClients` / `ScanAndPrintTerminals`, which **do** call `ReconcileClients` / `ReconcileTerminals` before printing — so the on-disk view stays in sync as a side effect of `sb ls`. The prune path is the only discovery surface that doesn't reconcile.

## Proposal

Call `ReconcileClients` / `ReconcileTerminals` at the start of `ScanAndPrune*`, before invoking `scanAndPruneMetadata`. One- or two-line change per function.

That way `sb prune` cleans up stale terminals without depending on a prior `sb ls` invocation, and the predicate stays unchanged.

## Acceptance criteria

- [ ] `ScanAndPruneClients` calls `ReconcileClients(ctx, logger, runPath, clients)` before `scanAndPruneMetadata`
- [ ] `ScanAndPruneTerminals` calls `ReconcileTerminals(ctx, logger, runPath, terminals)` before `scanAndPruneMetadata`
- [ ] New unit test: a terminal with a dead PID + `state=Ready` on disk is pruned by `ScanAndPruneTerminals`
- [ ] Same for clients

## Out of scope

- Auto-pruning on every `sb ls` (separate UX call).
- Changes to the `scanAndPruneMetadata` predicate shape.

## References

- `internal/discovery/clients.go:78-97`
- `internal/discovery/terminals.go:84-103`
- `internal/discovery/reconcile.go:89` (`ReconcileTerminals`)
- `internal/discovery/reconcile.go:135` (`ReconcileClients`)
- `internal/discovery/terminals.go:54` (positive example — `ScanAndPrintTerminals` does reconcile)
- Discovery audit: #224

eminwux:
## Summary
- `ScanAndPruneClients` / `ScanAndPruneTerminals` now call `ReconcileClients` / `ReconcileTerminals` before invoking `scanAndPruneMetadata`, mirroring the `ScanAndPrint*` siblings. `sb prune` now cleans stale terminals/clients without requiring a prior `sb ls` to flip metadata via reconcile side effects.
- The prune predicate (`Status.State == Exited` / `== ClientExited`) is unchanged — reconciliation just gives it accurate input.

## Test plan
- [x] `go test ./internal/discovery/... -run 'TestScanAndPrune' -v` — three new tests pass: stale-terminal pruned, stale-client pruned, live-terminal preserved
- [x] CI's main target: `go test -timeout=5m $(go list ./... | grep -v '/e2e$')`
- [x] `make e2e`
- [x] `make sbsh-sb` produces an ELF binary; `go build ./...`; `go vet ./...`

Closes #259

### DIFF
--- a/internal/discovery/clients.go
+++ b/internal/discovery/clients.go
@@ -81,6 +81,7 @@ func ScanAndPruneClients(ctx context.Context, logger *slog.Logger, runPath strin
 		logger.ErrorContext(ctx, "ScanAndPruneClients: failed to scan clients", "error", err)
 		return err
 	}
+	ReconcileClients(ctx, logger, runPath, clients)
 	_, err = scanAndPruneMetadata(
 		ctx,
 		logger,
--- a/internal/discovery/prune_test.go
+++ b/internal/discovery/prune_test.go
@@ -1,0 +1,118 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package discovery_test
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/eminwux/sbsh/internal/defaults"
+	"github.com/eminwux/sbsh/internal/discovery"
+	"github.com/eminwux/sbsh/pkg/api"
+)
+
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #263 Review — `fix(discovery): reconcile before pruning so stale terminals/clients get cleaned`**

LGTM — two-line fix mirrors the existing `ScanAndPrint*` siblings' reconcile-before-render pattern, three new tests cover the stale-prune golden path and a live-terminal regression, no scope creep.

Issue #224, opened 2026-05-18
audit(discovery): socket reuse, pidfile staleness, cross-instance collision sweep


eminwux:
## Context
`internal/discovery/` and `pkg/discovery/` together form the rendezvous
layer — how `sb` finds running terminals, how profiles are located, how
clients pick up sockets. Stale sockets, pid reuse, and cross-instance
collisions are the typical failure modes for this kind of layer. Recent
work has touched socket mode/gid (#189, #201, #208) but not the discovery
walk logic itself.

## Proposal
Read both `internal/discovery/` and `pkg/discovery/` with focus on:
- stale-socket detection: does discovery skip sockets whose owning process
  is gone? what about half-dead processes?
- pidfile handling: stale-pid detection, atomic create/replace
- cross-instance collisions: two `sbsh` instances at once, two operators on
  the same machine
- the reconcile path (`reconcile.go`): idempotency, partial-failure recovery
- profile-discovery edge cases: missing dirs, unreadable files, symlinks

For each defect: file a separate `bug` issue, link here. Small fixes in
one PR; larger ones through per-bug follow-ups.

If clean, close with an audit-report comment.

## Acceptance criteria
- [ ] All `.go` files under both discovery packages read with the bug lens
- [ ] Stale-socket and stale-pidfile detection verified
- [ ] Cross-instance behavior verified or filed as bug
- [ ] Reconcile-path idempotency verified
- [ ] Each finding filed as a separate `bug` issue linked here
- [ ] This issue closed with either a fix PR or an audit-report comment

## Notes
Recent neighbors touching socket attrs: #189, #201, #208.

eminwux:
## Audit report — discovery layer (#224)

Scanned every `.go` file under `internal/discovery/` and `pkg/discovery/`, with the bug lens called out in the issue body. Below are the findings, scored by severity, plus filed bug links.

### Premise note: pidfile handling

[...]

PR #264, opened 2026-05-18
fix(terminalrunner): probe socket for live peer before unlinking in OpenSocketCtrl

### LINKED ISSUE #261: bug(terminal): OpenSocketCtrl removes socket file with no liveness probe, can destroy a live peer on cross-instance collision
## Context

`internal/terminal/terminalrunner/sockets.go:137-145` (`OpenSocketCtrl`) calls `os.Remove(socketFile)` before `lc.Listen(sr.ctx, "unix", socketFile)` to clear a stale socket inode from a previous run:

```go
// Remove sockets if they already exist
if err := os.Remove(socketFile); err != nil {
    sr.logger.Warn(
        "OpenSocketCtrl: couldn't remove stale CTRL socket",
        "socket", socketFile,
        "error", err,
    )
}
```

The `Remove` runs unconditionally, with **no probe of whether a peer is currently listening on the path.** If a live `sbsh` process is already serving the same socket path — operator passes the same `--id` or `--socket-file` twice, two processes race on first start, an old process is still draining a slow client — this silently unlinks the live peer's socket path out from under it.

The live listener still `accept`s on the inode it already holds, but new clients cannot `Dial` it (they get `ENOENT`), and any later operation that re-resolves the path is misled. The failure is silent (only a `Warn` log on the removing side; the victim has no signal at all) and looks to the operator like "attach mysteriously stopped working on the older terminal."

The audit-issue (#224) explicitly calls out this failure mode under "cross-instance collisions: two `sbsh` instances at once."

## Proposal

Probe the socket for a live peer before removing it. Recommended shape:

```go
if conn, err := net.DialTimeout("unix", socketFile, 100*time.Millisecond); err == nil {
    _ = conn.Close()
    return fmt.Errorf("socket already in use by live peer: %s", socketFile)
}
// Dial failed (ENOENT, ECONNREFUSED, timeout) -> no peer; safe to unlink.
_ = os.Remove(socketFile)
```

A successful dial means a live peer is serving; refuse the bind (return an error so the launching `sbsh` exits with a clear diagnostic, rather than stomping the peer). A dial failure means no peer; the path is safe to unlink.

## Acceptance criteria

- [ ] `OpenSocketCtrl` refuses to bind when a live peer is already listening on the socket path
- [ ] A stale socket inode (`ENOENT` or `ECONNREFUSED` dial) still gets cleaned up and the bind proceeds
- [ ] New unit test: pre-bind a listener on a temp socket, call `OpenSocketCtrl` with the same path, assert error returned and pre-bound listener still accepts
- [ ] New unit test: pre-create a stale socket file (no listener), call `OpenSocketCtrl`, assert success and that the new listener is bound

## Out of scope

- Discovery-side stale-socket detection (the reconcile path already handles dead-PID -> Exited; socket inode is a derived artifact and re-binding it is a producer-side concern).
- Changes to `applySocketPerms` / `applyArtifactPerms` (recent neighbor work — see #189, #201, #208).
- I/O socket-pair handling (this concerns the named control socket only).

## References

- `internal/terminal/terminalrunner/sockets.go:137-145`
- Audit: #224 — "cross-instance collisions" lens
- Neighbor socket-attr PRs: #189, #201, #208

eminwux:
## Summary
- `OpenSocketCtrl` now probes the control-socket path with a 100ms `DialContext` before `os.Remove`. A successful dial (live peer accepting on the inode) refuses the bind with `socket already in use by live peer: <path>`; any dial failure (ENOENT, ECONNREFUSED, timeout) means no peer and the existing unlink+listen path runs as before.
- Closes the silent-destroy failure mode flagged by #224's cross-instance-collisions lens: an operator passing the same `--id` / `--socket-file` twice, two starts racing, or an old process still draining a slow client would unlink the live peer's inode and break new client dials with ENOENT.
- Probe uses `(*net.Dialer).DialContext` with `sr.ctx` (satisfies the repo's `noctx` lint; the runner ctx already gates the bind path) and a 100ms `Timeout`. No behavior change on the clean-start path (`os.Remove` still warns on non-ENOENT errors as before).

## Test plan
- [x] `go test ./internal/terminal/terminalrunner/ -run TestOpenSocketCtrl -v` — both new cases pass
- [x] `make test` — full suite green (cmd, internal/terminal, internal/client, integration, e2e)
- [x] `make sbsh-sb` — canonical build produces ELF binary; `file ./sbsh` confirms `ELF 64-bit LSB executable`
- [x] `go vet ./...` clean
- [x] `pre-commit run --files <touched>` — all hooks pass (golangci-lint, go fmt, addlicense)

### New unit tests
- `TestOpenSocketCtrl_RefusesBindOnLivePeer` — pre-binds a real `net.Listen("unix", ...)` peer on a temp socket, calls `OpenSocketCtrl` with the same path, asserts error returned **and** that the pre-bound peer still accepts a fresh dial (proves the live inode was not unlinked).
- `TestOpenSocketCtrl_CleansUpStaleSocket` — creates a stale socket inode via `SetUnlinkOnClose(false)` + `Close`, asserts `OpenSocketCtrl` succeeds, `lnCtrl` is bound, and a dial against the path is now served by the new listener.

Closes #261

### DIFF
--- a/internal/terminal/terminalrunner/sockets.go
+++ b/internal/terminal/terminalrunner/sockets.go
@@ -20,10 +20,18 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"time"
 
 	"github.com/eminwux/sbsh/pkg/api"
 	"golang.org/x/sys/unix"
 )
+
+// liveSocketProbeTimeout bounds the DialTimeout used by OpenSocketCtrl to
+// decide whether a peer is already serving the control socket path before
+// unlinking it. A successful dial inside this window means the inode is
+// live and the bind must refuse; any dial failure (ENOENT, ECONNREFUSED,
+// timeout) means no peer is accepting and the path is safe to unlink.
+const liveSocketProbeTimeout = 100 * time.Millisecond
 
 // defaultSocketMode is the legacy owner-only mode applied when the spec
 // leaves SocketMode at its zero value. Group/world access is opt-in via
@@ -131,6 +139,23 @@ func (sr *Exec) OpenSocketCtrl() error {
 	if errMetadata != nil {
 		sr.logger.Error("OpenSocketCtrl: failed to update metadata", "error", errMetadata)
 		return fmt.Errorf("update metadata: %w", errMetadata)
+	}
+
+	// Probe before unlinking: a successful dial means a live peer is
+	// serving this path (operator passed the same --id / --socket-file
+	// twice, two starts raced, an old process is still draining a slow
+	// client). Refusing the bind keeps the victim's socket intact and
+	// surfaces a clear diagnostic instead of silently breaking attach
+	// against the older terminal. A dial failure (ENOENT, ECONNREFUSED,
+	// timeout) means no peer is accepting; the path is safe to unlink.
+	probeDialer := net.Dialer{Timeout: liveSocketProbeTimeout}
+	if conn, errDial := probeDialer.DialContext(sr.ctx, "unix", socketFile); errDial == nil {
+		_ = conn.Close()
+		sr.logger.Error(
+			"OpenSocketCtrl: socket already in use by live peer",
+			"socket", socketFile,
+		)
+		return fmt.Errorf("socket already in use by live peer: %s", socketFile)
 	}
 
 	// Remove sockets if they already exist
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #264 Review — `fix(terminalrunner): probe socket for live peer before unlinking in OpenSocketCtrl`**

LGTM — addresses #261's two ACs with a `DialContext` probe before `os.Remove`, plus tests that bind a real peer and verify the inode survives the refusal.

PR #265, opened 2026-05-18
fix(discovery): convert profile-walk per-entry errors into ProfileWarnings

### LINKED ISSUE #260: bug(discovery): collectYAMLFiles aborts on first per-entry walk error, contradicts loader's warnings-not-errors contract
## Context

`pkg/discovery/profiles.go:142-160` (`collectYAMLFiles`) uses `filepath.WalkDir` with a callback that returns the per-entry error `e` verbatim:

```go
err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, e error) error {
    if e != nil {
        return e
    }
    ...
})
```

`WalkDir` propagates that error and aborts the walk; `LoadProfilesFromDir` then returns `(nil, nil, error)`, breaking profile loading entirely.

This contradicts the loader's documented contract (`profiles.go:67-75`):

> Any problem that is scoped to a single file or document becomes a `ProfileWarning` rather than aborting the scan: cannot open a file, malformed YAML, missing required apiVersion/kind/metadata.name, unsupported apiVersion or kind, duplicate profile name.

An unreadable subdirectory under `~/.sbsh/profiles/` (0o000 perms, broken symlink resolution under a subtree, etc.) takes down profile loading even though every "real" failure mode in the loader is warnings-only. Failure looks like `sb profile ls` returning an error and showing zero profiles, when the operator's expectation is "show me everything I can read, warn about the rest."

## Proposal

Treat per-entry walk errors as `ProfileWarning` and continue. Two equivalent shapes:

- Inside the `WalkDir` callback, append a warning and return `nil` (skip that branch when the entry is unreadable), or
- Return `filepath.SkipDir` for directory-level errors so the rest of the tree is still walked.

Surface the collected warnings out of `collectYAMLFiles` so they reach the user via `PrintProfileWarnings` like every other loader-level diagnostic. The signature would change to return `([]string, []ProfileWarning, error)` with `error` reserved for the unrecoverable initial-stat case.

## Acceptance criteria

- [ ] `collectYAMLFiles` no longer aborts on per-entry walk failures
- [ ] An unreadable subdir under the profiles dir produces a `ProfileWarning` and the rest of the tree still loads
- [ ] New unit test: subdir with `0o000` perms under the profiles dir yields a warning, not an error; sibling YAML files still load
- [ ] `LoadProfilesFromDir`'s docstring updated to include per-entry walk errors in the warnings-not-errors list

## Out of scope

- Symlink-loop detection (separate concern; `WalkDir` doesn't follow symlinks-to-dirs).
- Following symlinks to directories.

## References

- `pkg/discovery/profiles.go:142-160` (`collectYAMLFiles`)
- `pkg/discovery/profiles.go:67-75` (loader contract docstring)
- Discovery audit: #224

eminwux:
## Summary
- `collectYAMLFiles` no longer aborts the walk when a per-entry error surfaces (unreadable subdir, broken filesystem state under a subtree). The callback records a `ProfileWarning` for `path`, returns `filepath.SkipDir` when the entry is a directory (so the walker doesn't re-report the same `ReadDir` failure) and `nil` otherwise (including the initial-stat case where `d == nil`), so siblings still load.
- `collectYAMLFiles`'s signature is now `([]string, []ProfileWarning, error)`. The function is unexported and `LoadProfilesFromDir` is its sole caller; walk warnings are prepended to the per-file warning slice and surface through `PrintProfileWarnings` like every other loader diagnostic.
- `LoadProfilesFromDir`'s docstring lists per-entry walk failures among the warnings-not-errors cases and narrows the "error returned only when …" clause to the initial stat (the only path that still produces a hard error).

## Test plan
- [x] `go test ./pkg/discovery/... -run TestLoadProfilesFromDir -v` — new `TestLoadProfilesFromDir_UnreadableSubdirProducesWarning` passes; pre-existing cases unchanged. Test skips when `os.Geteuid() == 0` because root bypasses dir mode bits.
- [x] `make sbsh-sb` produces an ELF binary (`file ./sbsh` reports `ELF 64-bit LSB executable`).
- [x] `go build ./...`; `go vet ./...`.
- [x] CI's main suite: `go test -timeout=5m $(go list ./... | grep -v '/e2e$')`.
- [x] CI's integration build tag: `go test -tags=integration ./cmd/sb/get/...`.
- [x] `make e2e`.
- [x] `pre-commit run --files pkg/discovery/profiles.go pkg/discovery/profiles_test.go`.

Closes #260

### DIFF
--- a/pkg/discovery/profiles.go
+++ b/pkg/discovery/profiles.go
@@ -66,14 +66,16 @@ func (w ProfileWarning) String() string {
 //
 // Any problem that is scoped to a single file or document becomes a
 // ProfileWarning rather than aborting the scan:
+//   - cannot read a directory entry during the walk (unreadable subdir,
+//     broken filesystem state under a subtree),
 //   - cannot open a file,
 //   - malformed YAML,
 //   - missing required apiVersion/kind/metadata.name,
 //   - unsupported apiVersion or kind,
 //   - duplicate profile name (keeps the first and skips subsequent).
 //
-// An error is returned only for unrecoverable conditions such as a walk
-// failure on the directory itself.
+// An error is returned only for unrecoverable conditions such as the
+// initial stat on dir failing for a reason other than os.ErrNotExist.
 func LoadProfilesFromDir(
 	ctx context.Context,
 	logger *slog.Logger,
@@ -96,14 +98,14 @@ func LoadProfilesFromDir(
 		return nil, nil, fmt.Errorf("profiles path %q is not a directory", dir)
 	}
 
-	files, err := collectYAMLFiles(dir)
+	files, walkWarnings, err := collectYAMLFiles(dir)
 	if err != nil {
-		return nil, nil, err
+		return nil, walkWarnings, err
 	}
 
 	var (
 		profiles []api.TerminalProfileDoc
-		warnings []ProfileWarning
+		warnings = walkWarnings
 		// seen maps profile name -> file that first loaded that name.
 		seen = make(map[string]string)
 	)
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #265 Review — `fix(discovery): convert profile-walk per-entry errors into ProfileWarnings`**

LGTM — converts per-entry walk failures into `ProfileWarning`s so an unreadable subtree no longer aborts the whole scan; `SkipDir` on directory errors avoids re-reporting the same `ReadDir` failure, and sibling YAML still loads. Signature change is internal to the unexported `collectYAMLFiles`.

PR #267, opened 2026-05-18
fix(clientrunner): probe socket for live peer before unlinking in OpenSocketCtrl

### LINKED ISSUE #266: bug(clientrunner): OpenSocketCtrl removes socket file with no liveness probe (same pattern as #261)
## Context

Surfaced while reviewing #264, which fixed the same failure mode in `internal/terminal/terminalrunner/sockets.go` per #261. The client-side runner has the parallel bug:

`internal/client/clientrunner/sockets.go:33-44` (`OpenSocketCtrl`) calls `os.Remove(socketCtrl)` unconditionally when the path exists, with no probe of whether a live peer is currently serving it:

```go
if _, err := os.Stat(socketCtrl); err == nil {
    sr.logger.Warn("OpenSocketCtrl: removing stale socket", "socket", socketCtrl)
    if rmErr := os.Remove(socketCtrl); rmErr != nil {
        ...
    }
}
lnCfg := net.ListenConfig{}
ln, err := lnCfg.Listen(sr.ctx, "unix", socketCtrl)
```

Same destroy-live-peer semantics as #261: two client processes targeting the same socket path (operator collision, race on first start) silently unlink the live peer's inode; existing accept loop survives on the held fd, but new dials hit `ENOENT`.

Issue #261 explicitly scoped to the terminal runner only, so #264 correctly left this alone. Filing here so the client-side gets the same probe-before-unlink treatment.

## Proposal

Apply the same fix shape as #264:

```go
probeDialer := net.Dialer{Timeout: liveSocketProbeTimeout}
if conn, errDial := probeDialer.DialContext(sr.ctx, "unix", socketCtrl); errDial == nil {
    _ = conn.Close()
    return fmt.Errorf("socket already in use by live peer: %s", socketCtrl)
}
```

Probably worth extracting a shared helper at the same time (the constant and the probe block are now duplicated across two packages).

## Acceptance criteria

- [ ] `clientrunner.OpenSocketCtrl` refuses to bind when a live peer is listening
- [ ] Stale socket (ECONNREFUSED on dial) still gets cleaned up
- [ ] Unit tests mirroring `TestOpenSocketCtrl_RefusesBindOnLivePeer` / `TestOpenSocketCtrl_CleansUpStaleSocket` from #264

## References

- PR #264 (terminal-runner fix)
- Issue #261 (original report, terminal-runner scope)
- `internal/client/clientrunner/sockets.go:25-54`

eminwux:
## Summary
- Mirrors PR #264's terminal-runner fix on the client-runner side: `internal/client/clientrunner/sockets.go`'s `OpenSocketCtrl` now probes the control-socket path with a 100ms `DialContext` before `os.Remove`. A successful dial (live peer accepting on the inode) refuses the bind with `socket already in use by live peer: <path>`; any dial failure (ENOENT, ECONNREFUSED, timeout) means no peer and the existing unlink+listen path runs as before.
- Closes the parallel destroy-live-peer failure mode the client-side runner carried after #261 / #264 scoped the original fix to terminalrunner only: two client processes targeting the same socket path (operator collision, race on first start) would silently unlink the live peer's inode, leaving the held-fd accept loop alive but new dials hitting ENOENT.
- Probe uses `(*net.Dialer).DialContext` with `sr.ctx` (satisfies the repo's `noctx` lint; the runner ctx already gates the bind path) and a 100ms `Timeout`. No behavior change on the clean-start path (`os.Remove` still warns on non-ENOENT errors as before).
- **Shared-helper deferral.** The issue's proposal notes a shared helper is "probably worth extracting" since the constant and probe block are now duplicated across two packages. Held off to keep this PR's diff bug-fix-only — extracting now would also touch `internal/terminal/terminalrunner/sockets.go` (#264's code) and its tests. Easy follow-up once a clean home for the helper is decided (likely a new `internal/socketprobe` package or similar). Pushing back here is welcome if you'd rather the dedup land in the same PR.

## Test plan
- [x] `go test ./internal/client/clientrunner/ -run TestOpenSocketCtrl -v` — both new cases pass
- [x] `make test` — full suite green (cmd, internal/terminal, internal/client, integration, e2e)
- [x] `make sbsh-sb` — canonical build produces ELF binary; `file ./sbsh` confirms `ELF 64-bit LSB executable`
- [x] `go vet ./...` clean
- [x] `pre-commit run --files <touched>` — all hooks pass (golangci-lint, go fmt, addlicense)

### New unit tests (mirror #264)
- `TestOpenSocketCtrl_RefusesBindOnLivePeer` — pre-binds a real `net.Listen("unix", ...)` peer on a temp socket, calls `OpenSocketCtrl` with the same path, asserts error returned **and** that the pre-bound peer still accepts a fresh dial (proves the live inode was not unlinked).
- `TestOpenSocketCtrl_CleansUpStaleSocket` — creates a stale socket inode via `SetUnlinkOnClose(false)` + `Close`, asserts `OpenSocketCtrl` succeeds, `lnCtrl` is bound, and a dial against the path is now served by the new listener.

Closes #266

### DIFF
--- a/internal/client/clientrunner/sockets.go
+++ b/internal/client/clientrunner/sockets.go
@@ -20,7 +20,15 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"time"
 )
+
+// liveSocketProbeTimeout bounds the DialTimeout used by OpenSocketCtrl to
+// decide whether a peer is already serving the control socket path before
+// unlinking it. A successful dial inside this window means the inode is
+// live and the bind must refuse; any dial failure (ENOENT, ECONNREFUSED,
+// timeout) means no peer is accepting and the path is safe to unlink.
+const liveSocketProbeTimeout = 100 * time.Millisecond
 
 func (sr *Exec) OpenSocketCtrl() error {
 	sr.metadataMu.RLock()
@@ -28,6 +36,24 @@ func (sr *Exec) OpenSocketCtrl() error {
 	sr.metadataMu.RUnlock()
 
 	sr.logger.Debug("OpenSocketCtrl: preparing to listen", "socket", socketCtrl)
+
+	// Probe before unlinking: a successful dial means a live peer is
+	// serving this path (operator passed the same --id / --socket-file
+	// twice, two starts raced, an old process is still draining a slow
+	// client). Refusing the bind keeps the victim's socket intact and
+	// surfaces a clear diagnostic instead of silently breaking new
+	// dials against the older client. A dial failure (ENOENT,
+	// ECONNREFUSED, timeout) means no peer is accepting; the path is
+	// safe to unlink.
+	probeDialer := net.Dialer{Timeout: liveSocketProbeTimeout}
+	if conn, errDial := probeDialer.DialContext(sr.ctx, "unix", socketCtrl); errDial == nil {
+		_ = conn.Close()
+		sr.logger.Error(
+			"OpenSocketCtrl: socket already in use by live peer",
+			"socket", socketCtrl,
+		)
+		return fmt.Errorf("socket already in use by live peer: %s", socketCtrl)
+	}
 
 	// remove stale socket if it exists
 	if _, err := os.Stat(socketCtrl); err == nil {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #267 Review — `fix(clientrunner): probe socket for live peer before unlinking in OpenSocketCtrl`**

LGTM — verbatim mirror of #264's probe-before-unlink fix on the client-runner side, with both ACs from #266 covered by new unit tests and the shared-helper extraction deliberately deferred (already tracked on #266).

PR #269, opened 2026-05-18
test(terminalrunner): assert subscribeErrCount surfaces empty race window

### LINKED ISSUE #238: terminalrunner: assert on subscribeErrCount or drop it in TestSubscribe_NoOrphanDrainOnConcurrentClose
Followup from #237 review.

`internal/terminal/terminalrunner/subscribe_close_race_test.go:115` declares:

\`\`\`go
var subscribeErrCount atomic.Int32
\`\`\`

incremented at line 135, but never read. The comment at lines 130-134 claims it \"surfaces if every single subscriber was rejected (would indicate the race window never opened)\" — but no assertion checks that.

The risk: if a future change to `Subscribe` made the early ctx-check fire more aggressively (e.g. before the race window had a chance to open), every subscribe call would take the rejection path and the test would silently still pass without exercising the race the test exists to cover.

Fix options:
- Add an assertion after `wg.Wait()`: `if subscribeErrCount.Load() == concurrentSubscribers { t.Fatal(\"race window never opened — every subscriber was early-rejected\") }`.
- Or drop the variable and the comment if the assertion isn't wanted.

Either lines up the code with the comment's intent.

eminwux:
## Summary
- Adds the assertion the existing comment at `subscribe_close_race_test.go:130-134` already advertised: if every concurrent subscriber takes the early-rejection path, the race window the test exists to cover never opened. Pre-change `subscribeErrCount` was incremented but never read, so a future change to `Subscribe` that fired the ctx-check more aggressively would silently still pass.
- Picked option 1 from the issue (add the assertion) over option 2 (drop the var). Option 1 is strictly more test coverage and aligns the code with the existing comment's intent; the diff is a 3-line `if` plus a comment refresh at the increment site.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race -run 'TestSubscribe_NoOrphanDrainOnConcurrentClose|TestSubscribe_PostCloseRejectsCleanly' -count=5 ./internal/terminal/terminalrunner/`
- [x] `make sbsh-sb` (ELF verified for `sbsh` and `sb`)
- [x] `make test` (full CI test target — all packages green incl. e2e)

Closes #238

### DIFF
--- a/internal/terminal/terminalrunner/subscribe_close_race_test.go
+++ b/internal/terminal/terminalrunner/subscribe_close_race_test.go
@@ -128,8 +128,8 @@ func TestSubscribe_NoOrphanDrainOnConcurrentClose(t *testing.T) {
 				// Post-Close rejection is the expected outcome for the
 				// late-arriving half; the cliFD was already closed by
 				// Subscribe's error path. Count the rejection so the
-				// test surfaces if every single subscriber was rejected
-				// (would indicate the race window never opened).
+				// post-wg.Wait() assertion can fail if every single
+				// subscriber was rejected (race window never opened).
 				if !errors.Is(err, errTerminalClosing) {
 					t.Errorf("subscribe (idx=%d): unexpected error: %v", idx, err)
 				}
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #269 Review — `test(terminalrunner): assert subscribeErrCount surfaces empty race window`**

LGTM — implements option 1 from #238 exactly: 5-line post-`wg.Wait()` assertion that fails when every subscriber takes the early-rejection path, plus a comment refresh that points at the new assertion. Closes the dead-read gap without expanding scope.

PR #272, opened 2026-05-19
refactor(terminalstore): trim TerminalStore to Add-only and drop dead Mgr field

eminwux:
## Summary
- `terminalstore.TerminalStore` had six methods (`Add`, `Get`, `ListLive`, `Remove`, `Current`, `SetCurrent`); production code only ever calls `Add` from three sites in `internal/client/controller.go`. The other five were dead surface, with the `current api.ID` tracking field they implied. Pruning them and the orphan `LastGetID` / `LastRemovedID` / `LastSetCurrentID` / `CurrentID` trackers on the test fake keeps the package's contract aligned with its single production caller, per #250 / parent audit #223.
- `clientrunner.Exec.Mgr *terminalstore.Exec` was an exported field that no caller ever read or wrote; the audit confirmed it was a leaky concrete-type placeholder. Removing it also drops the now-unused `terminalstore` import from `client_runner_exec.go`.
- `session_store.go`'s mutex is now a plain `sync.Mutex` (no readers remain). `doc.go`'s package preamble is rewritten to describe the Add-only fail-fast guard the package actually is — including a corrected "Concurrency" section that no longer references read methods or `SetCurrent`'s check-then-assign.

## Test plan
- [x] `make sbsh-sb` — canonical build produces an ELF `sbsh` (verified via `file`)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full unit suite + integration + e2e (`./e2e` ran green, including the controller tests whose ~90 stub blocks were rewritten)
- [x] pre-commit hooks (gofmt, golangci-lint, addlicense) green

Closes #250

Issue #250, opened 2026-05-18
chore(terminalstore): trim unused TerminalStore surface area to match production callers


eminwux:
## Context

**Phase 1 of 2.** Phase 2 (delete the package outright once this trim lands) is tracked in #275.

Audit of `internal/client/terminalstore/session_store.go` (issue #223)
verified the concurrency story is sound but surfaced dead public
surface area worth trimming. Today the `TerminalStore` interface
advertises six methods (`Add`, `Get`, `ListLive`, `Remove`, `Current`,
`SetCurrent`) plus the related `current` field tracking in `Exec`,
but production code only ever exercises `Add` — the only non-test
callers are three `s.ss.Add(terminal)` sites in
`internal/client/controller.go` (lines 334, 368, 392). `Get`,
`ListLive`, `Remove`, `Current`, and `SetCurrent` are dead in
production; the test stubs in `controller_test.go` provide them
solely to match the interface shape, never to assert observed
behavior.

The same package is also reached indirectly through
`internal/client/clientrunner/client_runner_exec.go:54`:

```go
Mgr      *terminalstore.Exec
```

This exported field is never read or written anywhere in the repo
(`grep -rn '\.Mgr\b' --include='*.go'` returns nothing). It pins a
leaky concrete-type dependency on `*terminalstore.Exec` and pulls
in the `terminalstore` import on top of the dead-field carry.

The cumulative effect on a future reader: the file's name
(`session_store.go`), the interface design (registry with `Current`
/ `ListLive` / `Remove`), and the `Exec.Mgr` placeholder all
suggest a stateful client-side session manager with active lifecycle
tracking. The reality (verified during the #223 audit) is a single
in-memory `Add`-only fail-fast guard for ID collisions, with the
real on-disk inventory living under `<runPath>/terminals/*` and
resolved via `internal/discovery`. Code that doesn't run cannot
race, but it can mislead the next reader into reasoning about
properties that aren't load-bearing.

## Proposal

Shrink the package to match its single production caller:

1. **`internal/client/terminalstore/session_store.go`**
   - Reduce the `TerminalStore` interface to `Add(s *api.AttachedTerminal) error`.
   - Drop `Get`, `ListLive`, `Remove`, `Current`, `SetCurrent` methods on `Exec`.
   - Drop the `current api.ID` field and the now-orphaned `m.current = s.Spec.ID` / `m.current = ""` branches in `Add` / `Remove`.

[...]

Issue #223, opened 2026-05-18
audit(terminalstore): concurrency and stale-entry sweep


eminwux:
## Context
`internal/client/terminalstore/session_store.go` is the client-side store
of known terminals. Stores of this shape — registry of active processes —
are a classic source of TOCTOU races, stale entries after crashes, and
inconsistent state under concurrent modify. No recent commits have touched
this package, which usually means either it's solid or it's been quietly
accumulating risk.

## Proposal
Read `session_store.go` (and `session_store_fake.go` for the contract it
implements) with focus on:
- concurrent add / remove / lookup: locking discipline, atomicity
- crash-recovery: what happens to entries whose process died without a
  graceful stop?
- TOCTOU between "lookup by name" and "send signal" — does this exist?
- name collisions and reuse across stop → start cycles (cross-ref #182)
- file-format / serialization edge cases if persisted

For each defect: file a separate `bug` issue, link here. Small fixes in
one PR; larger ones through per-bug follow-ups.

If clean, close with an audit-report comment.

## Acceptance criteria
- [ ] `session_store.go` read end-to-end with the bug lens
- [ ] Concurrent-access invariants verified
- [ ] Crash-recovery / stale-entry handling verified
- [ ] Each finding filed as a separate `bug` issue linked here
- [ ] This issue closed with either a fix PR or an audit-report comment

## Notes
Related: #182 (fail-fast on name collision), #109 (pid-reuse safety).

PR #284, opened 2026-05-19
chore(terminalstore): remove client-side store package (phase 2)

### LINKED ISSUE #275: chore(terminalstore): remove client-side store package (phase 2)
## Context

**Phase 2 of 2.** Phase 1 is #250 (PR #272). Do not start this work until #272 merges.

After #250's trim lands, the `internal/client/terminalstore` package's only production method is `Add`, called exactly once per `Controller` lifetime from one of three sites in `internal/client/controller.go`:

- `controller.go:334` — `createAttachTerminal`, `SocketFile` branch
- `controller.go:368` — `createAttachTerminal`, discovery branch
- `controller.go:392` — `createRunNewTerminal`

For any given `sb` / `sbsh` invocation, exactly one of `createRunNewTerminal` (line 188) or `createAttachTerminal` (line 233) runs. Each calls `Add` once and returns the `*api.AttachedTerminal` to the caller, which uses it directly (e.g. `s.sr.StartTerminalCmd(terminal)` at line 204). Nothing reads the store after `Add` — the underlying `map[api.ID]*api.AttachedTerminal` holds at most one entry, ever, and is write-only.

The remaining surface (map + `sync.Mutex` + `TerminalStore` interface + `Exec` struct + `errdefs.ErrTerminalStore` wrap + fake) all exists to defend a one-slot, write-only registry behind a mutex behind an interface. The "fail-fast ID-collision guard" guards a double-Add path no production caller exercises.

## Proposal

Delete the `internal/client/terminalstore` package entirely:

1. Move `NewSupervisedTerminal(spec *api.TerminalSpec) *api.AttachedTerminal` to a fitting destination — `pkg/api` next to `AttachedTerminal`, or kept inline in `internal/client/controller.go` if not reused elsewhere. Dev's call.
2. Replace each of the three `s.ss.Add(terminal)` blocks in `controller.go` with direct construction; drop the wrap layer:
   ```go
   terminal := api.NewSupervisedTerminal(&spec) // or controller-local helper
   return terminal, nil
   ```
3. Remove the `s.ss` field, the `NewTerminalStore` factory field on `Controller` (line 46), and the `s.ss = s.NewTerminalStore()` line (line 130) in the constructor.
4. Drop `errdefs.ErrTerminalStore` (`internal/errdefs/errdefs.go:36`) — its only callers are the three wrap sites being removed plus the test asserts below.
5. Delete `internal/client/terminalstore/` (`session_store.go`, `session_store_fake.go`, `doc.go`).
6. Drop the now-unused `terminalstore` import from `controller.go`.
7. Remove the 18 `sc.NewTerminalStore = func() terminalstore.TerminalStore { ... }` overrides in `controller_test.go`, the `terminalstore` import, and the `Test_ErrTerminalStore` test (`controller_test.go:735`) plus the two remaining `errdefs.ErrTerminalStore` asserts at `controller_test.go:821-822` and `:1024-1026` — they assert behavior that no longer exists.

## Acceptance criteria

- [ ] `internal/client/terminalstore/` directory is deleted; `grep -rn 'terminalstore' --include='*.go' .` returns nothing.
- [ ] Three `s.ss.Add(...)` sites in `internal/client/controller.go` are replaced by direct construction; downstream callers receive the `*api.AttachedTerminal` unchanged.
- [ ] `Controller.ss` field, `Controller.NewTerminalStore` factory field, and `s.ss = s.NewTerminalStore()` line removed.
- [ ] `errdefs.ErrTerminalStore` removed (verified no callers remain after the controller cleanup).
- [ ] `Test_ErrTerminalStore` deleted and the two other `errdefs.ErrTerminalStore` asserts in `controller_test.go` removed; no `terminalstore` import remains in `controller_test.go`.
- [ ] `make sbsh-sb` produces an ELF binary (verified via `file`), `go build ./...`, `go vet ./...`, and `make test` (including `./e2e`) green.

## Out of scope

- The "real supervisor" feature (a single `sbsh terminal` hosting multiple child processes with `sb attach` cycling between their screens) is intra-terminal state at the terminal/runtime layer, not a client-side `AttachedTerminal` registry. If/when it lands, its primitives belong where the children's pty fds live, not in a revived client-side store. Deleting this package does not foreclose that design.
- Any change to `internal/discovery` (on-disk inventory) or `internal/pidutil` (PID-reuse safety).

## Notes

- Phase 2 of 2. Phase 1: #250 (implementation: PR #272).
- #250 explicitly anticipated this in its *Out of scope*: *"Reintroducing per-session lifecycle tracking later (e.g. for a multi-terminal client UI) is fine; this issue is only about pruning dead surface, not foreclosing future designs. If/when the use case materializes, the methods can come back targeted to the actual caller."* Phase 2 deletes the rest because no caller materialized; if the supervisor feature lands later, it will introduce different primitives at a different layer (see *Out of scope*).
- Parent audit: #223.

eminwux:
## Summary

Phase 2 of 2 (phase 1: #250, PR #272). Deletes the `internal/client/terminalstore` package entirely. After #272 trimmed `TerminalStore` to a one-slot, write-only `Add`, nothing reads the store after the single insert — the map + mutex + interface + `ErrTerminalStore` wrap layer all defended a registry no production caller exercises.

- Deleted `internal/client/terminalstore/` (`session_store.go`, `session_store_fake.go`, `doc.go`).
- Moved `NewSupervisedTerminal` into `internal/client/controller.go` as the unexported `newSupervisedTerminal` (it had a single internal caller — controller.go's three create paths — so a package-local helper keeps the blast radius minimal; the issue left the destination as dev's call).
- Replaced the three `s.ss.Add(terminal)` blocks with direct construction; dropped the wrap layer.
- Removed `Controller.ss`, `Controller.NewTerminalStore`, the `s.ss = s.NewTerminalStore()` line, and the `NewTerminalStore:` constructor field.
- Dropped `errdefs.ErrTerminalStore` and the `!errors.Is(errCreate, errdefs.ErrTerminalStore)` guard in `Run`.
- Tests: removed the 17 `sc.NewTerminalStore = …` overrides + 2 `sc.ss = …` assignments, the `terminalstore` import, and `Test_ErrTerminalStore`. The two `Test_CreateAttachTerminal_SocketFileOnly_*` tests are reworked to call `createAttachTerminal` directly (they verify synthetic-ID and no-caller-mutation behavior, which survives the store removal).

## Divergences from the literal AC (documented per dev playbook)

The issue body predated #272 and its Proposal/AC enumeration is slightly stale. Two items extend beyond the literal AC, both directly caused by the deletion and aligned with the dead-surface-pruning goal (parent audit #223):

1. **`Mgr *terminalstore.Exec`** — the AC's grep-returns-nothing criterion would have failed on this field, but #272 already removed it from `client_runner_exec.go`. No action needed here; noting it so the grep result is auditable.
2. **`Test_ErrTerminalExists` + `errdefs.ErrTerminalExists`** — the AC's test enumeration named only `Test_ErrTerminalStore` and the `ErrTerminalStore` asserts, but `Test_ErrTerminalExists` (and its `ErrTerminalExists` asserts) tests the store's `Add` ID-collision guard, which this PR deletes. The test can no longer compile (it referenced `terminalstore.Test`/`sc.NewTerminalStore`), so it is removed. With the store gone, `errdefs.ErrTerminalExists` ("terminal id already exists in store") had zero remaining callers, so it is dropped alongside `ErrTerminalStore`. If the reviewer prefers to keep the sentinel, it can be restored — but it is dead, store-specific surface.

`NewSupervisedTerminal` was kept controller-local rather than promoted to `pkg/api`: it has one internal caller and the issue's *Out of scope* notes the future supervisor feature will introduce different primitives at a different layer, so there is no reuse to plan for.

## Test plan

- [x] `make sbsh-sb` — ELF binary produced + hardlinked to `sb` (verified via `file`).
- [x] `go vet ./...` — clean.
- [x] `make test` — full CI target green (cmd/sb, cmd/sbsh, internal/terminal, internal/client, integration, e2e).
- [x] `pre-commit run --files …` — clean (golangci-lint, go fmt, go-mod-tidy, addlicense, etc.).
- [x] `grep -rn 'terminalstore' --include='*.go' .` returns nothing.

## Acceptance criteria

- [x] `internal/client/terminalstore/` deleted; `grep -rn 'terminalstore'` returns nothing.
- [x] Three `s.ss.Add(...)` sites replaced by direct construction; downstream callers receive the `*api.AttachedTerminal` unchanged.
- [x] `Controller.ss`, `Controller.NewTerminalStore`, and `s.ss = s.NewTerminalStore()` removed.
- [x] `errdefs.ErrTerminalStore` removed (no callers remain).
- [x] `Test_ErrTerminalStore` deleted, `ErrTerminalStore` asserts removed, no `terminalstore` import in `controller_test.go`. (Also removed `Test_ErrTerminalExists` — see Divergences.)
- [x] `make sbsh-sb` ELF, `go build ./...`, `go vet ./...`, `make test` (incl. `./e2e`) green.

Closes #275


eminwux:
**PR #284 Review — `chore(terminalstore): remove client-side store package (phase 2)`**

LGTM — clean phase-2 deletion of the write-only `terminalstore` registry: all of `terminalstore`, `NewTerminalStore`, `ErrTerminalStore`, `ErrTerminalExists`, and the exported `NewSupervisedTerminal` are gone with zero remaining references tree-wide, the new package-local `newSupervisedTerminal` is exercised at all three create paths, and the `Run` wrap-guard drop is consistent with the removed sentinel. CI green incl. E2E.

PR #293, opened 2026-05-21
fix: resolve reaper tracked exit reaped before RegisterChild

### LINKED ISSUE #286: terminalrunner: TestReaper_TrackedExitResolvesRace is flaky under -race
## Repro

Isolated worktree at the base commit (no feature branch changes):

```bash
git worktree add /tmp/sbsh-base eb71150
cd /tmp/sbsh-base
export GORACE="halt_on_error=1"
for i in $(seq 1 30); do
  go test -race -count=1 -run 'TestReaper_TrackedExitResolvesRace' ./internal/terminal/terminalrunner/... || echo "run $i FAILED"
done
```

Fails ~1/30 on `eb71150` (also reproduced on the current `main`). The non-race `make test` target does not surface it (timing is far less likely without `-race`).

## Observed

```
--- FAIL: TestReaper_TrackedExitResolvesRace (3.01s)
    reaper_linux_test.go:94: timed out waiting for reaper to deliver tracked exit
```

## Root cause

`internal/terminal/terminalrunner/reaper_linux_test.go:72` starts the child (`/bin/sh -c "exit 7"`) and only calls `r.RegisterChild(cmd.Process.Pid)` *after* `cmd.Start` returns. The reaper's `SIGCHLD` drain loop (`reaper.loop` → `drainOnce` → `handleReaped`) can reap the already-exited child before `RegisterChild` sets `trackedPid`. When that happens `handleReaped` sees `trackedPid == 0`, treats the pid as an orphan, and never delivers on `TrackedExitCh` — so the test's 3s `select` times out. `-race` widens the window enough to hit it intermittently.

This is a test-only race; the production reaper behavior (deliver-once on `TrackedExitCh`) is correct. Confirmed independent of the phase-2 child-set work in #285, whose reaper changes do not touch the tracked-exit delivery path.

## Suggested fix

Make registration race-free in the test by registering interest before the child can exit — e.g. spawn a child that blocks until released (`sh -c 'read; exit 7'` with the test closing stdin after `RegisterChild`), or have the reaper record exit codes for not-yet-registered pids so a late `RegisterChild` still resolves. The latter also hardens the real `startPty` register-after-start window, so it is the more durable option.

eminwux:
## Summary

`TestReaper_TrackedExitResolvesRace` was flaky (~1/30) under `-race` because the
fix it asserts didn't exist in the production reaper. `startPty`
(`terminal.go:180`) and the tests call `reaper.RegisterChild` *after*
`cmd.Start` (`terminal.go:148`), so a fast-exiting tracked child can be reaped
by the `SIGCHLD` drain loop before `trackedPid` is set. When that happened,
`handleReaped` saw `trackedPid == 0`, treated the pid as an orphan, and never
delivered on `TrackedExitCh` — the test's 3s `select` timed out. The same
register-after-start window exists in the real `startPty` path, not just the test.

This implements the issue's **suggested fix B** (record exit codes for
not-yet-registered pids), which the issue author flagged as "the more durable
option" because it also hardens the production window:

- `handleReaped` now records the exit code of any pid reaped while
  `trackedPid == 0` into a bounded `pendingExits` map (populated only during the
  pre-registration window).
- `RegisterChild` resolves immediately from `pendingExits` if the child already
  exited before its pid was known.
- The match/record decision is taken under a single `r.mu` acquisition so it
  cannot race `RegisterChild`. Delivery is extracted into
  `deliverTrackedExitLocked` (buffered-1 channel + `trackedOnce`, so the send
  never blocks under the lock).

A new deterministic test, `TestReaper_ExitBeforeRegisterResolves`, drives
`drainOnce` by hand to force the reap-before-register ordering (rather than
relying on `-race` timing), proving a late `RegisterChild` resolves the recorded
exit.

### Why fix B over fix A (test-only)

Fix A (block the child until after `RegisterChild`) would have made the test
pass by *avoiding* the race — but `TestReaper_TrackedExitResolvesRace` is named
and documented to verify the reaper *resolves* the race, so fix A defeats the
test's stated purpose and leaves the identical latent window in production
`startPty`. Fix B makes the asserted behavior real and is the author's
recommended option. Documented here so the reviewer can push back if the
broader-than-test-only blast radius oversteps.

## Test plan

- [x] `go build ./...`
- [x] `make sbsh-sb` (verified ELF output for `sbsh` and `sb`)
- [x] `make test` (full CI target, incl. e2e)
- [x] `make test-race` (race target — directly relevant to this fix)
- [x] 30× `go test -race -count=1 -run TestReaper ./internal/terminal/terminalrunner/...` — 0 failures (was ~1/30 before)

Closes #286

### DIFF
--- a/internal/terminal/terminalrunner/reaper_linux.go
+++ b/internal/terminal/terminalrunner/reaper_linux.go
@@ -45,6 +45,14 @@ type reaper struct {
 	trackedOnce    sync.Once
 	trackedExitCh  chan int // buffered 1; receives exit code for trackedPid
 	trackedStopped bool
+	// pendingExits records exit codes of children reaped before RegisterChild
+	// set trackedPid. startPty (and the tests) call RegisterChild *after*
+	// cmd.Start, so a fast-exiting child can be reaped here before its pid is
+	// known. Recording the exit lets a late RegisterChild still resolve the
+	// tracked exit instead of silently treating the child as an orphan and
+	// timing out on TrackedExitCh. Only populated while trackedPid == 0, so it
+	// is bounded to the (microsecond) pre-registration window.
+	pendingExits map[int]int
 }
 
 // newReaper constructs a reaper but does not start it. Call Start to install
@@ -93,6 +101,15 @@ func (r *reaper) RegisterChild(pid int) {
 		return
 	}
 	r.trackedPid = pid
+	// If the child already exited before we learned its pid, the reaper
+	// recorded its exit code in pendingExits; resolve the tracked exit now.
+	if code, ok := r.pendingExits[pid]; ok && !r.trackedStopped {
+		r.deliverTrackedExitLocked(code)
+		r.logger.Info("reaper: tracked child exited before registration", "pid", pid, "exit", code)
+	}
+	// Past registration, an unmatched reaped pid is a genuine orphan, so the
+	// recording window is closed; drop the map.
+	r.pendingExits = nil
 }
 
 // TrackedExitCh returns a channel that receives the exit code of the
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #293 Review — `fix: resolve reaper tracked exit reaped before RegisterChild`**

LGTM — folds the match/record decision into a single `r.mu`-held switch in `handleReaped` and resolves a late `RegisterChild` from a `pendingExits` map, closing the reap-before-register race while keeping the buffered(1) + `trackedOnce` send non-blocking under the lock; the `trackedPid == 0` guard keeps the map bounded to the pre-registration window, and the new deterministic test drives `drainOnce` by hand to force the ordering. Commit verified, full race target green.

PR #294, opened 2026-05-21
feat(server): custom RPC handler registration on pkg/terminal/server

### LINKED ISSUE #291: feat(server): custom RPC handler registration on pkg/terminal/server
## Context

kukeon's `kuketty` wrapper embeds sbsh's terminal server to serve the attach/control protocol inside an attachable container — it calls `sbshserver.New(&doc.Spec, logger)` then `srv.Serve(ctx, listener)` and is otherwise pure passthrough (no server code of its own).

kukeon now needs `kuketty` to surface **container-setup outcomes** to its daemon (`kukeond`): the results of work `kuketty` runs *before* `Serve` — repository clones and run-once init stages — must be reported back so the daemon can fold them into the container's `Status`. The daemon already dials `kuketty`'s control socket (it's the same socket `kuke attach` uses), so the natural channel is a custom RPC verb on that existing listener.

Today `New(spec, logger)` exposes no way to register additional JSON-RPC verbs alongside the built-in attach/control protocol. An embedder can't add a verb without forking the package or standing up a second listener on a separate socket — both worse than reusing the established, already-permissioned control socket.

## Proposal

Expose an extension point on `pkg/terminal/server` that lets an embedder register custom JSON-RPC method handlers, served on the **same listener** as the built-in protocol. Exact shape is sbsh's call — e.g. a functional option (`server.WithHandlers(...)`), a `Register(method, handler)` method, or a handler map argument to `New`.

Requirements:
- Custom verbs are served on the same socket/connection as the built-in attach/control protocol.
- Custom method names must not collide with built-in ones — namespacing convention or an explicit registration error.
- Built-in attach/control behavior is unchanged when no custom handler is registered.

## Acceptance criteria

- [ ] An embedder can register one or more custom JSON-RPC verbs via the public `pkg/terminal/server` surface.
- [ ] Registered verbs are dispatched over the same listener as the built-in protocol.
- [ ] The built-in attach/control protocol is unaffected (regression-tested).
- [ ] A custom name colliding with a built-in method is rejected (error at registration or documented precedence).
- [ ] Unit test registers a dummy verb and asserts dispatch + isolation from built-ins.
- [ ] `make sbsh-sb` + `go vet ./...` + `make test` green.

## Notes

- Consumer: kukeon issue eminwux/kukeon#617 (container-setup status over RPC into `ContainerStatus`). Cross-repo dependency — kukeon#617 is merge-blocked on this landing + an sbsh version bump.
- Observed current surface: `New(&doc.Spec, logger)` + `srv.Serve(ctx, listener)` (kukeon `cmd/kuketty/main.go`).

eminwux:
## Summary

- Adds an extension point on `pkg/terminal/server` so embedders (e.g. kuketty surfacing container-setup status to `kukeond`, eminwux/kukeon#617/#642) can register custom JSON-RPC verbs on the **same** control-socket listener as the built-in attach/control protocol — no fork, no second socket.
- Public surface: `server.New(spec, logger, server.WithHandlers(server.Handler{Name, Receiver})...)`. `Handler.Name` is the net/rpc service name (the part before the dot in a `"Service.Method"` wire call); `Receiver` is any value whose exported methods follow net/rpc's `func(args T1, reply *T2) error` scheme.

## Design notes (for reviewer push-back)

- **Namespacing comes free from net/rpc.** Custom verbs register under a distinct service name on the same `rpc.Server`; dispatch is by full `"Service.Method"`, so a custom service's methods can never collide with the built-in `TerminalController.*` methods. The only possible collision is at the *service-name* level.
- **Collision handling is belt-and-suspenders.** `New` validates up front and rejects an empty `Name`, a `Name` equal to the built-in service (`api.TerminalService`), a nil `Receiver`, or a duplicate `Name` within the set — a clean constructor error rather than an opaque failure on the Serve goroutine. As defense-in-depth, `StartServer` registers the built-in service *first*, then customs, so net/rpc's own "service already defined" rejection also fires (a receiver whose methods don't match net/rpc's scheme surfaces as an error from `Serve` at registration time).
- **Threading.** `terminalrunner.StartServer` gained a variadic `extra ...terminalrpc.ExtraHandler` param (interface, exec impl, fake updated). Variadic keeps the existing non-facade caller (`internal/terminal/controller.go`) compiling unchanged. The public `Handler` lives in `pkg/terminal/server` so embedders don't import `internal/...`.
- Built-in behavior is unchanged when no handler is registered (the existing facade tests still pass).

## Test plan

- [x] `make sbsh-sb` — builds `sbsh`, hardlinks `sb`; `file ./sbsh` → ELF executable
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (cmd/sb, cmd/sbsh, internal/terminal, internal/client, integration get, e2e — all green)
- [x] `go test ./pkg/terminal/server/...` — new `TestServer_CustomHandler_DispatchAndIsolation` (registers a dummy verb, asserts on-socket dispatch + built-in `Ping` isolation + unknown-method error) and `TestServer_New_RejectsBadHandlers` (collision/empty/nil/duplicate)
- [x] `pre-commit run --files <changed>` — golangci-lint, gofmt, addlicense all pass

## Acceptance criteria

- [x] Embedder can register one or more custom JSON-RPC verbs via the public `pkg/terminal/server` surface (`WithHandlers`)
- [x] Registered verbs dispatched over the same listener as the built-in protocol
- [x] Built-in attach/control protocol unaffected (regression-tested via `Ping` on the same socket)
- [x] A custom name colliding with a built-in method is rejected (error at registration: `New` + net/rpc)
- [x] Unit test registers a dummy verb and asserts dispatch + isolation from built-ins
- [x] `make sbsh-sb` + `go vet ./...` + `make test` green

Closes #291

### DIFF
--- a/internal/terminal/controller_test.go
+++ b/internal/terminal/controller_test.go
@@ -156,7 +156,7 @@ func Test_ErrStartRPCServer(t *testing.T) {
 			OpenSocketCtrlFunc: func() error {
 				return nil
 			},
-			StartServerFunc: func(_ context.Context, _ *terminalrpc.TerminalControllerRPC, readyCh chan error, _ chan error) {
+			StartServerFunc: func(_ context.Context, _ *terminalrpc.TerminalControllerRPC, readyCh chan error, _ chan error, _ ...terminalrpc.ExtraHandler) {
 				readyCh <- errors.New("make server fail")
 			},
 		}
@@ -197,7 +197,7 @@ func Test_ErrStartTerminal(t *testing.T) {
 			OpenSocketCtrlFunc: func() error {
 				return nil
 			},
-			StartServerFunc: func(_ context.Context, _ *terminalrpc.TerminalControllerRPC, readyCh chan error, _ chan error) {
+			StartServerFunc: func(_ context.Context, _ *terminalrpc.TerminalControllerRPC, readyCh chan error, _ chan error, _ ...terminalrpc.ExtraHandler) {
 				readyCh <- nil
 			},
 			StartTerminalFunc: func(_ chan<- terminalrunner.Event) error {
@@ -239,7 +239,7 @@ func Test_ErrSetupShell(t *testing.T) {
 			OpenSocketCtrlFunc: func() error {
 				return nil
 			},
-			StartServerFunc: func(_ context.Context, _ *terminalrpc.TerminalControllerRPC, readyCh chan error, _ chan error) {
+			StartServerFunc: func(_ context.Context, _ *terminalrpc.TerminalControllerRPC, readyCh chan error, _ chan error, _ ...terminalrpc.ExtraHandler) {
 				readyCh <- nil
 			},
 			StartTerminalFunc: func(_ chan<- terminalrunner.Event) error {
@@ -284,7 +284,7 @@ func Test_ErrInitShell(t *testing.T) {
 			OpenSocketCtrlFunc: func() error {
 				return nil
 			},
-			StartServerFunc: func(_ context.Context, _ *terminalrpc.TerminalControllerRPC, readyCh chan error, _ chan error) {
+			StartServerFunc: func(_ context.Context, _ *terminalrpc.TerminalControllerRPC, readyCh chan error, _ chan error, _ ...terminalrpc.ExtraHandler) {
 				readyCh <- nil
 			},
 			StartTerminalFunc: func(_ chan<- terminalrunner.Event) error {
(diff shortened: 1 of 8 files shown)


eminwux:
**PR #294 Review — `feat(server): custom RPC handler registration on pkg/terminal/server`**

LGTM — adds a `WithHandlers` extension point on the public `pkg/terminal/server` façade with constructor-time validation (empty/built-in-collision/nil/duplicate) backed by net/rpc's own service-name guard; built-in protocol unaffected, registration error path mirrors the existing built-in path exactly, and dispatch/isolation/reject cases are all tested.

PR #300, opened 2026-05-21
feat(terminalrunner): vt-parser screen model + Screenshot RPC

### LINKED ISSUE #295: terminal: introduce vt-parser screen model + Screenshot RPC
## Context

sbsh is currently a **byte-stream relay**: it stores the PTY capture (raw bytes, always-on, append-only at `internal/terminal/terminalrunner/terminal.go:199`) and replays it, but never *decodes* it — the client's terminal is the only decoder. `go.mod` carries only `creack/pty` + `golang.org/x/term`; there is no vt parser. Consequently there is no way to ask "what is on this terminal's screen right now," and replay/attach behavior can only manipulate raw bytes, not screen state.

This issue introduces the missing decode capability and proves it with its cheapest consumer.

## Proposal

- Introduce an in-server **vt-parser maintaining a screen model** (current grid: cells, cursor position, colors/SGR, alt-screen state) in `internal/terminal/terminalrunner`.
- Feed it **warm** from the existing `DynamicMultiWriter` (the same writer that writes the capture file) so the model stays current without re-reading the file. Warm feed keeps screenshot/attach latency O(1) and decouples the model from capture-file segmentation (matters once rotation lands).
- Expose it as a **built-in `Screenshot` `TerminalController` method** — added to the api interface, `internal/terminal/terminalrpc/rpc_server.go`, and `pkg/terminal/server/adapter.go`, alongside `Ping`/`Resize`/`Attach`/etc. **Not** via the #294 `ExtraHandler` seam, which is reserved for downstream library consumers; a screenshot that ships with sbsh is a first-class capability.
- Add a `sb screenshot <terminal>` CLI subcommand that calls the RPC and prints the current rendered screen (text grid and/or ANSI).

This is the parser's first and cheapest consumer — it de-risks the parser introduction before the more involved render-on-attach work (#71) builds on the same screen model.

## Acceptance criteria

- [ ] A vt-parser + screen model exists in `terminalrunner`, fed warm by the multiwriter. (Dependency choice — library vs minimal in-tree — is dev's call.)
- [ ] `Screenshot` is a built-in `TerminalController` method (api interface + `rpc_server.go` + `adapter.go`), not an `ExtraHandler`.
- [ ] `sb screenshot <terminal>` prints the current rendered screen.
- [ ] Tests cover: plain output, cursor positioning, color/SGR state, and an alt-screen app (e.g. vim/htop) — the screenshot reflects the live grid, not raw bytes.
- [ ] The capture file writer is unchanged — the parser is an *additional* reader of the stream, not a replacement.

## Notes

- **First phase of the capture-rendering arc.** Blocks the render-on-attach rewrite of #71, which repaints from this screen model.
- Downstream of the arc: capture rotation+retention and asciicast capture-format issues to follow (filed separately).
- Warm-vs-cold parser feed: **warm** (multiwriter-fed) chosen deliberately — see Proposal.

## Suggested slicing

Single PR.

eminwux:
## Summary

First phase of the capture-rendering arc (#295). sbsh was a pure byte-stream relay with no way to ask "what is on this terminal's screen right now." This introduces an in-server vt-parser maintaining a screen model and proves it with its cheapest consumer, `sb screenshot`.

- **Screen model** (`internal/terminal/terminalrunner/terminal_screen.go`): wraps a vt10x terminal emulator, registered as an *additional* warm sink on the existing `DynamicMultiWriter` — fed the same bytes as the capture file, so the model stays current without re-reading the file and screenshot latency is O(1). The capture writer is unchanged; the parser is a second reader of the stream.
- **`Screenshot` built-in `TerminalController` method**, wired through the full stack alongside `Ping`/`Resize`/`Subscribe`: api interface (`pkg/api/terminal.go`), `terminalrpc/rpc_server.go`, `pkg/terminal/server/adapter.go`, the `Controller`, the `TerminalRunner` interface + `Exec`, and the RPC client (`pkg/rpcclient/terminal`). **Not** the #294 `ExtraHandler` seam — that is reserved for downstream library consumers; a screenshot that ships with sbsh is first-class.
- **`sb screenshot <terminal>` CLI**: prints the rendered screen. Default output carries SGR color sequences; `--plain` prints the bare text grid.
- `Resize` now also resizes the screen model so the grid tracks the PTY winsize.

### Dependency choice (AC leaves this to dev)

Used `github.com/hinshun/vt10x` (MIT) rather than a hand-rolled parser. A minimal in-tree parser robust enough to pass the alt-screen + SGR ACs would be far higher blast radius; vt10x maintains the full grid (cells, cursor, colors, alt-screen), is internally synchronized, and is an `io.Writer` so it drops directly into the multiwriter fan-out. It is now a direct dependency in `go.mod`.

### Non-obvious calls (reviewer please confirm)

- **ScreenshotResult shape**: returns `Cols/Rows/CursorX/CursorY/CursorVisible/AltScreen` plus `Text` (plain grid) and `ANSI` (grid re-rendered with SGR FG/BG). ANSI renders foreground/background colors (the exported `Glyph.FG/BG`); vt10x's text-attribute bits (bold/underline/reverse) are unexported, so they are intentionally not reconstructed. Color is the SGR state the AC calls out.
- **Trailing blank rows are trimmed** from the rendered output for readability; `Cols`/`Rows` still report the true grid size.
- **UTF-8 carry**: vt10x.Write drops a trailing partial rune at a chunk boundary; the model carries the unconsumed bytes to the next Write so a multibyte rune split across PTY reads decodes intact (covered by a test).
- **No explicit teardown** of the screen sink on Close: unlike subscribers it holds no fd or goroutine, so it is GC'd with the runner; the capture-file close sequencing is untouched.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` (CI unit target)
- [x] `go test -tags=integration ./cmd/sb/get/...` (CI integration target)
- [x] `go test ./e2e` (CI e2e target)
- [x] `go test -race` on the new screen tests (concurrent warm-feed Write + Screenshot snapshot)
- [x] library-consumer example builds + vets (public interface change)
- [x] `pre-commit run` on staged snapshot (go-fmt, go-mod-tidy, golangci-lint, addlicense) — all pass
- [x] Live smoke: launched `sbsh` under a PTY, ran commands incl. `printf '\033[31mREDTEXT\033[0m'`, then `sb screenshot <name>` — `--plain` shows the decoded grid; default shows ANSI with `\x1b[31;49mREDTEXT\x1b[0m` and the colored prompt, proving SGR decode end-to-end over RPC.

New unit tests cover the AC matrix: plain output, cursor positioning (+ visibility), color/SGR (16-color, 256-color, background), alt-screen enter/exit (DECSET 1049, as vim/htop emit), the warm-feed-doesn't-disturb-capture invariant, and `Exec.Screenshot` delegation.

## Acceptance criteria

- [x] vt-parser + screen model in `terminalrunner`, fed warm by the multiwriter
- [x] `Screenshot` is a built-in `TerminalController` method (api interface + `rpc_server.go` + `adapter.go`), not an `ExtraHandler`
- [x] `sb screenshot <terminal>` prints the current rendered screen
- [x] Tests cover plain output, cursor positioning, color/SGR state, and an alt-screen app — reflecting the live grid, not raw bytes
- [x] The capture file writer is unchanged — the parser is an additional reader

Closes #295

### DIFF
--- a/cmd/sb/sb.go
+++ b/cmd/sb/sb.go
@@ -30,6 +30,7 @@ import (
 	"github.com/eminwux/sbsh/cmd/sb/get"
 	"github.com/eminwux/sbsh/cmd/sb/prune"
 	"github.com/eminwux/sbsh/cmd/sb/read"
+	"github.com/eminwux/sbsh/cmd/sb/screenshot"
 	"github.com/eminwux/sbsh/cmd/sb/stop"
 	"github.com/eminwux/sbsh/cmd/sb/validate"
 	"github.com/eminwux/sbsh/cmd/sb/version"
@@ -126,6 +127,7 @@ func setupRootCmd(rootCmd *cobra.Command) error {
 	rootCmd.AddCommand(get.NewGetCmd())
 	rootCmd.AddCommand(write.NewWriteCmd())
 	rootCmd.AddCommand(read.NewReadCmd())
+	rootCmd.AddCommand(screenshot.NewScreenshotCmd())
 
 	rootCmd.PersistentFlags().String("config", "", "config file (default is $HOME/.sbsh/config.yaml)")
 	if err := viper.BindPFlag(config.SB_ROOT_CONFIG.ViperKey, rootCmd.PersistentFlags().Lookup("config")); err != nil {
--- a/cmd/sb/screenshot/screenshot.go
+++ b/cmd/sb/screenshot/screenshot.go
@@ -1,0 +1,112 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package screenshot
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+
+	"github.com/eminwux/sbsh/cmd/config"
+	"github.com/eminwux/sbsh/cmd/sb/get"
+	"github.com/eminwux/sbsh/cmd/types"
+	"github.com/eminwux/sbsh/internal/discovery"
+	"github.com/eminwux/sbsh/internal/errdefs"
+	"github.com/eminwux/sbsh/pkg/api"
+	rpcterminal "github.com/eminwux/sbsh/pkg/rpcclient/terminal"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+type screenshotOpts struct {
+	runPath string
+	name    string
+	plain   bool
+}
+
+func NewScreenshotCmd() *cobra.Command {
+	screenshotCmd := &cobra.Command{
+		Use:   "screenshot <name>",
+		Short: "Print a terminal's current rendered screen",
+		Long: `Print the current rendered screen of a terminal.
+
+sbsh decodes the live PTY output into a screen model (the same grid an
+attached client would see) and prints it. By default the output includes
+SGR color sequences so colors are preserved; use --plain for the bare text
+grid with no escape sequences.`,
+		SilenceUsage:      true,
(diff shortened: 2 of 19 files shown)


eminwux:
**PR #300 Review — `feat(terminalrunner): vt-parser screen model + Screenshot RPC`**

LGTM — adds an O(1) screen snapshot as a warm second sink on the existing multiwriter (capture writer untouched) and threads `Screenshot` cleanly through every interface (`pkg/api`, `rpcclient` `Client`, adapter, both fakes); UTF-8 carry is bounded, concurrent feed/snapshot is proven under `-race`, and the SGR encoding faithfully mirrors vt10x's color packing.

PR #301, opened 2026-05-21
fix(terminalrunner): seed attach replay through ring to end write-deadline race

### LINKED ISSUE #299: terminal: attach to busy terminal disconnects new client (replay write races drain deadline)
## Context

When a client attaches to a terminal that **already has an attacher** (an active, output-producing session), the new client sees the full capture replay dumped to its terminal and is then **immediately disconnected**. Observed log:

```
ERROR "error in pty->conn initial write" err=write unix @->@: i/o timeout client=0xc000241ce0
INFO  "client connection handler exiting" client=0xc000241ce0
INFO  "writer removed" remaining_writers=2
ERROR "stdin->socket read error" error=read unix @->@: use of closed network connection
ERROR "read/write error" ... could not read stdin->socket: read unix @->@: use of closed network connection
INFO  "connManager: context cancelled or error received, beginning shutdown"
```

This breaks the headline multi-attach use case: attaching to a busy shared session kicks the new client.

## Root cause

In `internal/terminal/terminalrunner/connections.go`, `handleClient` performs the initial capture replay **directly on `client.conn`** *after* it has already started the bounded-writer drain goroutine on that **same conn**:

- `connections.go:113-116` — registers `aw` (a `subscriberWriter` wrapping `client.conn`) into the PTY fan-out `multiOutW` and starts `go aw.Run()`. From this instant the drain goroutine writes live PTY output to `client.conn`, calling `s.conn.SetWriteDeadline(now + subscriberWriteTimeout)` before **every** write (`subscriber.go:148`).
- `connections.go:147` — the main goroutine separately does `client.conn.Write(log)`, where `log` is the **entire capture file** (`io.go:23` `readLogFile` → `os.ReadFile(captureFile)`).

`SetWriteDeadline` on a `net.UnixConn` is **connection-global and absolute**, not per-`Write`-call. So the drain goroutine's `now+5s` deadline applies to the large replay write at `connections.go:147` too (and the drain never clears it), and the two goroutines interleave bytes on the same fd.

## Why it only bites an already-attached terminal

- `remaining_writers=2` confirms other attachers were live → the session was actively producing output → the drain goroutine was actively writing and resetting the 5s deadline *concurrently* with the replay.
- An already-attached, long-running terminal has a **large** capture file. The replay `Write(log)` blocks filling the socket buffer while a slow-rendering peer drains it; once the drain goroutine's `now+5s` deadline elapses mid-replay, the blocked replay write trips it → `i/o timeout` (`connections.go:148`) → `handleClient` returns → `detach` fires → handler exits.

A fresh terminal has a tiny/empty capture and little concurrent output, so the replay finishes inside the window and the bug stays hidden.

## Proposal

Restore the design invariant that the drain goroutine is the **sole** writer of `client.conn` (the `subscriberWriter` ring exists precisely so timed writes are owned by one goroutine — see the comment at `connections.go:110-112`). The replay write at `connections.go:147` violates that. Two options:

1. **(Preferred)** Enqueue the replay through the ring — `aw.Write(log)` instead of `client.conn.Write(log)` — so the single drain goroutine serializes replay + live output and owns all deadlines. Note the `defaultSubscriberBufferBytes` (1 MiB) cap in `subscriber.go:37`: a replay larger than the ring would trip the lagged path, so the replay enqueue likely needs to bypass or pre-size that bound.
2. Replay **before** wiring the fan-out — move `client.conn.Write(log)` ahead of `multiOutW.Add(aw)` / `go aw.Run()` so no second writer/deadline exists during replay. This reopens the snapshot-vs-live gap the ring was meant to close, so option 1 is the more faithful fix.

## Acceptance criteria

- [ ] Attaching a second client to a terminal with an active output-producing session and a large (>1 MiB) capture file no longer disconnects the new client with `i/o timeout`.
- [ ] The capture replay and live PTY output reach the newly attached client without byte interleaving/corruption.
- [ ] No `SetWriteDeadline` set by the drain goroutine can trip the initial replay write (single-writer ownership of `client.conn`, or replay completed before the drain starts).
- [ ] Regression test exercising attach-to-busy-terminal with capture larger than the subscriber ring bound.

## Notes

- Files: `internal/terminal/terminalrunner/connections.go` (`handleClient`, lines 113-116 and 147), `internal/terminal/terminalrunner/subscriber.go` (`Run`/`SetWriteDeadline`, line 148; ring cap line 37), `internal/terminal/terminalrunner/io.go:23` (`readLogFile`).
- Related but distinct: #71 (repaint current screen on attach; `--full-capture` for full replay) concerns replay *scope/policy*, not this concurrent-writer correctness bug.

eminwux:
## Summary

- Attaching a second client to a busy terminal with a large (>1 MiB) capture file dumped the full replay and then immediately disconnected the new client with `i/o timeout`. Root cause (per #299): `handleClient` wrote the entire capture replay **directly to `client.conn`** *after* it had already started the bounded-writer drain goroutine on the **same conn**. `SetWriteDeadline` on a `net.UnixConn` is connection-global and absolute, so the drain goroutine's `now+5s` deadline applied to (and tripped) the large replay write, and the two goroutines interleaved bytes on the same fd.
- This implements the issue's **preferred option 1**: the replay is now enqueued *through the subscriber ring* so the single drain goroutine is the sole writer of `client.conn` and owns all write deadlines.

## Changes

- `subscriber.go`: new `seedReplay(b)` — pre-loads the ring with the capture replay (so it drains ahead of any live PTY output) and grows `maxBytes` by the replay size. The bound growth is essential: without it, a replay near or above the 1 MiB ring cap would make the *first* live `Write` overflow and lagged-drop a freshly attached client. The original 1 MiB headroom for live output is preserved on top of the seed.
- `connections.go` (`handleClient`): read the capture **before** wiring the fan-out, seed it into the writer, then `multiOutW.Add` + `go aw.Run()`. Removed the direct `client.conn.Write(log)` second-writer path entirely.
- `subscriber_test.go`: regression test `TestSubscriber_SeededReplayPrecedesLiveOutputLargeCapture` — seeds a >1 MiB replay, delivers live output via the ring, and asserts the client receives `replay`-then-`live` in order with no `laggedNotice` and no truncation.

## Non-obvious calls (reviewer please confirm)

- **Behavior change on unreadable capture:** the old code did `return` (aborting `handleClient`) when `readLogFile` failed — but in the old ordering the writer/reader were already wired, so the attach survived with no replay. In the new ordering `readLogFile` runs *first*, so a `return` there would deny the attach entirely. I instead **fall through with an empty replay** (log the warning, `log = nil`) so a missing/unreadable capture never denies the live attach. This is strictly more lenient and matches the old observable outcome (attach continues, no replay).
- **Snapshot-vs-live window:** there is a microsecond gap between `readLogFile` returning and `multiOutW.Add`. The issue notes option 2 (replay before fan-out) "reopens the snapshot-vs-live gap"; option 1 minimizes it because the seed is an in-memory copy followed immediately by `Add`, versus writing the whole (potentially slow) replay to the conn before `Add`. The capture file is itself a writer in `multiOutW`, so eliminating the window entirely would require snapshotting under the multiwriter lock — out of scope for this fix and not required by the AC.
- **Elevated `maxBytes` persists for the client's lifetime:** after the replay drains, the ring's lag bound stays at `replay + 1 MiB` rather than resetting to 1 MiB. This is harmless (still bounded, per-client, torn down on detach) and avoids a reset race; chose simplicity over reclaiming the slack.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` (full CI unit target)
- [x] `go test ./internal/terminal/terminalrunner/... -run Subscriber -race` (race detector, incl. new regression test)
- [x] library-consumer example `go build ./... && go vet ./...`
- [x] `HOME=$HOME E2E_BIN_DIR=$(pwd) go test ./e2e`
- [x] `pre-commit run --files <changed>` (golangci-lint, gofmt, addlicense, etc.)

## Acceptance criteria

- [x] Attaching a second client to a terminal with an active session and a >1 MiB capture no longer disconnects with `i/o timeout` — replay is seeded into the ring, sole-writer ownership removes the deadline race.
- [x] Capture replay and live PTY output reach the client without byte interleaving/corruption — seeded ahead of live bytes, serialized through one drain goroutine (asserted by the regression test).
- [x] No drain-goroutine `SetWriteDeadline` can trip the initial replay write — the direct second-writer path is removed; the drain owns `client.conn` exclusively.
- [x] Regression test exercising attach-to-busy-terminal with capture larger than the subscriber ring bound.

Closes #299

### DIFF
--- a/internal/terminal/terminalrunner/connections.go
+++ b/internal/terminal/terminalrunner/connections.go
@@ -107,10 +107,28 @@ func (sr *Exec) handleClient(client *ioClient) {
 		})
 	}
 
-	// WRITER: bounded ring + per-write deadline + lagged-detach. Adding
-	// the writer to multiOutW *before* starting Run is safe — Write just
-	// enqueues into the ring until Run starts draining.
+	// Snapshot the capture replay *before* wiring the fan-out so it can be
+	// seeded ahead of any live PTY output in the ring. Reading here (rather
+	// than writing it directly to client.conn after Add, as before) keeps
+	// the single drain goroutine the sole writer of client.conn: replay and
+	// live output are serialized through one ring, so the drain's per-write
+	// SetWriteDeadline can never race a large initial replay written on a
+	// second goroutine (issue #299). An unreadable capture is non-fatal —
+	// fall through with an empty replay rather than denying the live attach.
+	log, errLog := sr.readLogFile()
+	if errLog != nil {
+		sr.logger.Warn("failed to read log file for client attach", "err", errLog)
+		log = nil
+	}
+
+	// WRITER: bounded ring + per-write deadline + lagged-detach. Seed the
+	// replay into the ring (seedReplay grows the byte bound to fit it so a
+	// >1 MiB capture cannot trip the lagged path on the first live Write).
+	// Adding the writer to multiOutW *before* starting Run is safe — Write
+	// just enqueues into the ring until Run starts draining, and the seeded
+	// replay already sits ahead of any live bytes that arrive after Add.
 	aw := newSubscriberWriter(client.conn, defaultSubscriberBufferBytes, detach, sr.logger)
+	aw.seedReplay(log)
 	client.outWriter = aw
 	multiOutW.Add(aw)
 	go aw.Run()
@@ -135,17 +153,6 @@ func (sr *Exec) handleClient(client *ioClient) {
 
 	if errAttach := sr.updateTerminalAttachers(); errAttach != nil {
 		sr.logger.Warn("failed to update metadata on attach", "err", errAttach)
-		return
-	}
-
-	log, errLog := sr.readLogFile()
-	if errLog != nil {
-		sr.logger.Warn("failed to read log file for client attach", "err", errLog)
-		return
-	}
-
-	if _, errWrite := client.conn.Write(log); errWrite != nil {
-		sr.logger.Error("error in pty->conn initial write", "err", errWrite, "client", client.id)
 		return
 	}
 }
--- a/internal/terminal/terminalrunner/subscriber.go
+++ b/internal/terminal/terminalrunner/subscriber.go
@@ -99,6 +99,27 @@ func (s *subscriberWriter) Write(p []byte) (int, error) {
 	return len(p), nil
 }
 
+// seedReplay pre-loads the ring with the capture replay so the single
+// drain goroutine emits it ahead of any live PTY output, and grows the
+// byte bound by the replay size so a >1 MiB capture cannot make the first
+// live Write trip the lagged path and disconnect a freshly attached
+// client. Must be called before the writer is registered in the fan-out
+// (i.e. before any concurrent Write) and before Run starts. This keeps the
+// drain goroutine the sole writer of conn — its per-write SetWriteDeadline
+// can never race a large initial replay written on a second goroutine
+// (issue #299). A zero-length replay is a no-op, leaving the default bound.
+func (s *subscriberWriter) seedReplay(b []byte) {
+	if len(b) == 0 {
+		return
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	_, _ = s.buf.Write(b)
+	// Preserve the original headroom for live output on top of the seeded
+	// replay; without this the first live Write would overflow the bound.
+	s.maxBytes += len(b)
+}
+
 // Close requests a graceful shutdown. It signals drain so any pending
 // bytes are flushed to the conn; drain itself closes the conn on exit.
 // Safe to call repeatedly.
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #301 Review — `fix(terminalrunner): seed attach replay through ring to end write-deadline race`**

LGTM — seeds the capture replay through the subscriber ring so the drain goroutine is the sole writer of `client.conn`, eliminating the cross-goroutine `SetWriteDeadline` race (#299); `seedReplay` grows `maxBytes` by the replay size to preserve the 1 MiB live headroom, byte ordering is buffer-preserved, the unreadable-capture fall-through matches the old observable outcome, no dead code, commits verified, and the >1 MiB regression test asserts replay-then-live order with no lag.

PR #302, opened 2026-05-21
feat(attach): repaint current screen on attach; --full-capture for full replay

### LINKED ISSUE #71: Terminal - On attach: flag to print all capture; default tail ~100 lines
Allow users to dump full capture history on attach (with default tail ~100), modifying attach options and capture replay logic.

eminwux:
## Summary

- On attach, the terminal used to replay the **entire** capture file to every newly attached client (`readLogFile` → write whole buffer). On a long-lived terminal this floods the client with the full session history on every attach. This implements #71: **default attach now repaints the current screen** (synthesized from #295's vt-parser screen model, bounded to the viewport regardless of session length), and **`--full-capture`** opts back into the legacy full-raw-replay as an explicit escape hatch.
- `readLogFile` is renamed to `readCaptureFile` (it reads `Spec.CaptureFile` and the old name collided with `api.ClientSpec.LogFile`), including the internal warn line and the caller string.

## Design / how it works

- **Repaint synthesis** (`screenModel.repaint()`, `terminal_screen.go`): under the vt10x state lock, emits alt-screen enter (if active), clear+home, each visible row at an **absolute** cursor position (`CSI row;1H` — not newlines, which stairstep in raw mode) with its SGR colors, then restores the cursor position and visibility. Trailing blank rows are skipped (the clear already blanked them).
- **Flag plumbing** mirrors the `--disable-detach` precedent: `--full-capture` → `SB_ATTACH_FULL_CAPTURE` viper key → `pkg/attach.Options.FullCapture` → `api.ClientSpec.FullCapture`. Builder parity: `WithClientFullCapture`. Spawn parity: `buildClientAttachArgs` re-emits `--full-capture`.
- **The decision is server-side** (the server owns the attach replay in `handleClient`), so the flag is carried across the Attach RPC. The `Attach` RPC arg changed from `*api.ID` to a new `api.AttachRequest{ClientID, FullCapture}` (same shape as the existing `SubscribeRequest`), threaded through every Attach implementation + fakes. `clientrunner` reads `Spec.FullCapture` and forwards it.
- **Integrated with #301** (merged mid-task): the initial paint is **seeded through the subscriber ring** (`aw.seedReplay`) rather than written directly to `client.conn`, preserving #301's sole-writer / no-write-deadline-race guarantee for the full-capture (>1 MiB) path. `initialAttachPaint` returns the repaint or the raw capture; the seed mechanism is unchanged.

## Non-obvious calls (reviewer please confirm)

- **Server-side consumption diverges from `DetachKeystroke`.** The issue says plumb "the same way `--disable-detach` does." That precedent is consumed purely client-side (`clientrunner`); the replay is server-side, so `FullCapture` additionally crosses the Attach RPC. The CLI→viper→`ClientSpec` portion matches the precedent exactly; the extra hop is inherent to where the behavior lives. This changes the public `Attach` RPC signature (`pkg/api`, `pkg/rpcclient`) — flagging in case that surface is sensitive.
- **Empty paint is non-fatal.** A missing/unreadable capture (full-capture path) or an absent screen model (repaint path) seeds nothing and the live attach proceeds — strictly more lenient than denying the attach.
- **`sbsh` root (create+attach) is left on the default repaint path.** A brand-new terminal has an empty screen, so repaint = clear screen; `--full-capture` is only wired on `sb attach`, matching the issue's `sb attach` framing.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full CI target: cmd/sb, cmd/sbsh, internal/terminal, internal/client, `-tags=integration ./cmd/sb/get`, e2e)
- [x] `go test ./pkg/...` (incl. server attach fd-leak / paused-attacher)
- [x] `go test -race ./internal/terminal/terminalrunner/... -run 'Subscriber|Repaint|InitialAttachPaint|Screen|Client'`
- [x] library-consumer example `go build ./... && go vet ./...`
- [x] `pre-commit run --files <changed>` (golangci-lint, gofmt, addlicense)
- [x] Rebased onto current `origin/main` (resolved the #301 overlap in `handleClient`)

## Acceptance criteria

- [x] Default `sb attach` repaints the current screen from the screen model, not the whole capture (`screenModel.repaint()`, seeded via `initialAttachPaint`).
- [x] Alt-screen reattach shows a correct, uncorrupted screen — `TestRepaintAltScreen` (re-enters the alt buffer, paints alt contents, no primary leak).
- [x] `--full-capture` replays the entire raw capture buffer; plumbs through `ClientSpec` like `--disable-detach` (+ builder + spawn parity) — `TestInitialAttachPaintFullCaptureReplaysRaw`, `TestBuildClientAttachArgs_FullCapture`, builder tests.
- [x] `readLogFile` renamed to `readCaptureFile` (function, log message, caller string).
- [x] Tests cover empty capture (`TestRepaintEmptyScreen`, `TestInitialAttachPaintFullCaptureMissingFile`), plain-output screen (`TestRepaintPlainOutput`), alt-screen app (`TestRepaintAltScreen`), and the `--full-capture` path (`TestInitialAttachPaintFullCaptureReplaysRaw`).

Closes #71

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -113,6 +113,8 @@ var (
 		"sb/attach/disableDetachKeystroke",
 	)
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_ATTACH_FULL_CAPTURE = DefineKV("SB_ATTACH_FULL_CAPTURE", "sb/attach/fullCapture")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SB_DETACH_NAME = DefineKV("SB_DETACH_NAME", "sb/detach/name")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SB_DETACH_ID = DefineKV("SB_DETACH_ID", "sb/detach/id")
--- a/cmd/sb/attach/attach.go
+++ b/cmd/sb/attach/attach.go
@@ -172,6 +172,9 @@ func setupAttachCmdFlags(attachCmd *cobra.Command) {
 	_ = viper.BindPFlag(config.SB_ATTACH_SOCKET.ViperKey, attachCmd.Flags().Lookup("socket"))
 	attachCmd.Flags().Bool("disable-detach", false, "Disable detach keystroke (^] twice)")
 	_ = viper.BindPFlag(config.SB_ATTACH_DISABLE_DETACH_KEYSTROKE.ViperKey, attachCmd.Flags().Lookup("disable-detach"))
+	attachCmd.Flags().
+		Bool("full-capture", false, "Replay the entire raw capture buffer on attach instead of repainting the current screen")
+	_ = viper.BindPFlag(config.SB_ATTACH_FULL_CAPTURE.ViperKey, attachCmd.Flags().Lookup("full-capture"))
 }
 
 func run(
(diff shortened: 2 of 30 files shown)


eminwux:
**PR #302 Review — `feat(attach): repaint current screen on attach; --full-capture for full replay`**

LGTM — repaint synthesis is correct for raw mode (absolute CUP positioning, alt-screen re-entry, per-line SGR containment via `renderLine`'s `sgrReset`, cursor position/visibility restore), empty-paint is non-fatal, and `FullCapture` is threaded through every public façade (`ClientSpec`, `pkg/attach.Options`, `WithClientFullCapture`, spawn re-emit, the new `AttachRequest`). The Attach RPC signature change is propagated through all impls + fakes; rename `readLogFile`→`readCaptureFile` is complete with no stale refs; commits verified; tests cover repaint/alt-screen/empty/full-capture paths.

Issue #297, opened 2026-05-21
terminal: compress closed capture segments


eminwux:
## Context

Once capture rotation lands (#296), a long-lived terminal accumulates closed capture segments. Retention bounds total disk by dropping old segments; compression stretches that budget by shrinking the segments that are kept.

## Proposal

- **Compress closed segments with gzip on rotation.** The closed `<canonical>.NNNNNN` segment is gzip-compressed to `<canonical>.NNNNNN.gz` (stdlib `compress/gzip`, zero new dependency). The **live segment stays raw** — streaming-compressing a live append-only stream kills the seekable random access that `--full-capture` and `sb read` need, so only *closed* segments are compressed.
- **gzip is the chosen format**: reads are always sequential (`capture.Dump` / `ReadAll` `io.Copy` each segment end-to-end, never seeking), so gzip's lack of a seek index is a non-issue. The reader already anticipates `.gz` (the all-digit-suffix filter in `internal/capture/capture.go` was written to ignore compressed variants on the live splice). Terminal output is highly compressible and the retention budget is bounded (64 MiB), so gzip's ratio is more than sufficient; zstd/xz would add a dependency and CPU cost to win a marginal, capped saving.
- **Readers gain decompression**: the segment-reassembly logic from #296 (`--full-capture`, `sb read`, parser feed) learns to transparently decompress `.gz` closed segments. Dispatch on file extension at read time (`dumpOne` wraps a `.gz` file in a `gzip.Reader`) so the format stays swappable. This is a small extension of #296's reassembly, not a new reader path.

## Dependencies

- **Depends on #296** (rotation + retention). Compression operates on the closed segments rotation produces and extends the reader reassembly that issue introduces. Cannot ship without it — there is nothing to compress before rotation exists.
- Cross-link: asciicast capture-format issue — if capture is asciicast *and* compressed, the same raw-vs-compressed segment handling applies; coordinate at whichever lands second.

## Acceptance criteria

- [ ] Closed segments are gzip-compressed to `<canonical>.NNNNNN.gz` on rotation; the live segment remains raw and seekable.
- [ ] `--full-capture`, `sb read`, and the parser feed transparently decompress `.gz` closed segments during reassembly, dispatching on file extension.
- [ ] Tests cover: a reader spanning raw-live + gzip-closed segments, and round-trip integrity (compressed-then-read == original bytes).

## Suggested slicing

[...]

PR #303, opened 2026-05-21
feat(capture): rotate capture file into segments with retention

### LINKED ISSUE #296: terminal: capture file rotation + retention
## Context

The capture file is a single, operator-specified path (`Spec.CaptureFile`, public API at `pkg/api/terminal.go:71`, set via `SBSH_TERM_CAPTURE_FILE`), opened append-only (`internal/terminal/terminalrunner/terminal.go:199`) and fed by the `DynamicMultiWriter`. There is **no rotation or retention machinery** anywhere — a long-lived terminal grows its capture file unbounded. `internal/defaults` has no capture-size knobs.

The file has **three readers** today, all of which assume a single file: attach replay (`connections.go:141`), `sb read` (`cmd/sb/read/read.go:106`), and the `--full-capture` path landing in #71.

## Proposal

Rotate the capture into segments and bound total on-disk size by retention. The shape that fits our constraints (live append-only stream, random-access readers):

- **Live segment** stays raw, append-only, seekable — the canonical path operators know.
- **On a size threshold** (hard-coded default in `internal/defaults`, per #71's "hard-code now, configurable later" precedent), close the current segment → rename to a deterministic sibling → open a fresh live segment at the canonical path.
- **Retention** prunes oldest closed segments by count and/or total bytes, so disk is actually bounded (rotation alone only bounds per-file size).
- **Teach all three readers segment reassembly**: ordered closed segments + live segment, presented as one logical stream to `--full-capture`, `sb read`, and the parser feed.
- **Contract decision (load-bearing):** the canonical path is the **live segment**; reassembling full history is the job of `sb read` / `--full-capture`. A raw `cat` of the path shows the live tail only — document this. `CaptureMode` (already in `Spec`, `terminal.go:188`) is the natural integration seam.

## Dependencies

- **Depends on #71** — the reader reassembly extends the same `--full-capture` / `sb read` read path #71 settles. Building on #71's finished read path avoids two PRs rewriting the same seam.
- The **warm** parser (#295) is unaffected by segmentation — the screen model is independent of file layout — so attach/screenshot fidelity does not depend on this issue.
- Cross-link: interacts with the asciicast capture-format issue — if both land, segment boundaries must not split an asciicast record. Coordinate at whichever lands second; neither blocks the other.

## Acceptance criteria

- [ ] Capture rotates to a new segment at a hard-coded size threshold (default in `internal/defaults`).
- [ ] Closed segments are named deterministically; the canonical path is always the current live segment.
- [ ] Retention prunes oldest segments by count and/or total bytes; total capture disk is bounded.
- [ ] `--full-capture`, `sb read`, and the parser feed reassemble segments + live into one logical stream (correct order, no gaps/dupes at boundaries).
- [ ] `cat`-ing the canonical path shows the live tail; behavior is documented.
- [ ] Tests cover: no rotation (under threshold), single rotation, multiple rotations + retention prune, and a reader spanning the segment boundary.

## Suggested slicing

Single PR (rotation + retention together — rotation without retention adds files without bounding disk, so they only ship value as a pair). Large but atomic: the rotating writer can't ship without the readers understanding segments, or they break. If dev finds it too large at pickup, flag `resize-issue`.

## Notes

Third phase of the capture-rendering arc (#295 → #71 → this → compression follow-up). Blocks the compression issue (compression operates on closed segments produced here).

eminwux:
## Summary

Implements capture file rotation + retention (#296), the third phase of the capture-rendering arc (#295 → #71 → this).

- **New `internal/capture` package** owns the on-disk layout and reader-side reassembly. The canonical path is the **live, append-only segment**; closed segments are deterministic siblings `<canonical>.NNNNNN` (zero-padded, monotonically increasing — lower seq is older). Helpers: `SegmentPath`, `ClosedSegments` (ordered oldest-first), `NextSeq`, `OrderedPaths`, `ReadAll`, `Dump`. The matcher requires an all-digit suffix, so a future compressed `<canonical>.NNNNNN.gz` (#297) is ignored, not spliced in.
- **`captureWriter`** (`internal/terminal/terminalrunner/capture_writer.go`) replaces the raw `*os.File` as the always-on sink in `multiOutW`. On crossing the size threshold it closes → renames the live segment to its sibling → reopens a fresh canonical → prunes oldest closed segments by count **and** total bytes. Rotation failures never kill capture: it always restores a writable live segment and logs (degrades to "rotation skipped this round").
- **Hard-coded defaults** in `internal/defaults` (per #71's "hard-code now, configurable later" precedent): 8 MiB segment threshold, keep ≤8 closed segments / ≤64 MiB total.
- **Readers reassemble**: attach `--full-capture` (`readCaptureFile` → `capture.ReadAll`) and `sb read` (`dumpCapture` → `capture.Dump`) now present ordered closed segments + live as one logical stream.
- **Documented contract**: `cat`-ing the canonical path shows the live tail only — documented on `Spec.CaptureFile`, in the `internal/capture` package doc, and in `sb read --help`.

### Non-obvious calls (please push back if these overstep)

- **Rotation is always-on**, not gated on `CaptureMode`. The issue's "Contract decision" note calls `CaptureMode` the "natural integration seam", but `CaptureMode` (`pkg/api/terminal.go`) is the file-**permission** `os.FileMode`, not a behavior toggle, and no AC checkbox references it. I read the seam note as "perms apply per segment" (which they do — `applyArtifactPerms` runs on the canonical and every fresh segment) and made rotation unconditional with hard-coded thresholds, matching #71's precedent and the "hard-code now, configurable later" framing. This is a behavior change the AC explicitly accepts (the `cat` contract + reader reassembly are the point).
- **The "parser feed" reader needs no change.** The issue lists the parser feed as a third reader to "teach segment reassembly", but #295's screen model is fed warm as a live sink on `multiOutW` (`multiOutW.Add(screen)`) — it never reads the capture file, so file rotation is transparent to it. It already sees the complete byte stream regardless of segment layout. No code change; flagged here so the reviewer can confirm the reading.
- **Cross-segment splitting**: a single `Write` is never split across segments (rotation happens *after* a write that crosses the threshold), so a segment may exceed the threshold by up to one write. This keeps byte-stream integrity simple and matches logrotate-style "rotate after exceeding".

## Test plan

- [x] `make sbsh-sb` — canonical build; `file ./sbsh` confirms ELF 64-bit executable
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full CI suite (incl. `e2e`) passes
- [x] `go test ./internal/capture/...` — naming, ordering, `.gz`/non-digit filtering, reassembly order, never-written cases
- [x] `go test ./internal/terminal/terminalrunner/...` — no rotation under threshold; single rotation; multiple rotations + count prune; byte-based prune; reader spanning the boundary (gap/dupe-free); seq resumes across restart; plus the existing #229 fd-close regressions adapted to the rotating writer
- [x] `pre-commit run --files <changed>` — golangci-lint, gofmt, addlicense all pass

### Coverage note

`make test`'s globs don't include `internal/capture`, so its standalone unit tests don't run in CI. The rotation/retention/reassembly behavior is *also* exercised through `internal/terminal/terminalrunner` (which CI runs) — those tests drive `capture.ReadAll`/`ClosedSegments`/`SegmentPath` end-to-end. Surfacing for the reviewer; broadening the test glob is a separate Makefile concern.

## Acceptance criteria

- [x] Capture rotates to a new segment at a hard-coded size threshold (`defaults.CaptureSegmentMaxBytes`)
- [x] Closed segments named deterministically (`<canonical>.NNNNNN`); canonical path is always the live segment
- [x] Retention prunes oldest by count (`CaptureRetentionMaxSegments`) and total bytes (`CaptureRetentionMaxBytes`); total disk bounded
- [x] `--full-capture`, `sb read` reassemble segments + live in order, no gaps/dupes at boundaries (parser feed unaffected — see non-obvious calls)
- [x] `cat`-ing the canonical path shows the live tail; documented on `Spec.CaptureFile`, package doc, and `sb read --help`
- [x] Tests cover: no rotation, single rotation, multiple rotations + retention prune, and a reader spanning the segment boundary

Closes #296

### DIFF
--- a/cmd/sb/read/read.go
+++ b/cmd/sb/read/read.go
@@ -31,6 +31,7 @@ import (
 	"github.com/eminwux/sbsh/cmd/config"
 	"github.com/eminwux/sbsh/cmd/sb/get"
 	"github.com/eminwux/sbsh/cmd/types"
+	"github.com/eminwux/sbsh/internal/capture"
 	"github.com/eminwux/sbsh/internal/discovery"
 	"github.com/eminwux/sbsh/internal/errdefs"
 	"github.com/eminwux/sbsh/internal/naming"
@@ -51,6 +52,11 @@ func NewReadCmd() *cobra.Command {
 		Use:   "read <name>",
 		Short: "Print a terminal's capture log; with -f, follow live output",
 		Long: `Print the full capture log for a terminal.
+
+The capture is rotated into segments to bound on-disk size, so the file at the
+recorded capture path holds only the live tail. This command reassembles the
+full transcript — closed segments oldest-first, then the live segment — so you
+see the complete history; a raw 'cat' of the path shows the live tail only.
 
 With -f/--follow, also stream live PTY output after replaying the file, like
 'tail -f'. The follower subscribes to live output BEFORE replaying the capture
@@ -145,21 +151,12 @@ func runRead(
 	return streamLive(conn, stdout, stderr)
 }
 
+// dumpCapture streams the full transcript to stdout. The recorded capture
+// path is the live segment; capture.Dump reassembles closed rotated segments
+// oldest-first, then the live segment, into one logical stream. An absent
+// capture (never written to) produces no output and no error.
 func dumpCapture(path string, stdout io.Writer) error {
-	f, err := os.Open(path)
-	if err != nil {
-		if errors.Is(err, os.ErrNotExist) {
-			// Empty capture is indistinguishable from "never written to";
-			// treat absence as zero-byte output rather than an error.
-			return nil
-		}
-		return fmt.Errorf("open capture file %q: %w", path, err)
-	}
-	defer f.Close()
-	if _, err := io.Copy(stdout, f); err != nil {
-		return fmt.Errorf("read capture file %q: %w", path, err)
-	}
-	return nil
+	return capture.Dump(path, stdout)
 }
 
 // laggedNotice mirrors the bracketed core of the server sentinel (see
--- a/internal/capture/capture.go
+++ b/internal/capture/capture.go
@@ -1,0 +1,202 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Package capture defines the on-disk layout of a terminal's PTY transcript
+// and the helpers that reassemble it for readers.
+//
+// A capture is rotated into segments to bound per-file size, with retention
+// pruning the oldest closed segments to bound total disk use (see the rotating
+// writer in internal/terminal/terminalrunner). The layout is:
+//
+//   - The canonical path (the operator-facing capture file) is always the
+//     live, append-only, raw segment — the current tail of the transcript.
+//   - Closed segments are deterministic siblings named "<canonical>.NNNNNN",
+//     where NNNNNN is a zero-padded, monotonically increasing sequence number.
+//     Lower sequence numbers are older.
+//
+// Contract: a raw `cat` of the canonical path shows the live tail only.
+// Reassembling the full history is the job of the readers here (used by
+// `sb read`, attach `--full-capture`, and any other consumer of full
+// transcript): ordered closed segments oldest-first, then the live segment.
+//
+// The segment matcher requires an all-digit suffix, so unrelated siblings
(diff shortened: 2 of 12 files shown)


eminwux:
**PR #303 Review — `feat(capture): rotate capture file into segments with retention`**

Solid implementation. The `internal/capture` layout/reassembly helpers and the rotating `captureWriter` are clean; every error path in `rotate`/`reopenCanonical` restores a writable live segment so a transient failure degrades to "rotation skipped this round" rather than killing capture, and errors are wrapped with `%w`. Perms route through the existing `applyArtifactPerms` callback (extends the socket precedent rather than forking it), defaults follow #71's hard-code-now precedent, all symbols are used, commits verified, and tests cover no-rotation / single / multi+prune / byte-prune / boundary-spanning reader / seq-resume.

One non-blocking observation:

- **Concurrent read vs rotation** `internal/capture/capture.go:191-206`: `OrderedPaths` snapshots the closed-segment list, then `ReadAll`/`Dump` reads the live `canonical` path separately. If a rotation lands between the list and the live read, the just-rotated segment isn't in the path list and the live path now resolves to the fresh empty file — that segment's bytes are skipped. This is an inherent cross-process race (reader is a separate `sb read`/attach process that can't share the writer's mutex) and doesn't affect the post-close reassembly contract the tests verify; flagging only so it's a known limitation, not a regression.

LGTM — non-blocking, ship it.

PR #314, opened 2026-05-21
feat(capture): gzip-compress closed capture segments on rotation

### LINKED ISSUE #297: terminal: compress closed capture segments

eminwux:
## Summary

- Closes #297 — the fourth/final phase of the capture-rendering arc (#295 → #71 → #296 → this), building on the rotation + retention landed in #296/#303.
- On rotation, the just-closed `<canonical>.NNNNNN` segment is gzip-compressed to `<canonical>.NNNNNN.gz` (stdlib `compress/gzip`, zero new dependency). The **live segment stays raw and seekable** — only closed segments are compressed.
- Readers (`--full-capture`, `sb read`, the attach parser feed) transparently decompress `.gz` closed segments. All three consumers funnel through `capture.ReadAll`/`capture.Dump`, so adding decompression there — dispatching on the `.gz` file extension — covers every reader path with no new reader.

## Implementation notes (for reviewer scrutiny)

- **`capture.CompressSegment`** writes to a temp `.gz.tmp` sibling, renames it atomically into place, then removes the raw segment. This ordering guarantees a concurrent reader never observes a partial `.gz`, and at least one complete representation of the segment is on disk at every instant.
- **Crash/transient safety:** compression is best-effort in `rotate()` — on any compression error the raw segment is left on disk (still spliceable by readers) and the writer logs and continues. A compression failure degrades to "kept raw this round," never lost capture.
- **`ClosedSegments` now matches both `.NNNNNN` and `.NNNNNN.gz`** and collapses a same-seq raw+`.gz` pair (the transient mid-rotation window) to a single segment, **preferring the raw copy** (original bytes, skips a decompress; the two are byte-identical). In-flight `.gz.tmp` files are ignored by the all-digit subsuffix check.
- **Retention accounting** uses the on-disk (compressed) size, so the byte budget bounds actual disk use.
- **Premise note:** the issue's "Suggested slicing" was correct as written; #296 landed via PR #303 (rotation + retention + the reader reassembly this extends). No premise rot — implemented as specified.

## Test plan

- [x] `make sbsh-sb` — canonical build; `file ./sbsh` confirms ELF executable
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — full suite green (incl. `e2e`, `pkg/terminal/server`)
- [x] pre-commit (golangci-lint, go fmt, addlicense) — passed
- [x] New tests: `CompressSegment` round-trip (compressed-then-read == original), reader spanning gzip-closed + raw-live segments (`ReadAll` and `Dump`), same-seq raw-wins-over-gz collapse, updated rotation/retention tests to assert compression + verify content through the decompressing reassembler

## Acceptance criteria

- [x] Closed segments are gzip-compressed to `<canonical>.NNNNNN.gz` on rotation; the live segment remains raw and seekable.
- [x] `--full-capture`, `sb read`, and the parser feed transparently decompress `.gz` closed segments during reassembly, dispatching on file extension.
- [x] Tests cover a reader spanning raw-live + gzip-closed segments, and round-trip integrity (compressed-then-read == original bytes).

Closes #297

### DIFF
--- a/internal/capture/capture.go
+++ b/internal/capture/capture.go
@@ -25,19 +25,27 @@
 //     live, append-only, raw segment — the current tail of the transcript.
 //   - Closed segments are deterministic siblings named "<canonical>.NNNNNN",
 //     where NNNNNN is a zero-padded, monotonically increasing sequence number.
-//     Lower sequence numbers are older.
+//     Lower sequence numbers are older. On rotation a closed segment is gzip-
+//     compressed to "<canonical>.NNNNNN.gz" to stretch the retention budget;
+//     the live segment always stays raw and seekable.
 //
 // Contract: a raw `cat` of the canonical path shows the live tail only.
 // Reassembling the full history is the job of the readers here (used by
 // `sb read`, attach `--full-capture`, and any other consumer of full
 // transcript): ordered closed segments oldest-first, then the live segment.
+// Readers transparently decompress ".gz" closed segments, dispatching on file
+// extension, so a capture spanning gzip-closed and raw-live segments reads
+// back as the original byte stream.
 //
-// The segment matcher requires an all-digit suffix, so unrelated siblings
-// (e.g. a future compressed "<canonical>.NNNNNN.gz") are ignored rather than
-// spliced into the stream.
+// The segment matcher accepts both the raw "<canonical>.NNNNNN" form and the
+// compressed "<canonical>.NNNNNN.gz" form (collapsing the two to one segment
+// when both transiently coexist mid-rotation, preferring the raw copy).
+// Other siblings (a non-digit suffix, an in-flight ".gz.tmp") are ignored
+// rather than spliced into the stream.
 package capture
 
 import (
+	"compress/gzip"
 	"fmt"
 	"io"
 	"os"
@@ -52,6 +60,12 @@ import (
 // so a sequence that outgrows the width still sorts correctly.
 const seqWidth = 6
 
+// GzSuffix is appended to a closed segment's path once it has been gzip-
+// compressed: "<canonical>.NNNNNN.gz". Compression is sequential-read only
+// (readers io.Copy each segment end-to-end, never seeking), so gzip's lack of
+// a seek index is a non-issue and it adds no dependency beyond the stdlib.
+const GzSuffix = ".gz"
+
 // SegmentPath returns the deterministic closed-segment path for seq, a sibling
 // of canonical: "<canonical>.NNNNNN".
 func SegmentPath(canonical string, seq uint64) string {
@@ -62,7 +76,12 @@ func SegmentPath(canonical string, seq uint64) string {
 type Segment struct {
 	Path string
 	Seq  uint64
+	// Size is the on-disk size: for a compressed segment this is the gzip
+	// size, so byte-based retention bounds actual disk use.
 	Size int64
+	// Compressed is true when Path is a ".gz" segment that readers must
+	// decompress.
+	Compressed bool
 }
 
 // ClosedSegments returns the closed segments siblings of canonical, ordered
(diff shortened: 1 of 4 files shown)


eminwux:
**PR #314 Review — `feat(capture): gzip-compress closed capture segments on rotation`**

Solid, well-documented change. Compression is best-effort with the right ordering (temp → atomic rename → remove raw), the raw-wins same-seq collapse is correct under any `ReadDir` order, retention accounts for the on-disk compressed size, `NextSeq`/`prune` are `.gz`-aware, and the new compressed file is routed through the existing `applyPerms` mechanism rather than a second permission path. New symbols all have callers; commits verified; CI green.

One non-blocking note:

- **Latent race** `internal/capture/capture.go:198`: a closed segment listed by `OrderedPaths` as raw (`.NNNNNN`) can be removed by a concurrent `CompressSegment` (raw→`.gz`) between listing and read; `readSegment` then hits `os.IsNotExist` and the skip at line 198 silently drops a *still-retained* segment (its `.gz` exists but isn't re-resolved), leaving a transient gap in the reassembled transcript. Window is microseconds and self-heals on the next read, so non-blocking — but unlike the accepted prune/rotation skips, this drops retained history. Fix: on ENOENT in `readSegment`, retry the sibling path (`path+GzSuffix`, or strip it).

LGTM to proceed — the note above is non-blocking.

PR #315, opened 2026-05-21
fix(terminalrunner): restore backpressure for interactive attachers on high-throughput output

### LINKED ISSUE #312: terminal: restore backpressure for interactive attachers on high-throughput output
## Context

Running a high-throughput command inside an attached session disconnects the client:

```
$ cat /dev/urandom | hexdump
[sbsh: subscriber lagged, disconnecting]
Error: child routine exited: close requested: peer closed connection: read/write routines exited
```

This worked before #217. The shell's output used to be *flow-controlled* to the attached terminal's render speed — slow but never dropped, exactly how a real PTY paces a process to its terminal.

**Root cause.** #217 (`f6b3f4f`, "bound interactive attacher writer to unblock fan-out") replaced the interactive attacher's `os.Pipe` (a blocking writer — a full pipe blocks `Write`, propagating backpressure up through the PTY reader to the shell) with a bounded 1 MiB in-memory ring that **drops on overflow**. On overflow `subscriberWriter.Write` flips `lagged=true; closed=true` (`internal/terminal/terminalrunner/subscriber.go:89-95`) and the drain goroutine emits the lagged sentinel and closes the conn.

The interactive attach path (`connections.go:132`) now shares the **same** drop-on-lag `subscriberWriter` as the passive `Subscribe`/`sb read` path (`subscribe.go:124`). Drop-on-lag is correct for a passive observer that must never stall the session — but wrong for the primary interactive terminal, whose screen *is* the session's output device and should be paced, not dropped.

## Proposal

Split the fan-out policy by consumer class:

- **Interactive attachers** (`connections.go` `handleClient`): apply blocking **backpressure** — when the ring is full, the write path waits (the writer already has a `sync.Cond`, `subscriber.go:72`) until the drain frees space, instead of marking the attacher lagged+closed. This throttles the producer to terminal speed (the pre-#217 behavior).
- **Passive observers** (`subscribe.go` `Subscribe` / `sb read`, the capture writer, the vt-parser screen model): keep **drop-on-lag** unchanged — they must never stall the session.
- **Preserve the #217 invariant.** A paused or abandoned attacher must still not head-of-line-block the shared fan-out (which would freeze sibling attachers and the capture sink). Naively making the shared writer's `Write` block reintroduces exactly the problem #217 fixed, so the implementation must pace the *producer* on behalf of the live interactive attacher without blocking the capture/passive sinks (e.g. per-attacher paced drain; backpressure the producer only on the primary/slowest interactive terminal).

## Acceptance criteria

- [ ] An attached client running a sustained firehose (`cat /dev/urandom | hexdump`, `yes`) is throttled, not disconnected — no lagged sentinel for an interactive attach that is actively reading.
- [ ] Passive `Subscribe`/`sb read` consumers still receive the lagged sentinel and disconnect on overflow (behavior unchanged).
- [ ] A paused/abandoned interactive attacher does not stall sibling attachers or the capture file — the #217 invariant still holds.
- [ ] Regression test: a slow interactive attacher consuming a high-volume stream completes without a lagged-disconnect, locking in the firehose case.
- [ ] A design-rationale comment at the policy site (`subscriber.go` / `connections.go`) records why interactive attach uses backpressure while passive subscribe uses drop-on-lag, citing this regression and #217, so the policy is not silently re-merged into one drop-on-lag path again.

## Notes

- Regressed by #217 (`f6b3f4f`).
- Shared writer constructed at `subscriber.go:62` (`newSubscriberWriter`); interactive use at `connections.go:132`, passive use at `subscribe.go:124`.
- Companion documentation issue forthcoming (architecture flow-control policy in `docs/site/concepts/multi-attach.md`); link bidirectionally once filed.

eminwux:
## Summary

- Interactive attach (`connections.go`) and passive `Subscribe`/`sb read` (`subscribe.go`) shared one **drop-on-lag** `subscriberWriter`. #217 (`f6b3f4f`) collapsed them onto that single path, regressing the firehose case: an attached `cat /dev/urandom | hexdump` overflows the 1 MiB ring and the client is disconnected with the lagged sentinel instead of being flow-controlled to terminal speed.
- This splits the fan-out policy **by consumer class** via a per-writer `backpressure` flag set at the policy site (`enableBackpressure`, called only from the interactive path):
  - **Interactive attacher** → on ring overflow, `Write` paces the producer (PTY reader) until the drain frees space — flow-control, the pre-#217 PTY behavior — instead of dropping the attacher.
  - **Passive observers** (Subscribe, `sb read`, capture writer, vt-parser screen) → keep **drop-on-lag** unchanged (no `enableBackpressure`), so a slow observer can never stall the session.
- **#217 invariant preserved.** The backpressure pacing is bounded by a new `subscriberBackpressureStall` budget (default 2s). A *live* reader frees ring space within microseconds, so its overflowing writes return well inside the budget (throttle). A *paused/abandoned* attacher frees nothing, so its `Write` trips the budget once, drops the attacher on lag, and frees the producer — a bounded transient, never an indefinite head-of-line block of the capture sink or sibling attachers. The drain also `Broadcast`s on its write-deadline close so a parked producer unblocks immediately rather than waiting out the budget.

### Design notes for the reviewer
- **Why backpressure runs through the single PTY reader:** the only way to flow-control the shell is to stop reading the ptmx, which inherently paces *all* sinks to terminal speed. That's intended PTY semantics (the session paces to its terminal). The #217 invariant is specifically about a *stuck* attacher not *permanently* blocking siblings/capture — satisfied here because a non-draining attacher is ejected after `subscriberBackpressureStall`.
- **`subscriberBackpressureStall` carries a `//nolint:gochecknoglobals`.** It mirrors the existing test-tunable `subscriberWriteTimeout` package var; that neighbor is grandfathered by pre-commit's new-issues-only golangci filter, but a *new* global trips the gate, so an explicit suppression (matching the repo's existing `//nolint:gochecknoglobals` convention) is required. Open to making it a struct field instead if preferred.
- AC5's design-rationale comment lives on the `subscriberWriter` type doc plus the two call sites (`connections.go`, `subscribe.go`), citing #217 and #312 so the policy isn't silently re-merged.

## Test plan
- [x] `make test` (unit + integration + e2e) — pass
- [x] `make test-race` (race detector across all packages) — pass, no data races
- [x] `make sbsh-sb` + `file ./sbsh` — ELF executable
- [x] `go build ./...`, `go vet ./internal/terminal/terminalrunner/...` — clean
- [x] `pre-commit run --files <changed>` — pass
- [x] New regression tests (AC4):
  - `TestSubscriber_BackpressureThrottlesSlowInteractiveReader` — slow-but-live interactive reader consuming 8× the ring bound completes without a lagged-disconnect, full stream delivered.
  - `TestSubscriber_BackpressureStuckAttacherDropsAndFreesProducer` — a non-reading backpressure attacher is dropped after the stall budget and the producer is freed (#217 invariant).
- [x] AC2 (passive drop-on-lag unchanged) covered by existing `TestSubscriber_LaggedReaderIsDropped` / `TestSubscriber_LaggedNoticeReachesClient`, which use the default (non-backpressure) policy.

## Acceptance criteria
- [x] Attached firehose is throttled, not disconnected — no lagged sentinel for an actively-reading interactive attach.
- [x] Passive `Subscribe`/`sb read` consumers still get the lagged sentinel on overflow (unchanged).
- [x] A paused/abandoned interactive attacher does not stall siblings or capture (#217 invariant holds).
- [x] Regression test locking in the firehose case.
- [x] Design-rationale comment at the policy site citing this regression and #217.

Closes #312

### DIFF
--- a/internal/terminal/terminalrunner/connections.go
+++ b/internal/terminal/terminalrunner/connections.go
@@ -129,7 +129,18 @@ func (sr *Exec) handleClient(client *ioClient) {
 	// Write). Adding the writer to multiOutW *before* starting Run is safe —
 	// Write just enqueues into the ring until Run starts draining, and the
 	// seeded paint already sits ahead of any live bytes that arrive after Add.
+	//
+	// Backpressure policy (issue #312): the interactive attacher is the
+	// session's primary terminal — its screen *is* the output device — so on
+	// ring overflow the producer is paced to this attacher's drain rate
+	// (flow-control, the pre-#217 PTY behavior) rather than dropping it with a
+	// lagged sentinel. enableBackpressure bounds that pacing so a paused or
+	// abandoned attacher still falls through to drop-on-lag and never
+	// head-of-line-blocks the capture sink or sibling attachers (the #217
+	// invariant). Passive Subscribe / `sb read` consumers (subscribe.go)
+	// deliberately stay on the unbounded drop-on-lag path — see subscriber.go.
 	aw := newSubscriberWriter(client.conn, defaultSubscriberBufferBytes, detach, sr.logger)
+	aw.enableBackpressure()
 	aw.seedReplay(initial)
 	client.outWriter = aw
 	multiOutW.Add(aw)
--- a/internal/terminal/terminalrunner/subscribe.go
+++ b/internal/terminal/terminalrunner/subscribe.go
@@ -121,6 +121,12 @@ func (sr *Exec) Subscribe(req *api.SubscribeRequest, response *api.ResponseWithF
 		})
 	}
 
+	// Passive observer: keep the drop-on-lag policy (no enableBackpressure).
+	// A Subscribe / `sb read` consumer that falls behind must be disconnected
+	// with the lagged sentinel rather than pacing the producer, so it can
+	// never stall the session for the interactive attacher. The interactive
+	// attach path (connections.go) is the one that opts into backpressure;
+	// the policy split is documented on subscriberWriter. See issue #312.
 	sub = newSubscriberWriter(conn, defaultSubscriberBufferBytes, detach, sr.logger)
 	sr.addSubscriber(sub)
 	multiOutW.Add(sub)
--- a/internal/terminal/terminalrunner/subscriber.go
+++ b/internal/terminal/terminalrunner/subscriber.go
@@ -36,6 +36,21 @@ var subscriberWriteTimeout = 5 * time.Second
 // abandoned subscriber must never be able to stall the shared fan-out.
 const defaultSubscriberBufferBytes = 1 << 20 // 1 MiB
 
+// subscriberBackpressureStall bounds how long a single overflowing Write may
+// pace the producer on behalf of a backpressure-mode (interactive) attacher
+// before giving up and dropping the attacher on lag. A *live* interactive
+// reader frees ring space within microseconds, so its overflowing Writes
+// return well inside this budget and the producer is throttled to terminal
+// speed (issue #312). A *paused or abandoned* attacher frees nothing, so its
+// Write trips the budget once, drops the attacher, and frees the producer —
+// the bounded transient that keeps the #217 invariant intact (a stuck
+// attacher must not head-of-line-block the capture sink and sibling
+// attachers). Generous enough not to false-drop a momentarily slow live
+// terminal; var (not const) so tests can shorten it.
+//
+//nolint:gochecknoglobals // test-tunable knob, mirrors subscriberWriteTimeout
+var subscriberBackpressureStall = 2 * time.Second
+
 // laggedNotice is appended to a subscriber stream when the ring overflows
 // so a caller can distinguish disconnection from a clean EOF. The \r\n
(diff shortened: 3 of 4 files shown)


eminwux:
**PR #315 Review — `fix(terminalrunner): restore backpressure for interactive attachers on high-throughput output`**

LGTM — splits the fan-out policy by consumer class so the interactive attacher flow-controls the PTY (pre-#217 behavior) while passive observers keep drop-on-lag. The concurrency is sound: `Signal()`→`Broadcast()` correctly handles the two waiter classes now sharing the cond (drain in `Run`, producer in `waitForRoom`); `waitForRoom` holds `mu` across `cond.Wait()` and loops on its predicate; the `AfterFunc` timer only touches `mu` while the producer is parked, so no self-deadlock. The #217 invariant holds — a stuck attacher trips the budget once, drops, frees the producer, and subsequent writes no-op on `closed`. Commits verified, all new symbols referenced, both regression tests exercise the throttle and stuck-attacher paths.

PR #317, opened 2026-05-21
fix(capture): re-resolve raw/.gz sibling on ENOENT in segment readers

### LINKED ISSUE #316: capture: re-resolve raw→.gz sibling on ENOENT in readSegment
Surfaced during review of PR #314.

`internal/capture/capture.go:198` (the `os.IsNotExist` skip in `ReadAll`; same shape in `dumpOne`): a closed segment listed by `OrderedPaths` as raw (`.NNNNNN`) can be removed by a concurrent `CompressSegment` (raw→`.gz`) in the window between listing and read. `readSegment` then returns `os.IsNotExist`, and the skip silently drops a **still-retained** segment — its `.gz` copy is on disk but the stale raw path isn't re-resolved — leaving a transient gap in the reassembled transcript.

The window is microseconds and self-heals on the next read, so it was judged non-blocking for #314. Unlike the accepted prune/rotation skips (which drop genuinely-gone or live-tail data), this one omits retained history, so it's a latent correctness risk worth tightening.

Suggested fix: on ENOENT in `readSegment`, retry the sibling path — try `path+GzSuffix` when the raw is gone (and strip `GzSuffix` for the inverse), reading whichever representation currently exists before giving up and skipping.

eminwux:
## Summary

- `OrderedPaths` (via `ClosedSegments`) can list a closed segment under its raw `.NNNNNN` name an instant before a concurrent `CompressSegment` swaps it to `.NNNNNN.gz` (raw→`.gz` rename + raw removal). When `readSegment`/`dumpOne` then `os.Open` the now-vanished raw path, the old code returned `os.IsNotExist` and the caller **silently dropped a still-retained segment** — its `.gz` copy is on disk, so the transcript loses a chunk of retained history until the next read re-lists. Surfaced during review of PR #314.
- Fix: route both `readSegment` (used by `ReadAll`) and `dumpOne` (used by `Dump`) through a new `openSegment` helper. On `ENOENT` for the listed path it re-resolves the **sibling representation** — `path+GzSuffix` when the raw is gone, `strings.TrimSuffix(path, GzSuffix)` for the inverse — and reads whichever currently exists, dispatching gzip-decompression on the path actually opened (not the one requested).
- Only when *neither* representation exists does `openSegment` surface the original `os.IsNotExist`, so the legitimate skip for a genuinely pruned/rotated segment is preserved unchanged.

### Design notes for the reviewer
- The retry is symmetric (raw↔`.gz`) per the issue's "strip GzSuffix for the inverse" request. In practice `CompressSegment` only goes raw→`.gz`, so the inverse arm is defensive — a pruned `.gz` whose raw never reappears still falls through to the skip.
- `openSegment` returns the original `ENOENT` (on the listed path), not the sibling's, when both are gone — so `os.IsNotExist` stays true for the caller's skip branch and a real prune is never masked as a different error.
- The race window can't be reproduced through the public `ReadAll`/`Dump` (they re-list and find the `.gz`), so the regression tests exercise `readSegment`/`dumpOne` directly with the exact post-`CompressSegment` on-disk state (raw gone, `.gz` present).

## Test plan
- [x] `make test` (unit + integration + e2e) — pass
- [x] `go build ./...`, `go vet ./internal/capture/...` — clean
- [x] `go test ./internal/capture/... -race` — pass
- [x] `make sbsh-sb` + `file ./sbsh` — ELF executable
- [x] `pre-commit run --files <changed>` — pass
- [x] New regression tests:
  - `TestReadSegment_ResolvesGzSiblingOnRawENOENT` — raw listed, compressed away, recovered from `.gz`.
  - `TestDumpOne_ResolvesGzSiblingOnRawENOENT` — same for the `Dump` path.
  - `TestReadSegment_ResolvesRawSiblingOnGzENOENT` — inverse (strip `.gz`).
  - `TestReadSegment_BothGoneReturnsIsNotExist` — genuine prune still surfaces `os.IsNotExist` (skip preserved).

Closes #316

### DIFF
--- a/internal/capture/capture.go
+++ b/internal/capture/capture.go
@@ -206,19 +206,59 @@ func ReadAll(canonical string) ([]byte, error) {
 	return out, nil
 }
 
-// readSegment reads one segment fully into memory, decompressing transparently
-// when the path is a ".gz" closed segment. An absent path returns an
-// os.IsNotExist error so ReadAll can skip a segment pruned between listing and
-// read.
-func readSegment(path string) ([]byte, error) {
+// siblingPath returns the alternate on-disk representation of a closed
+// segment: the ".gz" form for a raw path, or the raw form for a ".gz" path.
+func siblingPath(path string) string {
+	if strings.HasSuffix(path, GzSuffix) {
+		return strings.TrimSuffix(path, GzSuffix)
+	}
+	return path + GzSuffix
+}
+
+// openSegment opens a closed segment for reading, recovering from the raw↔".gz"
+// rename race: a segment listed by OrderedPaths as raw can be swapped to ".gz"
+// by a concurrent CompressSegment (or vice-versa) in the microsecond window
+// before the reader opens it. On ENOENT for the listed path we re-resolve the
+// sibling representation — the same retained segment under its other name —
+// rather than dropping still-retained history (issue #316). It returns the
+// opened file and whether that file is gzip-compressed (dispatch follows the
+// path actually opened, not the one requested). Only when *neither*
+// representation exists is the original os.IsNotExist returned, so callers
+// keep skipping a segment genuinely pruned or rotated away.
+func openSegment(path string) (*os.File, bool, error) {
 	f, err := os.Open(path)
+	if err == nil {
+		return f, strings.HasSuffix(path, GzSuffix), nil
+	}
+	if !os.IsNotExist(err) {
+		return nil, false, err
+	}
+	sib := siblingPath(path)
+	sf, serr := os.Open(sib)
+	if serr != nil {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #317 Review — `fix(capture): re-resolve raw/.gz sibling on ENOENT in segment readers`**

LGTM — `openSegment` re-resolves the raw↔`.gz` sibling on ENOENT in both `readSegment` and `dumpOne` (closes the #316 race), dispatching gzip on the path actually opened rather than the one requested; the symmetric retry returns the *original* ENOENT only when neither representation exists, so the genuine-prune skip is preserved, and the four new direct-reader tests cover both directions plus the both-gone case.

Issue #304, opened 2026-05-21
test: umbrella — raise test coverage across all packages to ≥80%


eminwux:
## Context

Test coverage is uneven across the codebase. A baseline `go test -cover ./...` run (2026-05-21) shows packages ranging from 0% (no tests at all) to 100%, with several critical runtime packages under 40%. The single biggest risk is `internal/client/clientrunner` at **10.1%** across 11 source files — the client runtime is almost entirely untested. Coverage gaps in runtime, RPC, and CLI layers mean regressions in those paths ship undetected.

This umbrella tracks a phased effort to review existing coverage and add tests until every package reaches **≥80% statement coverage**. Each phase is one cohesive layer, scoped to land as a reviewable PR (or a small set of PRs where a layer is large).

## Baseline coverage (2026-05-21)

Untested (0% / no tests): `cmd/parser`, `cmd/types`, `cmd/sb/{read,screenshot,version,autocomplete}`, `cmd/sbsh/{version,autocomplete}`, `internal/{dualcopier,filter,logging,naming,shared,defaults,errdefs}`, `pkg/rpcclient/terminal`.

Low (<40%): `internal/client/clientrunner` (10.1%), `cmd/config` (18%), `internal/discovery` (23%), `cmd/sbsh` (27%), `internal/terminal/terminalrpc` (30%), `cmd/sb/prune` (35%), `cmd/sb/get` (37%), `pkg/api` (37.5%), `internal/terminal` (40%).

Already strong (>80%): `pkg/builder` (97%), `pkg/logging` (93%), `internal/initmode` (100%) — out of scope, listed for reference.

## Phases

- [ ] #305 — Phase 1: Leaf utility packages
- [ ] #306 — Phase 2: Public `pkg/` API surface
- [ ] #307 — Phase 3: Client runtime
- [ ] #308 — Phase 4: Terminal runtime
- [ ] #309 — Phase 5: Supporting internals
- [ ] #310 — Phase 6: CLI command layer
- [ ] #311 — Phase 7: e2e harness fix + integration coverage

## Acceptance criteria

- [ ] Every non-trivial package reaches ≥80% statement coverage (packages with no executable statements, e.g. `pkg/errors`, are exempt).
- [ ] All seven phase issues are closed via merged PRs.
- [ ] `go test ./...` passes with no skipped-due-to-missing-binary failures (see Phase 7 / #311).

## Notes

[...]

PR #319, opened 2026-05-21
test: raise pkg/ public API coverage above 80%

### LINKED ISSUE #306: test: phase 2 — public pkg/ API surface (≥80%)
Part of #304 (umbrella).

## Context

The `pkg/` tree is the project's public API surface — the contract external consumers depend on. Two of its packages are under-tested: `pkg/api` (37.5%) and `pkg/rpcclient/terminal` (0%). Regressions here break downstream callers silently.

## Proposal

Raise the public API packages to ≥80% statement coverage:

- `pkg/api` (37.5% → ≥80%)
- `pkg/rpcclient/terminal` (0% → ≥80%)
- `pkg/rpcclient/client` (67.5% → ≥80%)

## Acceptance criteria

- [ ] Each listed package reaches ≥80% statement coverage.
- [ ] Tests exercise request/response marshalling, error returns, and boundary inputs on the RPC client surface.
- [ ] `go test ./...` and `go vet ./...` pass.

## Notes

- `pkg/builder` (97%) and `pkg/logging` (93%) already exceed the bar — out of scope.

eminwux:
## Summary

Phase 2 of the `pkg/` coverage push (#304 umbrella). Raises the three under-tested public API packages above the ≥80% statement-coverage bar with no production-code changes — tests only.

| Package | Before | After |
| --- | --- | --- |
| `pkg/api` | 37.5% | **81.2%** |
| `pkg/rpcclient/terminal` | 0% | **88.8%** |
| `pkg/rpcclient/client` | 67.5% | **100%** |

What the new tests exercise:

- **`pkg/api`** — adds `Test_ClientStatusMode_String` covering every `ClientStatusMode` value plus the `Unknown` default, mirroring the existing `TerminalStatusMode` table test. (The package is almost entirely type declarations; the two `String()` methods are its only executable code.)
- **`pkg/rpcclient/terminal`** — new internal test suite stands up a real `net/rpc` server backed by `internal/terminal.ControllerTest` and the production `terminalrpc.NewUnixJSONServerCodec`, so request/response marshalling rides the real wire format. Covers Ping (roundtrip + error), Resize, Detach, Metadata/State/Screenshot (incl. error propagation), Write (incl. nil-req substitution), and the FD-passing **Attach/Subscribe** paths end-to-end — passing a real socketpair fd over SCM_RIGHTS and reading bytes back through the returned `net.Conn`, plus the no-FD and error branches. Also covers `callWithCodec`'s dial-failure retry exhaustion, context cancellation during the retry delay and during the in-flight call, and `WithDialTimeout`.
- **`pkg/rpcclient/client`** — adds the defensive nil-argument substitution paths (Ping/Metadata/State/Stop), dial failure against a missing socket, and context cancellation against a blocking handler, taking it to 100%.

## Non-obvious implementation notes (for reviewer)

- The `terminal.call()` non-`*net.UnixConn` branch (the generic `LoggingConn` path) is unreachable through the public `NewUnix` constructor, which always builds a unix dialer. I cover it via an in-memory `net.Pipe` whose conns are not `*net.UnixConn`. I first tried a real TCP loopback jsonrpc server, but **TCP loopback hangs in this sandbox** (even a bare `net/rpc` jsonrpc roundtrip with no project code stalled ~130s), so `net.Pipe` is both more hermetic and avoids the environmental flake.
- The terminal tests are an internal `package terminal` test (not `terminal_test`) so they can construct `&client{...}` with custom dialers to drive the retry/cancellation branches directly. They carry `//go:build unix` to match the package's `ReadMsgUnix`/SCM_RIGHTS reliance, consistent with the existing client-package test.

## Test plan

- [x] `make sbsh-sb` — ELF executable verified via `file ./sbsh`
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full CI test target, all pass
- [x] `go test -cover ./pkg/api/ ./pkg/rpcclient/...` — confirms each package ≥80% (81.2% / 88.8% / 100%)
- [x] pre-commit (golangci-lint et al.) on changed files

Closes #306

### DIFF
--- a/pkg/api/client_test.go
+++ b/pkg/api/client_test.go
@@ -1,0 +1,42 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package api
+
+import "testing"
+
+func Test_ClientStatusMode_String(t *testing.T) {
+	tests := []struct {
+		name     string
+		state    ClientStatusMode
+		expected string
+	}{
+		{name: "Initializing", state: ClientInitializing, expected: "Initializing"},
+		{name: "Ready", state: ClientReady, expected: "Ready"},
+		{name: "Attached", state: ClientAttached, expected: "Attached"},
+		{name: "Exiting", state: ClientExiting, expected: "Exiting"},
+		{name: "Exited", state: ClientExited, expected: "Exited"},
+		{name: "Unknown", state: ClientStatusMode(999), expected: "Unknown"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := tt.state.String(); got != tt.expected {
+				t.Errorf("ClientStatusMode.String() = %v, want %v", got, tt.expected)
+			}
+		})
+	}
+}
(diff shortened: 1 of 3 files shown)


eminwux:
**PR #319 Review — `test: raise pkg/ public API coverage above 80%`**

LGTM — tests-only coverage push (766 additions, 0 production-code changes); real `net/rpc` server + SCM_RIGHTS FD-passing exercise the live wire format, every test asserts meaningfully, no dead helpers, CI green including E2E.

PR #320, opened 2026-05-21
feat(capture): add asciicast capture format selector

### LINKED ISSUE #298: terminal: asciicast capture format (replay/seek/portability)
## Context

Capture is always-on **raw PTY bytes**, written by the single `captureWriter` sink (`internal/terminal/terminalrunner/capture_writer.go`), append-only into the live segment. Raw carries no timing or structure, so it can't be replayed at speed, seeked by time, or handed to standard tooling (asciinema player, `agg` for gif export).

**Correction to the prior body:** there is no capture *format* selector today. `Spec.CaptureMode` (`pkg/api/terminal.go:104`) is the **chmod** applied to the capture file — a permission bitmask, not a format enum. This issue introduces a new selector field.

Format choice is orthogonal to screen-state rendering: asciicast is an output-event stream, not a screen snapshot. Rendering a frame still goes through the vt-parser screen model (#295, shipped). Asciicast's value is replay-at-speed, seek-by-time, and portability.

## Proposal

- Add a new `Spec.CaptureFormat` field (named after the capability it gates: the on-disk format), values `raw` (default) and `asciicast`. **Mutually exclusive per terminal** — one capture file, one format. Default stays raw, so existing behavior and the zero-overhead path are unchanged.
- In asciicast mode the writer emits asciicast v2: a JSON header line, then one timed `[t,"o",data]` record **per `Write` call** (whole-event writes, so rotation never bisects a record line).
- All readers funnel through `capture.ReadAll` / `capture.Dump` (`internal/capture/capture.go`), which already reassemble gzip-closed + raw-live segments. They gain a format **sniff** (asciicast header vs raw) and, for asciicast, a **decode step on the reassembled stream**:
  - `--full-capture` attach (`connections.go` → `readCaptureFile`): decode asciicast events → bytes before writing to the client PTY (raw is blasted verbatim, as today).
  - `sb read` and the parser feed: same sniff-and-decode.
- Default repaint attach is unaffected — it reads the live vt-parser screen model, never the capture file, so it is format-independent.

## Dependencies

The capture-rendering arc this composes with is **fully shipped** — the prior body's "independent / coordinate at whichever lands second" framing is stale:

- #295 — vt-parser screen model + Screenshot RPC (closed)
- #71 — repaint on attach + `--full-capture` (closed)
- #296 / PR #303 — capture rotation + retention (closed)
- #297 / PR #314 — gzip-compress closed segments (merged)

Concretely, the reassembled byte stream for an asciicast capture is: decompress `.gz` closed segments + raw live segment (already done by `capture.ReadAll`), concatenated, **then** asciicast-decoded by the reader. Asciicast integrates with this existing pipeline; it does not coordinate with in-flight work.

## Tricky bits

- **Header survival under pruning (new failure mode).** An asciicast file is one header line + event lines; the header is written once at the top of segment 0. Reassembly concatenates segments, so the stream is valid asciicast only while segment 0 survives. Retention prunes oldest closed segments by count (8) and bytes (64 MiB), so on a long session segment 0 is eventually dropped → headerless event lines = invalid asciicast. Raw has no header and is immune. Resolve by one of: re-emit the header at the top of every segment (each segment self-describing — composes cleanly with #314's per-segment gzip), exempt segment 0 from pruning, or synthesize a header in the reader from spec metadata.
- **Whole-event writes.** Reassembly only parses if no rotation splits an event line. The writer already guarantees "a single write is never split across segments"; asciicast must therefore emit exactly one record per `Write`.
- **Timestamps survive concat.** asciicast v2 timestamps are absolute offsets from session start, so events concatenated across segments need no rebasing.
- **Byte-fidelity tradeoff.** Raw is byte-exact arbitrary bytes; asciicast wraps `data` in JSON strings that assume valid UTF-8, so non-UTF-8 PTY output may not round-trip. Asciicast is chosen for replay/portability, not byte-exact archival (see non-goals).

## Non-goals

- Screen-state rendering / fidelity-on-attach — that's the vt-parser model (#295, shipped). Asciicast does not provide a current-screen snapshot.
- An additive "always raw **and** asciicast" dual-write. Asciicast is a superset of raw for replay (records carry the bytes + timing), so dual storage would double disk for no replay benefit; the only thing it buys is byte-exact non-UTF-8 fidelity, which is out of scope here. If byte-exact archival becomes a requirement, file a separate issue.

## Acceptance criteria

- [ ] A new `Spec.CaptureFormat` field selects `asciicast`; default remains `raw`. (Distinct from `CaptureMode`, which stays the chmod field.)
- [ ] In asciicast mode the live capture file is valid asciicast v2 (header + timed `o` events) playable by the asciinema player, with one record per write.
- [ ] Readers sniff format and decode asciicast or pass through raw correctly, operating on the reassembled (gzip-closed + raw-live) stream; the raw path is byte-for-byte unchanged.
- [ ] `--full-capture` attach replays an asciicast capture correctly (decodes events → PTY bytes), including across a rotation boundary.
- [ ] A reassembled asciicast capture remains valid after segment 0 is pruned by retention (header survives via the chosen mechanism).
- [ ] Tests cover: raw mode unchanged, asciicast mode produces a valid file, a reader handles both formats, reassembly spans a rotation boundary, and header-survives-pruning.

## Suggested slicing

Single PR for the writer + reader format support and the sniff/decode path — a format change touches the writer and all readers at once and resists clean splitting (a writer-only or readers-only PR leaves a broken intermediate). The header-survival mechanism ships in the **same** PR (without it, asciicast + rotation + retention produces invalid files). If dev finds it too large at pickup, flag `resize-issue`; the only clean cut would be deferring header-survival as a fast-follow *if* the first PR disables retention-pruning of segment 0 in the interim.

## Notes

Independent track within the (now fully shipped) capture-rendering arc. No longer blocked by anything.

eminwux:
## Summary

- Adds `Spec.CaptureFormat` (`raw` default | `asciicast`), the new per-terminal on-disk capture-format selector requested in #298. Raw stays the default and the zero-overhead path, so existing behavior and serialization are unchanged (the field is `omitempty` and is left empty unless explicitly set).
- In asciicast mode the capture writer emits **asciicast v2**: a JSON header line at the top of every segment, then exactly one timed `[t,"o",data]` record per `Write`. One-record-per-write means rotation (which checks size *after* the write) never bisects a record line — the whole-event-writes guarantee the decoder relies on.
- **Header survives pruning (the load-bearing new AC):** the header is re-emitted at the top of *every* segment (chosen over "exempt segment 0" / "synthesize in reader" because it composes cleanly with #314's per-segment gzip and keeps each segment self-describing). Once retention prunes segment 0, the new oldest segment still starts with a header, so the reassembled stream stays sniffable and decodable.
- Both readers funnel through a new `capture.Replay` (sniff + decode): `--full-capture` attach (`io.go`) and `sb read` (`read.go`). Raw passes through byte-for-byte; asciicast is decoded to its output bytes on the reassembled (gzip-closed + raw-live) stream, so a client PTY / parser feed sees terminal output either way. The decoder tolerates a header line anywhere (multi-segment reassembly), skips foreign event codes, and skips a torn tail rather than failing the whole replay.
- Wired end-to-end so the field is usable by both consumers: `builder.WithCaptureFormat` (embedder API), and a `--capture-format` / `--terminal-capture-format` CLI flag + env/viper binding mirroring `--capture-mode`. Invalid values are rejected at spec-build time via `api.NormalizeCaptureFormat` (verified: `invalid capture format "bogus": must be "raw" or "asciicast"`).

## Design notes for review

- **Geometry:** the asciicast header records the session's *initial* PTY geometry (VT100 80×24, the size `terminal.go` sets at launch). Resizes after launch do not rewrite the header — that is asciicast v2's model (the header is the starting size).
- **Byte-fidelity tradeoff (per the issue's non-goals):** asciicast wraps `data` in JSON strings (valid-UTF-8 assumption), so non-UTF-8 PTY output may not round-trip byte-exactly. Asciicast is chosen for replay/portability, not byte-exact archival; raw remains the byte-exact format. No additive dual-write (explicit non-goal in #298).
- **`sb read` buffering:** `dumpCapture` switched from streaming `capture.Dump` to `capture.Replay` (buffers the full transcript to sniff/decode). Total transcript size is bounded by retention (~64 MiB closed + 8 MiB live), so buffering once for a CLI read is acceptable. `capture.Dump` is retained (still has tests) for any future streaming raw consumer.
- **Deferred (not required by the AC, kept out to bound scope):** a profile-YAML `captureFormat` field. The CLI flag + builder option cover both consumers today; a profile-spec field can follow if profiles need to pin a format.

## Test plan

- [x] `make sbsh-sb` + ELF verify (via ELF magic; `file` not installed on host)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test` over all non-e2e packages
- [x] `make e2e` (E2E_BIN_DIR=$(pwd) go test ./e2e)
- [x] `docs/examples/library-consumer` build + vet (additive `WithCaptureFormat`)
- [x] `pre-commit run --files` (golangci-lint, go fmt, addlicense, …)
- [x] New unit tests: asciicast encode/sniff/decode round-trip, payload escaping, multi-header + foreign-event skip, torn-tail skip, `Replay` raw+asciicast, header-survives-pruning; writer-level asciicast live-file validity, one-record-per-write, rotation-with-per-segment-header + post-prune validity
- [x] Real-binary smoke: `sbsh terminal --capture-format asciicast …` produced a valid asciicast v2 file (`{"version":2,"width":80,"height":24,...}` + `[t,"o",...]` events); raw mode unchanged (no header, raw bytes); invalid format rejected at spec build

Acceptance criteria (#298):
- [x] `Spec.CaptureFormat` selects asciicast; default raw; distinct from `CaptureMode`
- [x] asciicast mode produces a valid asciicast v2 live file, one record per write
- [x] Readers sniff format and decode asciicast / pass raw through byte-for-byte, on the reassembled stream
- [x] `--full-capture` attach replays asciicast (decode → PTY bytes), incl. across a rotation boundary
- [x] Reassembled asciicast stays valid after segment 0 is pruned (per-segment header)
- [x] Tests cover raw-unchanged, asciicast-valid, reader-handles-both, reassembly-across-rotation, header-survives-pruning

Closes #298

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -170,6 +170,8 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_TERM_CAPTURE_GID = DefineKV("SBSH_ROOT_TERM_CAPTURE_GID", "sbsh.root.terminal.captureGid")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_ROOT_TERM_CAPTURE_FORMAT = DefineKV("SBSH_ROOT_TERM_CAPTURE_FORMAT", "sbsh.root.terminal.captureFormat")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_TERM_LOG_FILE_MODE = DefineKV("SBSH_ROOT_TERM_LOG_FILE_MODE", "sbsh.root.terminal.logFileMode")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_ROOT_TERM_LOG_FILE_GID = DefineKV("SBSH_ROOT_TERM_LOG_FILE_GID", "sbsh.root.terminal.logFileGid")
@@ -206,6 +208,8 @@ var (
 	SBSH_TERM_CAPTURE_MODE = DefineKV("SBSH_TERM_CAPTURE_MODE", "sbsh.terminal.captureMode")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_CAPTURE_GID = DefineKV("SBSH_TERM_CAPTURE_GID", "sbsh.terminal.captureGid")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_TERM_CAPTURE_FORMAT = DefineKV("SBSH_TERM_CAPTURE_FORMAT", "sbsh.terminal.captureFormat")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SBSH_TERM_LOG_FILE_MODE = DefineKV("SBSH_TERM_LOG_FILE_MODE", "sbsh.terminal.logFileMode")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/sb/read/read.go
+++ b/cmd/sb/read/read.go
@@ -151,12 +151,24 @@ func runRead(
 	return streamLive(conn, stdout, stderr)
 }
 
-// dumpCapture streams the full transcript to stdout. The recorded capture
-// path is the live segment; capture.Dump reassembles closed rotated segments
-// oldest-first, then the live segment, into one logical stream. An absent
-// capture (never written to) produces no output and no error.
+// dumpCapture writes the full transcript to stdout. The recorded capture path
+// is the live segment; capture.Replay reassembles closed rotated segments
+// oldest-first, then the live segment, into one logical stream and sniffs the
+// format — a raw capture passes through byte-for-byte, an asciicast capture is
+// decoded to its output bytes so `sb read` (and the parser feed behind it)
+// always sees terminal output, matching the live stream that follows. An
+// absent capture (never written to) produces no output and no error. Total
+// transcript size is bounded by capture retention, so buffering it to sniff
+// the format is acceptable for a CLI read.
 func dumpCapture(path string, stdout io.Writer) error {
-	return capture.Dump(path, stdout)
+	data, err := capture.Replay(path)
+	if err != nil {
+		return err
+	}
+	if _, werr := stdout.Write(data); werr != nil {
+		return fmt.Errorf("write capture transcript: %w", werr)
+	}
+	return nil
 }
 
 // laggedNotice mirrors the bracketed core of the server sentinel (see
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -170,6 +170,7 @@ You can also use sbsh with parameters. For example:
 					SocketGID:        readRootSocketGID(cmd),
 					CaptureMode:      viper.GetString(config.SBSH_ROOT_TERM_CAPTURE_MODE.ViperKey),
 					CaptureGID:       readRootCaptureGID(cmd),
+					CaptureFormat:    viper.GetString(config.SBSH_ROOT_TERM_CAPTURE_FORMAT.ViperKey),
 					LogFileMode:      viper.GetString(config.SBSH_ROOT_TERM_LOG_FILE_MODE.ViperKey),
 					LogFileGID:       readRootLogFileGID(cmd),
 					DisableSetPrompt: viper.GetBool(config.SBSH_ROOT_TERM_DISABLE_SET_PROMPT.ViperKey),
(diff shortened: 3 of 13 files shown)


eminwux:
**PR #320 Review — `feat(capture): add asciicast capture format selector`**

Clean, well-scoped feature. Raw stays the default zero-overhead path, the asciicast writer's one-record-per-write guarantee composes correctly with rotation/pruning, and the reader sniff-and-decode funnels both consumers through `capture.Replay`. Correctness, error wrapping (`%w`), and the per-segment-header survival mechanism all check out; no secret/PII surface; no dead code (`capture.Dump` is retained with live tests). Commits are signed; CI is green. Public-API parity is satisfied — `builder.WithCaptureFormat` threads the field through `paramsFromConfig` → params → spec alongside the CLI flag.

One non-blocking note:

- **Config-surface asymmetry** `internal/profile/profile.go:462`: `CaptureMode` is reachable via profile YAML (`spec.capture.mode`), but the new `CaptureFormat` is CLI/builder/env only — a profile-YAML user can't pin a format. Author documented this as a deliberate scope deferral and #298's AC doesn't require it, so non-blocking; filing a followup to track.

LGTM — ship it.

Issue #334, opened 2026-05-22
test: terminalrunner package suite drops a core dump into the working tree


eminwux:
## Repro
On `origin/main` (reproduced at `12318b0`):

```bash
cd internal/terminal/terminalrunner
rm -f core.*
go test -count=1 .          # all tests PASS
ls core.*                   # => one core.<pid> file (~446 KB ELF) appears
```

- Reproduces consistently on a clean `origin/main` checkout (verified by removing the new `lifecycle_pty_cluster_test.go` from PR #333 and re-running — the core still appears, so it is **not** introduced by that PR).
- Does **not** reproduce when running individual test families in isolation (`-run TestReaper`, `-run TestSignal`, `-run TestSignals`, `-run TestStartTerminalFullLifecycle`, `-run TestStartTerminalEmitsExitEvent` each produce 0 cores). Only the full-package run trips it — suggests a timing/ordering interaction between tests that spawn real child processes.

## Observed
- Exactly one `core.<pid>` (e.g. `core.19522`), 446464 bytes, ELF, written into `internal/terminal/terminalrunner/` (the test's working directory).
- The `go test` run reports **PASS** for every test — the crash is in a spawned **child process**, not the test binary, so it is invisible in normal CI output.
- `core.*` is **not** in `.gitignore`, so a stray dump shows up as an untracked file in `git status` and can be accidentally `git add`-ed.

## Root cause
Not yet pinpointed. Candidate shortlist (tests that fork real children and signal them, where a core-dumping signal under `ulimit -c unlimited` would land in the cwd):
- `lifecycle_shutdown_test.go` — `TestShutdownChild_*` send SIGTERM/SIGKILL to child process groups.
- `signals_linux_test.go` / `reaper_linux_test.go` — PID-1 init-mode signal forwarding and zombie reaping; a forwarded signal hitting a child at the wrong moment is a plausible source.
- The PTY-lifecycle tests (`lifecycle_pty_cluster_test.go`, post-#333) and any other test that starts `/bin/cat` / `/bin/sh` children and tears them down.

The interaction-only nature points at a signal delivered to a child during a window where its disposition produces a core dump (e.g. SIGQUIT/SIGABRT/SIGTRAP rather than the intended SIGTERM/SIGKILL), surfacing only when prior tests have left shared state (init-mode reaper, signal forwarder) active.

## Suggested fix
Two separable mitigations:
[...]

PR #347, opened 2026-05-22
test: cover stop terminal RPC-graceful, PID-reuse, and --all aggregation paths

### LINKED ISSUE #107: test: deeper coverage for `sb stop terminal` paths
Follow-up from PR #106 review. Current `cmd/sb/stop/terminals_test.go` covers only flag resolution and `processAlive(0/-1)`. The following paths should get dedicated tests, using the `ControllerTest.StopFunc` fake already wired in `controller_fake.go`:

- **RPC graceful path** — controller returns success on `Stop`, verify no signal is sent and metadata cleanup runs.
- **SIGTERM fallback** — RPC unreachable / dial timeout, process exits within `--timeout`, verify only SIGTERM is sent.
- **`--kill-after` escalation** — SIGTERM times out, verify SIGKILL is sent after the timeout and waited for no more than the fixed 2s grace (see `sigkillGrace` in `cmd/sb/stop/terminals.go`).
- **`--all` aggregation** — multiple terminals, mix of successes / `--ignore-not-found` hits / per-terminal errors; verify aggregated exit behavior.
- **PID-reuse rejection** — process is alive but `PidStart` doesn't match `pidutil.Read`; verify `processIsOurs` returns false and no signal is sent (complements the unit test already in `internal/pidutil`).

Context: raised as non-blocking in PR #106 review; the fix commit (8500b11) focused on correctness (PID-reuse race, double-timeout) and left broader stop-command coverage for a follow-up.

eminwux:
## Summary
- Adds dedicated tests for the `sb stop terminal` paths #107 spec'd, focusing on the three gaps not already covered by the phase-6 coverage work in `stop_test.go`.
- New `cmd/sb/stop/terminals_paths_test.go`:
  - **RPC graceful path** — stands up a real jsonrpc `TerminalController.Stop` server on a Unix socket; the recorded process traps SIGTERM so only the RPC handler's SIGKILL can reap it, proving the graceful path ran (no fallback signal).
  - **PID-reuse rejection** (at the `stopOneTerminal` level) — a live PID with a mismatched `PidStart` token is treated as already-exited and left unsignalled.
  - **`--all` aggregation** — multiple targets (one stopped, one failed, one dead-PID filtered by reconcile); asserts the summary counts and the aggregated `ErrStopTerminal`.
- `cmd/sb/stop` statement coverage 81.7% → **84.4%**; `runStopTerminals` 86.4% → 100.0%.

## Notes for reviewer (please push back if any oversteps)
- **Rotted issue premise, intent preserved.** The issue says to use "the `ControllerTest.StopFunc` fake already wired in `controller_fake.go`". That fake lives in `internal/terminal` and drives the *in-process* controller; `cmd/sb/stop` reaches the controller over a Unix socket via `pkg/rpcclient/terminal`, so it cannot exercise the command-layer graceful path. I reinterpreted the intent (cover the path) and stood up a real jsonrpc listener instead. No production code changed.
- **Two of the five spec'd paths were already covered** by prior phase-6 work and are intentionally not duplicated: the **SIGTERM fallback** (`Test_RunStopTerminals_GracefulSIGTERM`) and the **`--kill-after` escalation** (`Test_StopOneTerminal_KillAfterEscalates`), both in `cmd/sb/stop/stop_test.go`.
- **`alreadyExited` is unreachable via `--all`**: `reconcileTerminal` (`internal/discovery/reconcile.go:106`) marks dead/reused-PID terminals `Exited`, and the `--all` branch of `resolveTargets` (`cmd/sb/stop/terminals.go:215`) filters those out before aggregation, so the `--all` summary's already-exited count is always 0. The single-name already-exited path is covered by the existing `Test_RunStopTerminals_AlreadyExited`.
- Removed an unrelated `core.14297` core dump that the terminalrunner suite drops into the working tree (tracked separately as #334); not part of this diff.

## Test plan
- [x] `go test -count=1 ./cmd/sb/stop/` — pass (84.4% coverage)
- [x] `go build ./...`
- [x] `go vet ./cmd/sb/stop/`
- [x] `make sbsh-sb` — ELF verified (`7f 45 4c 46`), `sb` hardlink present
- [ ] `make test-race` — not run; `-race` needs cgo and no C compiler (`gcc`) is on the dev host. The CI `test` target does not use `-race` (that is the separate `test-race` target).

Closes #107

### DIFF
--- a/cmd/sb/stop/terminals_paths_test.go
+++ b/cmd/sb/stop/terminals_paths_test.go
@@ -1,0 +1,210 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package stop
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"net"
+	"net/rpc"
+	"net/rpc/jsonrpc"
+	"os/exec"
+	"sync/atomic"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/eminwux/sbsh/internal/errdefs"
+	"github.com/eminwux/sbsh/internal/pidutil"
+	"github.com/eminwux/sbsh/pkg/api"
+)
+
+// stopRPCService is a minimal net/rpc service exposing only TerminalController.Stop.
+// The command layer (cmd/sb/stop) talks to the controller over a Unix socket via
+// pkg/rpcclient/terminal, so the graceful path cannot be driven by the in-process
+// ControllerTest fake the issue referenced; it needs a real jsonrpc listener.
+type stopRPCService struct {
+	onStop func(args *api.StopArgs) error
+}
+
+func (s *stopRPCService) Stop(args *api.StopArgs, _ *api.Empty) error {
+	return s.onStop(args)
+}
+
+// startStopRPCServer stands up a jsonrpc Unix server at sockPath that registers a
+// TerminalController.Stop handler. The server is torn down via t.Cleanup.
+func startStopRPCServer(t *testing.T, sockPath string, onStop func(*api.StopArgs) error) {
+	t.Helper()
+	srv := rpc.NewServer()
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #347 Review — `test: cover stop terminal RPC-graceful, PID-reuse, and --all aggregation paths`**

LGTM — test-only (+210/-0), no production code touched. All referenced symbols exist on the head, commits verified, CI green. The RPC-graceful test's logic is sound: a `trap "" TERM` process paired with an RPC handler that only SIGKILLs means `resultStopped` genuinely proves the graceful path ran rather than the SIGTERM fallback. PID-reuse and `--all` aggregation paths are likewise well-targeted, and the `0x7fffffff` dead-PID choice is safely above any valid `pid_max`.

Non-blocking: `make test-race` is unchecked but documented as host-gated (no cgo/gcc on dev host; CI's `test` target doesn't run `-race` either). Worth a manual race run pre-merge since the new tests spawn goroutines (RPC accept loop, background waiters), but not a blocker.

PR #353, opened 2026-05-22
test: make terminal profile-fallback tests hermetic to SBSH_TERM_PROFILE

### LINKED ISSUE #350: test: cmd/sbsh/terminal profile tests are not hermetic to SBSH_TERM_PROFILE
## Repro
On clean `origin/main` (reproduced at `1d0ab74`), with `SBSH_TERM_PROFILE` exported in the environment — which is the case for any developer running the suite from *inside* an sbsh session (sbsh dogfoods itself):

```bash
export SBSH_TERM_PROFILE=sbsh-dev-1   # or any profile name not on disk
go test -count=1 ./cmd/sbsh/terminal/
```

```
--- FAIL: Test_buildTerminalSpecFromFlags_DefaultProfile (0.00s)
    terminal_test.go:690: buildTerminalSpecFromFlags() error = profile "sbsh-dev-1" not found in /tmp/.../002
--- FAIL: Test_processSpec_BuildsFromFlags (0.00s)
    terminal_test.go:715: processSpec() error = failed to build terminal spec from flags: profile "sbsh-dev-1" not found in /tmp/.../002
--- FAIL: Test_TerminalCmd_PreRunE_BuildsSpecFromFlags (0.00s)
    terminal_test.go:737: PreRunE() error = failed to build terminal spec from flags: profile "sbsh-dev-1" not found in /tmp/.../002
```

Passes cleanly with `env -u SBSH_TERM_PROFILE go test ./cmd/sbsh/terminal/`. CI is green because the CI container has no `SBSH_TERM_PROFILE` set.

## Observed
Three tests in `cmd/sbsh/terminal/terminal_test.go` (`Test_buildTerminalSpecFromFlags_DefaultProfile` ~L684, `Test_processSpec_BuildsFromFlags` ~L700, `Test_TerminalCmd_PreRunE_BuildsSpecFromFlags` ~L725) intend to exercise the *default-profile fallback*. They isolate one input — `viper.Set(config.SBSH_ROOT_PROFILES_DIR.ViperKey, t.TempDir())` points the profiles dir at an empty temp dir "so the loader falls back to the hardcoded default profile rather than reading the caller's \$HOME" — but they do **not** isolate the profile *name*. Because the loader resolves the profile name through viper, which has `AutomaticEnv` binding active, the ambient `SBSH_TERM_PROFILE=sbsh-dev-1` overrides the intended default and the loader then fails to find `sbsh-dev-1` in the empty temp dir.

The result is a non-hermetic suite: it fails for any developer running tests from inside an sbsh terminal (the project's own primary dev environment), while passing in CI.

## Root cause
The tests neutralize the profiles *directory* but not the profile *name* env binding. `SBSH_TERM_PROFILE` (and plausibly any other `SBSH_TERM_*` / `SBSH_ROOT_*` key these tests don't explicitly `viper.Set`) leaks in via `viper.AutomaticEnv`. `viper.Reset()` at the top of each test clears prior `viper.Set` calls but does not stop `AutomaticEnv` from re-reading the live process environment.

## Suggested fix
Make the profile-fallback tests hermetic to the ambient profile name — e.g. `t.Setenv("SBSH_TERM_PROFILE", "")` (or explicitly `viper.Set` the profile key to the intended default) in each test's setup, so the fallback path is exercised regardless of the caller's environment. Consider a shared test helper that scrubs the `SBSH_TERM_*` env surface for the package's profile-resolution tests. Surfaced alongside the core-dump fix in #334 / PR #348 (independent code path).

eminwux:
## Summary
- Three default-profile-fallback tests in `cmd/sbsh/terminal/terminal_test.go` (`Test_buildTerminalSpecFromFlags_DefaultProfile`, `Test_processSpec_BuildsFromFlags`, `Test_TerminalCmd_PreRunE_BuildsSpecFromFlags`) isolated the profiles *directory* but not the profile *name*. With `SBSH_TERM_PROFILE` exported — the case for any dev running the suite from inside an sbsh session, since sbsh dogfoods itself — viper's rebound env binding leaked the ambient profile name, so the loader looked for that profile in the empty temp dir and failed instead of exercising the default fallback. CI stayed green only because the container exports no such var.
- Added a `scrubProfileEnv(t)` helper that clears the `SBSH_TERM_*`/`SBSH_ROOT_*` env surface (via `t.Setenv` to empty, auto-restored at cleanup) and called it in each of the three tests after `viper.Reset()`. An empty profile name resolves to `"default"` and then to the hardcoded default profile (`internal/profile/profile.go:352`), which is exactly the path these tests intend to cover.

Implementation note: chose the issue's suggested shared-helper approach (over per-test `viper.Set` of the profile key) because it scrubs the whole prefix surface, guarding against any other `SBSH_TERM_*`/`SBSH_ROOT_*` key a test doesn't explicitly `viper.Set` leaking in the same way.

## Test plan
- [x] `SBSH_TERM_PROFILE=sbsh-dev-1 go test -count=1 ./cmd/sbsh/terminal/` — now passes (previously failed the three tests)
- [x] `env -u SBSH_TERM_PROFILE go test -count=1 ./cmd/sbsh/terminal/` — still passes
- [x] `make sbsh-sb` + `file ./sbsh` confirms ELF executable
- [x] `go build ./...`, `go vet ./cmd/sbsh/terminal/`
- [x] `make test` (full CI target, incl. e2e) — green
- [x] pre-commit hooks passed (gofmt, golangci-lint, addlicense, etc.)

Closes #350

### DIFF
--- a/cmd/sbsh/terminal/terminal_test.go
+++ b/cmd/sbsh/terminal/terminal_test.go
@@ -672,9 +672,27 @@ func Test_setLoggingVarsFromFlags_Defaults(t *testing.T) {
 	}
 }
 
+// scrubProfileEnv neutralizes ambient SBSH_TERM_*/SBSH_ROOT_* environment
+// variables that viper.AutomaticEnv/BindEnv would otherwise leak into
+// profile-resolution tests. viper.Reset() clears prior viper.Set calls but
+// does not stop the rebound env bindings from re-reading the live process
+// environment, so a developer running the suite from inside an sbsh session
+// (which exports SBSH_TERM_PROFILE) would see the default-profile fallback
+// tests fail. Clearing the prefix surface keeps those tests hermetic. See #350.
+func scrubProfileEnv(t *testing.T) {
+	t.Helper()
+	for _, e := range os.Environ() {
+		k, _, _ := strings.Cut(e, "=")
+		if strings.HasPrefix(k, "SBSH_TERM_") || strings.HasPrefix(k, "SBSH_ROOT_") {
+			t.Setenv(k, "")
+		}
+	}
+}
+
 func Test_buildTerminalSpecFromFlags_DefaultProfile(t *testing.T) {
 	viper.Reset()
 	t.Cleanup(viper.Reset)
+	scrubProfileEnv(t)
 
 	logger := slog.New(slog.NewTextHandler(io.Discard, nil))
 	viper.Set(config.SB_ROOT_RUN_PATH.ViperKey, t.TempDir())
@@ -700,6 +718,7 @@ func Test_buildTerminalSpecFromFlags_DefaultProfile(t *testing.T) {
 func Test_processSpec_BuildsFromFlags(t *testing.T) {
 	viper.Reset()
 	t.Cleanup(viper.Reset)
+	scrubProfileEnv(t)
 
 	runPath := t.TempDir()
 	viper.Set(config.SB_ROOT_RUN_PATH.ViperKey, runPath)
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #353 Review — `test: make terminal profile-fallback tests hermetic to SBSH_TERM_PROFILE`**

LGTM — scrubs the `SBSH_TERM_*`/`SBSH_ROOT_*` env surface via `t.Setenv` (auto-restored) so the three default-profile-fallback tests exercise the real fallback path instead of leaking the ambient profile name when run from inside an sbsh session. Test-only, no public surface, both `os`/`strings` already imported, all commits verified.

PR #357, opened 2026-05-22
feat: add captureFormat field to profile YAML spec

### LINKED ISSUE #321: profile: add captureFormat field to profile YAML spec
Non-blocking follow-up surfaced during review of #320.

**`internal/profile/profile.go:462`** (`applyCaptureFormat`): the new `Spec.CaptureFormat` selector is reachable via the `--capture-format` CLI flag, the `SBSH_*_TERM_CAPTURE_FORMAT` env var, and `builder.WithCaptureFormat`, but **not** via profile YAML. Its sibling `CaptureMode` *is* profile-YAML-reachable (`spec.capture.mode` via `FilePermSpec` → `applyPermOverrides`), so a profile-only user can pin the capture file's chmod but cannot pin the capture format.

PR #320 documents this as a deliberate scope deferral ("a profile-spec field can follow if profiles need to pin a format"), and #298's acceptance criteria do not require it — hence non-blocking.

**Suggested fix:** add a `captureFormat` field to the profile terminal spec type and copy it through `CreateTerminalFromProfile` with the same explicit-override > profile > default precedence used for `CaptureMode`, so the profile config surface reaches parity with the CLI/builder surfaces.

eminwux:
## Summary

- Closes the profile-YAML gap surfaced in #321: `Spec.CaptureFormat` was reachable via the `--capture-format` flag, `SBSH_*_TERM_CAPTURE_FORMAT` env vars, and `builder.WithCaptureFormat`, but **not** via profile YAML — while its sibling `CaptureMode` *was* (`spec.capture.mode`). A profile-only user could pin the capture file's chmod but not its serialization format.
- Adds a top-level `spec.captureFormat` field to `api.TerminalProfileSpec` and maps + validates it in `CreateTerminalFromProfile` (mirroring `applyProfilePermBlock`'s build-time validation, with the error naming the YAML field path `spec.captureFormat`).
- Precedence (`explicit override > profile > default`) matches `CaptureMode` for free: `CreateTerminalFromProfile` sets the profile value, and the existing `applyCaptureFormat` overwrites it only when a launch-time override is non-empty.
- Documents the new field in `docs/profiles/README.md` (example + field reference, raw/asciicast, flag/env overrides).

## Design note for the reviewer

- **Placement: top-level `spec.captureFormat`, not nested under the `capture` block.** The `spec.capture` block is an `*api.FilePermSpec`, a type shared by `spec.socket` and `spec.logFile`. Adding a `format` field there would expose a meaningless `socket.format` / `logFile.format` on the other two artifacts. The issue's wording ("add a `captureFormat` field to the profile terminal spec type") also points at `TerminalProfileSpec`. This is the lowest-blast-radius reading; flag if you'd prefer a dedicated capture spec type instead.
- Validation lives at spec-build time (`CreateTerminalFromProfile`), matching how `CaptureMode`/`spec.capture.mode` is validated — `validateProfileDoc` does not validate perm blocks, so adding format validation there would exceed `CaptureMode` parity.

## Test plan

- [x] `make sbsh-sb` — canonical build; `file ./sbsh` confirms ELF executable
- [x] `go build ./...`
- [x] `go vet ./internal/profile/... ./pkg/api/...`
- [x] New unit tests: `CreateTerminalFromProfile` applies/omits/rejects `captureFormat`; `BuildTerminalSpecFromProfile` threads `captureFormat` from YAML end-to-end and explicit override beats profile

Closes #321

### DIFF
--- a/docs/profiles/README.md
+++ b/docs/profiles/README.md
@@ -91,6 +91,7 @@ spec:
   logFile: # Per-terminal log-file permissions (optional)
     mode: "0640" # Octal string; defaults to 0600 if omitted
     gid: 986 # Numeric host GID; omit to leave the group unchanged
+  captureFormat: asciicast # On-disk capture format: "raw" (default) or "asciicast"
 ```
 
 ## Field Reference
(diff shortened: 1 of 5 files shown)


eminwux:
Re-running failed CI jobs once (`Test` workflow). Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #357 Review — `feat: add captureFormat field to profile YAML spec`**

LGTM — closes the #321 profile-YAML gap by threading `spec.captureFormat` through `CreateTerminalFromProfile` with build-time validation; precedence (`explicit > profile > default`) is correct and the override-beats-profile test proves it, top-level placement avoids leaking `format` onto the shared `FilePermSpec` socket/logFile blocks, and the runner already consumes `spec.CaptureFormat`.

PR #358, opened 2026-05-22
fix: synchronize ioClient.outWriter access under clientsMu

### LINKED ISSUE #355: terminalrunner: data race on ioClient.outWriter between handleClient and Detach
## Repro
On clean `origin/main` (reproduced at `40e67cd`), run `TestAttachDetach` under the race detector:

```bash
go test -race -count=2 -run TestAttachDetach .
```

Reproduces on the first run (the `outWriter` field is accessed unsynchronized; `-count=2` just widens the timing window).

## Observed
```
WARNING: DATA RACE
Write at 0x... by goroutine 9:
  ...terminalrunner.(*Exec).handleClient()
      internal/terminal/terminalrunner/connections.go:145
  ...terminalrunner.(*Exec).CreateNewClient.gowrap1()
      internal/terminal/terminalrunner/sockets.go:226

Previous read at 0x... by goroutine 8:
  ...terminalrunner.(*Exec).Detach()
      internal/terminal/terminalrunner/lifecycle.go:405
  ...terminalrunner.TestAttachDetach()
```

`handleClient` (launched as a goroutine by `CreateNewClient`) writes `client.outWriter = aw` at `connections.go:145`. Concurrently, the `Detach` RPC reads `ioClient.outWriter` at `lifecycle.go:405` (`if ioClient.outWriter != nil`). Neither access holds a lock — `ioClient.outWriter` has no synchronization, so a `Detach` racing the just-launched `handleClient` setup is a data race on the field itself.

CI does not run with `-race` (`.github/workflows/test.yaml:50` is plain `go test`), so this is latent in CI but a real concurrency bug: a `Detach` arriving immediately after `Attach` (before `handleClient` finished wiring the writer) reads a torn/zero `outWriter`.

## Root cause
The `ioClient.outWriter` field is written by the per-client `handleClient` goroutine (`connections.go:145`) and read/written by the `Detach` (`lifecycle.go:405`), `cleanupClient` (`connections.go`), and `Close` (`lifecycle.go`) paths with no mutex guarding the field. Distinct from #230 (which addressed `Close()` skipping the `Remove`/`Close` *teardown steps*, not unsynchronized access to the field).

## Suggested fix
Guard `ioClient.outWriter` with a per-client mutex (or set it under `clientsMu` before the goroutine that reads it can run, and read it under the same lock in `Detach`/`cleanupClient`/`Close`). Wiring `outWriter` synchronously in `CreateNewClient` before `addClient` makes it visible — and lockable — to every reader, mirroring how `addClient` already publishes the client. Surfaced during the #351 de-flake (PR #354) via `go test -race`.

eminwux:
## Summary

- `ioClient.outWriter` was written by the per-client `handleClient` goroutine (`connections.go`) and read by the `Detach` RPC (`lifecycle.go`) and `cleanupClient` (`connections.go`) with no synchronization. A `Detach` arriving immediately after `Attach` — before `handleClient` finished wiring the writer — raced the field write (issue #355). `Close` already read the field under `clientsMu.Lock()`, so the field had one synchronized reader and several unsynchronized ones.
- Fix: publish the write in `handleClient` under `clientsMu` (the same lock `Close` already takes), and snapshot the field under `clientsMu.RLock()` in `Detach` and `cleanupClient` before use. No new mutex — the field is part of the client registry `clientsMu` already guards, which is also the issue's suggested approach.

## Design notes

- **Reused `clientsMu` rather than a new per-client mutex.** `Close` already reads `outWriter` under `clientsMu.Lock()`, so the field is conceptually registry-scoped; adding the missing lock at the remaining sites is lower blast radius than introducing a dedicated field mutex.
- **No lock-ordering change.** All three touched sites take `ptyPipesMu` and `clientsMu` only as acquire-release (never nested), preserving the existing `ptyPipesMu`-then-`clientsMu` ordering invariant noted in `Close` (`lifecycle.go`) and `cleanupClient`.
- The early-`Detach`-before-writer-wired case is now a clean `nil` read (Detach skips Remove/Close); the writer is later torn down via the dualcopier reader-error path into `cleanupClient`'s `detachOnce`, unchanged.

## Test plan

- [x] `make sbsh-sb` — canonical build; ELF magic `7f 45 4c 46` confirmed (`file` not installed on this host)
- [x] `go build ./...`
- [ ] `go test -race -count=2 -run TestAttachDetach .` (the issue's repro) — **not run: no C compiler on the dev host, and `-race` requires cgo (`CGO_ENABLED=1` fails with `gcc: not found`).** Note CI does not run `-race` either (`.github/workflows/test.yaml:50` is plain `go test`), so this is not a CI gate; the fix is verified by reasoning + the plain suite passing. Reviewer with a cgo-capable host can run the repro to confirm the DATA RACE warning is gone.

Closes #355

### DIFF
--- a/internal/terminal/terminalrunner/connections.go
+++ b/internal/terminal/terminalrunner/connections.go
@@ -39,8 +39,16 @@ func (sr *Exec) cleanupClient(client *ioClient) {
 	sr.ptyPipesMu.RLock()
 	multiOutW := sr.ptyPipes.multiOutW
 	sr.ptyPipesMu.RUnlock()
-	if multiOutW != nil && client.outWriter != nil {
-		multiOutW.Remove(client.outWriter)
+
+	// Snapshot outWriter under clientsMu — handleClient publishes it under
+	// the same lock, so reading it unsynchronized here races the writer
+	// (issue #355).
+	sr.clientsMu.RLock()
+	outWriter := client.outWriter
+	sr.clientsMu.RUnlock()
+
+	if multiOutW != nil && outWriter != nil {
+		multiOutW.Remove(outWriter)
 	}
 
 	// Closing outWriter signals its drain goroutine to flush any pending
@@ -49,8 +57,8 @@ func (sr *Exec) cleanupClient(client *ioClient) {
 	// remains the sole owner of the lifecycle — double-close on a
 	// *net.UnixConn just logs and is harmless, but a single owner keeps
 	// the path easy to reason about.
-	if client.outWriter != nil {
-		_ = client.outWriter.Close()
+	if outWriter != nil {
+		_ = outWriter.Close()
 	} else if client.conn != nil {
 		// Early-failure fallback: handleClient never wired the writer,
 		// so there is no drain goroutine to take the conn with it.
@@ -142,7 +150,14 @@ func (sr *Exec) handleClient(client *ioClient) {
 	aw := newSubscriberWriter(client.conn, defaultSubscriberBufferBytes, detach, sr.logger)
 	aw.enableBackpressure()
 	aw.seedReplay(initial)
+	// outWriter is read by the Detach RPC, cleanupClient, and Close on
+	// other goroutines, so publish it under clientsMu — the same lock those
+	// readers take — rather than racing the unsynchronized field write a
+	// just-launched handleClient would otherwise expose to a concurrent
+	// Detach (issue #355).
+	sr.clientsMu.Lock()
 	client.outWriter = aw
+	sr.clientsMu.Unlock()
 	multiOutW.Add(aw)
 	go aw.Run()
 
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #358 Review — `fix: synchronize ioClient.outWriter access under clientsMu`**

LGTM — adds the missing `clientsMu` guard to the three unsynchronized `outWriter` sites (publish in `handleClient`, snapshot in `Detach` and `cleanupClient`), matching the lock `Close` already takes; grepped all `outWriter` accesses on head and every reader/writer is now synchronized on the same mutex, lock ordering (`ptyPipesMu`→`clientsMu`, never nested) preserved. The `-race` repro is a documented host-gated smoke (no cgo on the dev host; CI doesn't run `-race` either) — worth running pre-merge on a cgo-capable host to confirm the DATA RACE warning is gone, non-blocking.

PR #359, opened 2026-05-22
test: de-flake TestSubscribe_NoOrphanDrainOnConcurrentClose under parallel load

### LINKED ISSUE #356: test: TestSubscribe_NoOrphanDrainOnConcurrentClose flaky under full-package parallel load
## Repro
On clean `origin/main` (reproduced at `40e67cd`), run the full `terminalrunner` package repeatedly (not the test in isolation):

```bash
for i in $(seq 1 60); do go test -count=1 . ; done
```

Intermittently fails (observed ~1/60 here under full-package parallel load):

```
--- FAIL: TestSubscribe_NoOrphanDrainOnConcurrentClose (0.04s)
    subscribe_close_race_test.go:159: race window never opened — every subscriber was early-rejected
FAIL    github.com/eminwux/sbsh/internal/terminal/terminalrunner
```

## Observed
`subscribe_close_race_test.go:158-160` asserts that **not** every concurrent subscriber took the early ctx-rejection path:

```go
if subscribeErrCount.Load() == concurrentSubscribers {
    t.Fatal("race window never opened — every subscriber was early-rejected")
}
```

Under contention from the rest of the package's goroutines, the `sr.Close(nil)` at line 150 occasionally completes (cancels ctx) before *any* of the concurrent subscriber goroutines reach `Subscribe`'s post-spawn registration — so all `concurrentSubscribers` take the early ctx-check rejection, `subscribeErrCount == concurrentSubscribers`, and the guard fires. The race window the test exists to cover legitimately fails to open under load; the assertion is too strict to be load-robust.

## Root cause
The assertion added by #238 (to ensure the race window opened) has no tolerance for the scheduler legitimately closing the window under parallel load. It is an all-or-nothing equality check (`== concurrentSubscribers`) on a timing-sensitive count. Same class as #351 — a test assertion that races scheduler jitter rather than a logic error.

## Suggested fix
Make the window-opened guard robust to jitter rather than asserting on a single full-package run: e.g. retry the concurrent-subscribe burst until at least one subscriber registers before Close (bounded attempts), or relax the guard to `subscribeErrCount.Load() < concurrentSubscribers` only after confirming the burst raced Close at all, or deterministically gate Close behind a barrier that guarantees the window opens. The intent (fail loudly if the race window never opens across the suite) should be preserved without flaking on a single contended run. Surfaced alongside the #351 de-flake (PR #354); independent test, independent fix.

eminwux:
## Summary
- The `subscribeErrCount == concurrentSubscribers` guard at `subscribe_close_race_test.go:158` was an all-or-nothing equality check on a timing-sensitive count. Under full-package parallel load `sr.Close(nil)` could cancel ctx before *any* of the 100 subscriber goroutines reached `addSubscriber`, so all took the early ctx-rejection path and the guard fired spuriously (~1/60). Same class as #351 (PR #354): a test assertion racing scheduler jitter, not a logic error.
- Replaced the fixed `time.Sleep(2ms)` + post-hoc count guard with a deterministic barrier: poll until the registry is non-empty (`len(sr.subscribers) > 0`) before calling `Close`. A registered subscriber having passed `Subscribe`'s early ctx-check (`subscribe.go:108`) and run `addSubscriber` (`subscribe.go:131`) *is* the meaningful #226 window — its drain goroutine is what `closeAllSubscribers` / the late ctx-check must reap. Nothing drains the registry before `Close`, so once non-empty it stays non-empty until `Close` (no TOCTOU).
- With 100 concurrent subscribers, firing `Close` right after the first registration still leaves the bulk of the burst arriving post-`Close`, so the post-snapshot arrival path the test exists to cover stays well exercised. Both downstream invariants (registry empties, goroutine count settles) are unchanged. Dropped the now-unused `sync/atomic` import.

## Test plan
- [x] `go build ./...`
- [x] Issue repro: `for i in $(seq 1 60); do go test -count=1 .; done` in `internal/terminal/terminalrunner` — 60/60 clean (was ~1/60 fail on `main`)
- [x] `make test` (full CI suite incl. e2e) — all green
- [ ] `make test-race` / `go test -race` — not run: `-race` requires cgo and this host has no C compiler (`cgo: C compiler "gcc" not found`). The change is test-only and does not alter the subsMu/ctxCancel ordering the race detector covers.

Closes #356

### DIFF
--- a/internal/terminal/terminalrunner/subscribe_close_race_test.go
+++ b/internal/terminal/terminalrunner/subscribe_close_race_test.go
@@ -24,7 +24,6 @@ import (
 	"log/slog"
 	"runtime"
 	"sync"
-	"sync/atomic"
 	"testing"
 	"time"
 
@@ -113,7 +112,6 @@ func TestSubscribe_NoOrphanDrainOnConcurrentClose(t *testing.T) {
 	goroutinesBefore := runtime.NumGoroutine()
 
 	var wg sync.WaitGroup
-	var subscribeErrCount atomic.Int32
 	start := make(chan struct{})
 
 	for i := range concurrentSubscribers {
@@ -127,13 +125,10 @@ func TestSubscribe_NoOrphanDrainOnConcurrentClose(t *testing.T) {
 			if err != nil {
 				// Post-Close rejection is the expected outcome for the
 				// late-arriving half; the cliFD was already closed by
-				// Subscribe's error path. Count the rejection so the
-				// post-wg.Wait() assertion can fail if every single
-				// subscriber was rejected (race window never opened).
+				// Subscribe's error path.
 				if !errors.Is(err, errTerminalClosing) {
 					t.Errorf("subscribe (idx=%d): unexpected error: %v", idx, err)
 				}
-				subscribeErrCount.Add(1)
 				return
 			}
 			for _, fd := range resp.FDs {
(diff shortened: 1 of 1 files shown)


eminwux:
**PR #359 Review — `test: de-flake TestSubscribe_NoOrphanDrainOnConcurrentClose under parallel load`**

LGTM — swaps the timing-sensitive `subscribeErrCount==concurrentSubscribers` guard for a deterministic `len(sr.subscribers) > 0` barrier before `Close`; the TOCTOU claim holds (registry only grows pre-`Close` since no drain self-closes until ctx cancels), both invariants stay exercised with ~99/100 arriving post-`Close`, and the dead symbol + `sync/atomic` import are removed cleanly. `-race` is a documented host-gated skip (no cgo), test-only change.

PR #362, opened 2026-05-27
fix: close bind-then-chmod EACCES window in OpenSocketCtrl

### LINKED ISSUE #361: terminal: bind-then-chmod race in OpenSocketCtrl briefly exposes 0o600 socket — group attach hits EACCES
## Context

Follow-up to #188. That issue made the control socket's final mode/gid
configurable so a host operator in the kukeon group can attach across uid
boundaries. The configured perms are applied via `applySocketPerms(mode, gid)`
**after** `net.ListenConfig.Listen(...)` binds the socket inode. Between
`Listen` and `applySocketPerms`, the socket already accepts `connect(2)` but
is still `srw-------` (mode `0o666 & ~umask`, owner-only when the daemon's
umask is the common `0o077`). A client that dials inside that window gets
`EACCES` even though the final post-chmod state is the configured `0o660`.

In the kukeon embedding this surfaces as: `kuke run <cell>` on a Stopped
cell does `StartCell` → returns when containerd reports the work container
`Running` (post-start liveness probe, 300ms grace) → immediately dials the
sbsh control socket. The dial races sbsh's chmod and hits EACCES:

```
$ kuke run kukeon-dev-root-0
Cell "kukeon-dev-root-0" (realm "default", space "default", stack "default")
  - metadata: already existed
  - cgroup: already existed
  - root container: already existed
  - container "root": already existed
  - container "work": already existed
  - containers: started
Error: error attaching client: ping failed: dial unix /opt/kukeon/s/68eae86c9bdf7052: connect: permission denied
```

A retry seconds later succeeds — the chmod has landed by then.

## Repro (sbsh-only, no kukeon needed)

```bash
# Group the operator is in but the listener uid is not
sudo groupadd -g 4242 sbtest
sudo usermod -aG sbtest "$USER"; newgrp sbtest   # or open a new login

# Listener runs as root, configured to chown :4242 mode 0660 *after* Listen
sudo install -d -m 02750 -o 0 -g 4242 /tmp/sbtest
sudo sbsh terminal \
  --run-path /tmp/sbtest \
  --socket /tmp/sbtest/socket \
  --socket-mode 0660 \
  --socket-gid 4242 \
  -- /bin/sleep 60 &

# Race the dial: a tight loop that connects as the unprivileged user
# catches the EACCES window most launches
for i in $(seq 1 50); do
  sbsh attach --socket /tmp/sbtest/socket </dev/null 2>&1 | head -1
done
```

A subset of attempts print `connect: permission denied`; the rest succeed.
The window is short (microseconds in normal load) but is reliably hit when
the dial fires immediately after a fresh listener bind — which is exactly
what `kuke run` against a Stopped cell does.

## Observed

`internal/terminal/terminalrunner/sockets.go:172-186`:

```go
ctrlLn, errListen := lc.Listen(sr.ctx, \"unix\", socketFile)
// ↑ socket inode now exists at mode (0o666 & ~umask), e.g. 0o600
// ↑ kernel accepts connect(2) from any uid the inode permits — for
//   0o600 root-owned, that's only root.

sr.lnCtrl = ctrlLn

if errPerms := sr.applySocketPerms(socketFile, socketMode, socketGID); errPerms != nil {
    _ = ctrlLn.Close()
    return errPerms
}
// ↑ inode now at configured 0o660 :kukeon. Group-mode dials succeed
//   from here on.
```

The pure-Go `net.ListenConfig` has no `Mode`/`Gid` hook, so closing the
window means either (a) bind with the right umask in effect, (b) bind on
a side path and rename, or (c) gate readers off a metadata flag that's
only updated after the chmod.

## Root cause

`OpenSocketCtrl` treats listener creation and perm-application as two
independently-recoverable steps. That layering is fine for failure
handling, but it means the socket is dial-reachable in its pre-chmod state.
Issue #188 fixed the *final* mode but didn't address the transition.

## Suggested fixes

In rough order of cleanness:

### Option A — Set umask around \`Listen\` so the socket is born with the right mode

\`\`\`go
// inside OpenSocketCtrl, just before Listen
prev := syscall.Umask(int(^socketMode & 0o777))
ctrlLn, errListen := lc.Listen(sr.ctx, \"unix\", socketFile)
syscall.Umask(prev)
// applySocketPerms still runs to apply the chown and re-assert mode
// (idempotent), but the EACCES window is gone.
\`\`\`

Pros: one extra syscall pair, no metadata schema change, closes the
window for every caller of \`OpenSocketCtrl\` including \`UseListener\` if it
takes the same path.

Cons: \`syscall.Umask\` is process-wide. If sbsh is embedded as a library
inside a multi-goroutine host (kukeond is the concrete case), the
goroutine-local umask trick needs a \`runtime.LockOSThread\` around it, or
a mutex serialising listener creation. The locked-thread variant is the
cleanest IMO.

### Option B — Bind on a side path, chmod, rename into place

\`\`\`go
tmp := socketFile + \".tmp.\" + randomToken()
ctrlLn, _ := lc.Listen(sr.ctx, \"unix\", tmp)
_ = os.Chmod(tmp, mode)
if gid != nil { _ = os.Chown(tmp, -1, *gid) }
_ = os.Rename(tmp, socketFile)
\`\`\`

Pros: no global umask side effects.

Cons: \`rename(2)\` of a bound unix-domain socket is a kernel quirk to
double-check (it works on Linux today; portability not great, but sbsh
is Linux-only). Adds a small window where a stale-path dial sees ENOENT
instead of EACCES — strictly an improvement (\`net.Dial\` distinguishes
the two and retries are easier on ENOENT).

### Option C — Expose a \"ready\" gate via metadata; have callers wait on it

Keep \`applySocketPerms\` where it is, but only flip
\`metadata.Status.SocketReady = true\` (new field) and \`updateMetadata\`
*after* it succeeds. Then kukeon's \`AttachContainer\` server-side handler
can poll the metadata file (or sbsh exposes a healthz over the socket
that doesn't require the chmod to have landed — chicken-and-egg there).

Pros: explicit handshake, no umask gymnastics.

Cons: requires every embedder to opt in to the gate. Doesn't help direct
\`sbsh attach\` racing against \`sbsh terminal &\`.

I'd suggest **Option A with \`runtime.LockOSThread\`** as the closing fix —
it's local to \`OpenSocketCtrl\`, doesn't change any public API or schema,
and covers every caller.

## Out of scope

- The kukeon side has its own potential mitigation (retry on EACCES in
  the run path) but that's a band-aid; the right fix lives here.
- Capture/log file perms (#200) — those are opened with \`O_CREATE\` so
  \`O_CREAT\` + \`mode\` already lands in one syscall; not the same race.

eminwux:
## Summary

- OpenSocketCtrl previously let `net.ListenConfig.Listen` bind the control socket with whatever umask the daemon process happened to carry (commonly `0o077`), leaving the inode briefly reachable at `0o666 & ~daemonUmask` (typically `0o600`) before `applySocketPerms` chmoded it to the configured mode. A group-member client that dialed inside that window hit EACCES even though the final post-chmod state would have admitted it — exactly the symptom reported in #361 against `kuke run`.
- This change introduces `listenUnixWithMode`, which temporarily sets the process umask so the socket inode is born at the configured `SocketMode`. The save/set/Listen/restore sequence runs under `runtime.LockOSThread` to keep the goroutine pinned to a single OS thread, and a package-level mutex serializes the umask manipulation against in-package concurrent callers. The follow-up `applySocketPerms` chmod becomes idempotent at that point but is left in place so the chown path still applies when `SocketGID` is set and the parent directory does not propagate gid.
- This is Option A from the issue, with the `runtime.LockOSThread` variant the issue author called out as cleanest. Option B (bind-on-side-path + `rename(2)`) was rejected per the issue's own portability caveat about `rename(2)` of a bound AF_UNIX socket. Option C (metadata gate) was rejected because it requires every embedder to opt in and does not help direct `sbsh attach` racing `sbsh terminal &`.

### Note on residual scope

`syscall.Umask` is process-wide on Linux. The mutex stops in-package concurrent OpenSocketCtrl callers from racing on the global umask, but the brief Listen window can still leak the temporary umask to file creations in *other* packages running in the same process. This is documented inline at `listenerUmaskMu` and called out in the issue itself; closing it would require `unshare(CLONE_FS)` or moving the listener to a dedicated process. UseListener (the externally-bound-listener path used by `pkg/terminal/server`) is unchanged — the bind happens in the caller, so the umask trick has to live there if the embedder needs the same guarantee.

## Test plan

- [x] `make sbsh-sb` (canonical build; `file ./sbsh` returns ELF executable)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./internal/terminal/terminalrunner/...` — new tests pass
- [x] `make test` — full CI suite passes (note: `Test_RunStopTerminals_AllAggregation` is a pre-existing flake — reproduces at the same rate on baseline `origin/main`, unrelated to this change)
- [x] New `TestListenUnixWithMode_SocketBornAtRequestedMode` proves the inode mode immediately after the helper returns equals the configured mode under a hostile `0o077` umask — a regression replacing the helper with bare `lc.Listen` would observe `0o600` here.
- [x] New `TestListenUnixWithMode_DefaultModeWhenZero` locks in the `mode==0` fallback to `defaultSocketMode`.
- [x] New `TestOpenSocketCtrl_SocketReachableAtConfiguredMode` end-to-end check under hostile umask.

Closes #361

### DIFF
--- a/internal/terminal/terminalrunner/sockets.go
+++ b/internal/terminal/terminalrunner/sockets.go
@@ -17,9 +17,12 @@
 package terminalrunner
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"os"
+	"runtime"
+	"sync"
 	"time"
 
 	"github.com/eminwux/sbsh/pkg/api"
@@ -45,6 +48,50 @@ const defaultSocketMode os.FileMode = 0o600
 // flags, env vars, or profile blocks. Same contract shape as
 // defaultSocketMode for the control socket.
 const defaultArtifactMode os.FileMode = 0o600
+
+// listenerUmaskMu serializes the umask save/set/Listen/restore sequence
+// in listenUnixWithMode against in-package concurrent callers. umask(2)
+// is process-wide (it lives in the kernel's shared fs_struct on Linux),
+// so two concurrent OpenSocketCtrl callers without this mutex could
+// observe each other's temporary umask. The mutex does not isolate the
+// brief Listen window from file creations in *other* packages that
+// share this process — that's an inherent cost of using umask to plug
+// the bind-then-chmod EACCES window. The alternative (bind-on-side-path
+// then rename(2)) relies on a kernel quirk for AF_UNIX sockets and is
+// worse on portability and on the ENOENT-vs-EACCES dial-error story.
+//
+//nolint:gochecknoglobals // process-wide invariant guard, like the umask it serializes
+var listenerUmaskMu sync.Mutex
+
+// listenUnixWithMode binds an AF_UNIX listener with the process umask
+// temporarily set so the socket inode is born at the configured mode,
+// not at (0o666 & ~daemonUmask). This closes the bind-then-chmod
+// EACCES window in OpenSocketCtrl: pre-fix, the socket was briefly
+// reachable at the daemon's restrictive default (typically 0o600 under
+// a 0o077 umask), so a group-member client that dialed before
+// applySocketPerms landed hit EACCES even though the final post-chmod
+// state would have admitted it.
+//
+// runtime.LockOSThread pins the goroutine so the save/restore sequence
+// runs on a single OS thread; listenerUmaskMu serializes the temporary
+// umask against in-package concurrent callers. Callers should still
+// run applySocketPerms after this returns — the chmod becomes
+// idempotent at that point, but the chown still needs to apply when
+// SocketGID is set and the parent directory does not propagate gid.
+func listenUnixWithMode(ctx context.Context, socketFile string, mode os.FileMode) (net.Listener, error) {
+	if mode == 0 {
+		mode = defaultSocketMode
+	}
+	listenerUmaskMu.Lock()
+	defer listenerUmaskMu.Unlock()
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
+	//nolint:mnd // 0o777 is the standard permission-bit mask for umask(2)
+	prev := unix.Umask(int(^mode & 0o777))
+	defer unix.Umask(prev)
+	var lc net.ListenConfig
+	return lc.Listen(ctx, "unix", socketFile)
+}
 
 // applySocketPerms chmods (and optionally chowns the group of) the
 // control-socket inode after a listener has been bound to it. Called
@@ -169,9 +216,12 @@ func (sr *Exec) OpenSocketCtrl() error {
 		)
 	}
 
-	// Listen to CONTROL SOCKET
-	var lc net.ListenConfig
-	ctrlLn, errListen := lc.Listen(sr.ctx, "unix", socketFile)
+	// Listen to CONTROL SOCKET — bind with the umask set so the
+	// inode is born at socketMode and there is no window during
+	// which a group-member client dialing the socket hits EACCES
+	// because the daemon's umask masked off group access. See #361
+	// and listenUnixWithMode for the full rationale.
+	ctrlLn, errListen := listenUnixWithMode(sr.ctx, socketFile, socketMode)
 	if errListen != nil {
 		sr.logger.Error("OpenSocketCtrl: failed to listen", "socket", socketFile, "error", errListen)
 		return fmt.Errorf("listen ctrl: %w", errListen)
--- a/internal/terminal/terminalrunner/sockets_test.go
+++ b/internal/terminal/terminalrunner/sockets_test.go
@@ -28,6 +28,7 @@ import (
 	"time"
 
 	"github.com/eminwux/sbsh/pkg/api"
+	"golang.org/x/sys/unix"
 )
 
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #362 Review — `fix: close bind-then-chmod EACCES window in OpenSocketCtrl`**

LGTM — `listenUnixWithMode` binds the AF_UNIX inode at the configured mode under a package mutex + `runtime.LockOSThread`, closing the EACCES window per #361 Option A. Math, defer order (umask → thread → mutex), and the `mode==0` → `defaultSocketMode` fallback match `applySocketPerms`. The three new tests pin the inode mode immediately after the helper returns under a hostile `0o077` umask, the zero-mode fallback, and the end-to-end `OpenSocketCtrl` path. Residual cross-package umask leak and the `UseListener` carve-out are documented in the PR body.

PR #367, opened 2026-05-29
fix: restore parent terminal on client detach and process exit

### LINKED ISSUE #364: client: restore parent terminal on detach and on process exit
## Context

When an attached session ends — either the user detaches (`^]^]`) or the spawned process exits and the session tears down — the **parent terminal is left in a broken state: the shell prompt disappears**, keystrokes aren't echoed, and the shell's read loop misbehaves. The user has to `reset`/`stty sane` to recover. This happens on every session teardown, so it affects the core UX of both `sbsh` (run) and `sb attach`.

Root cause (two compounding defects in the client):

1. **The stdin→socket copier is never unblocked before process exit.** It is started at `internal/client/clientrunner/io.go:98` reading `sr.stdin` (os.Stdin) directly, and `internal/dualcopier/copier.go` only checks `ctx.Done()` *between* reads (`copier.go:176-189`) — it cannot interrupt the blocking `r.Read(buf)` in `readWriteBytes` (`copier.go:49-53`). No teardown path closes os.Stdin or sets a read deadline on it; `CopierManager` only unblocks the *socket* side via deadlines on the `*net.UnixConn` (`copier.go:144-154`). So the goroutine stays parked in `os.Stdin.Read()` until the process exits. Because os.Stdin (fd 0, shared with the parent shell) was put into non-blocking mode by Go's runtime poller and is never returned to blocking — the abandoned read prevents that restoration — the parent shell inherits fd 0 in `O_NONBLOCK` and its reads return `EAGAIN`. This is the symptom common to **both** the detach and process-exit paths (cf. Go stdin non-blocking leak, golang/go#29137). The repeated `.Fd()` calls in `toExitShell` (`terminal.go:47`) and `pty.Getsize` (`io.go:177,203`) flip the same fd's poll/blocking state, compounding the fragility.

2. **`Exec.Detach()` skips the termios restore.** `Exec.Close()` calls `sr.toExitShell()` (`internal/client/clientrunner/lifecycle.go:192`) to run `term.Restore` (`terminal.go:44-57`), but `Exec.Detach()` (`lifecycle.go:213-225`) returns without it — it writes the `Detached` message with explicit `\r\n` (i.e. while still in raw mode) and never restores. So on a graceful detach the raw→cooked restore is skipped unless the conn-close path happens to drive `Close()` via `EvError`. (There is a commented-out `defer term.Restore(...)` at `terminal.go:35` hinting a guaranteed restore was intended.)

Net effect: clean exit is broken by the leaked non-blocking stdin (termios itself is restored by `Close`); detach is doubly broken (raw mode possibly left set *and* the non-blocking leak).

## Proposal

Guarantee that the client restores the parent terminal on every session-end path — user detach, remote process exit, peer hangup, and ctx cancellation:

- Shut the stdin reader down deterministically before the process exits, so the os.Stdin descriptor is returned to blocking mode (e.g. interrupt the blocked read on teardown rather than abandoning it).
- Restore termios (raw→cooked) unconditionally on the detach path, not only via `Close()`.

A single guaranteed "restore tty + unblock stdin" at the session boundary is more robust than the current event-driven, `Close()`-only restore.

## Acceptance criteria

- [ ] After a user detach (`^]^]`), the parent shell prompt, echo, and line editing work normally with no manual `reset`/`stty sane`.
- [ ] After the spawned process exits and `sbsh`/`sb` exits, the parent shell prompt, echo, and line editing work normally.
- [ ] The parent terminal's stdin descriptor is left in blocking mode (not `O_NONBLOCK`) after the client exits, on both paths above.
- [ ] Termios is restored to its pre-attach (cooked) state on the detach path, not only on the `Close()` path.
- [ ] Restoration also holds for peer-hangup and ctx-cancel (SIGINT/SIGTERM) teardown.
- [ ] A regression test asserts the tty state (blocking flag and `-icanon`/`-echo` cleared) is restored after a simulated detach and after a simulated process exit.

## Notes

- Symptom reproduces on both `sbsh` (RunNewTerminal) and `sb attach` (AttachToTerminal).
- Key code: `internal/client/clientrunner/lifecycle.go` (`Detach` 213-225, `Close` 162-203), `internal/client/clientrunner/terminal.go` (`toBashUIMode`/`toExitShell`/`toRawMode`), `internal/client/clientrunner/io.go:98` (stdin copier), `internal/dualcopier/copier.go` (`CopierManager`, `runReadWriter`, `readWriteBytes`).
- Manual confirmation: capture `stty -a` before launch and right after exit — expect `-icanon -echo` lingering on the detach path and a non-blocking stdin descriptor on both paths.
- Reference: Go stdin non-blocking leak on exit — golang/go#29137.

eminwux:
## Summary

Fixes the broken parent terminal after every attached-session teardown (issue #364): on user detach (`^]^]`) and on spawned-process exit, the shell prompt vanished, keystrokes stopped echoing, and the read loop misbehaved until a manual `reset`/`stty sane`.

Two compounding defects, both addressed:

- **Leaked stdin reader / non-blocking descriptor.** The stdin→socket copier parks in `os.Stdin.Read()` and was never unblocked before exit. Root issue: `term.MakeRaw(int(stdin.Fd()))` (and `pty.Getsize(stdin)`) call `os.File.Fd()`, which pulls the descriptor out of the runtime poller and flips it to blocking — so `SetReadDeadline` could not interrupt the parked read, and the descriptor was left in a bad state for the parent shell (cf. golang/go#29137).
- **Detach skipped the termios restore.** `Exec.Close()` ran the raw→cooked restore but `Exec.Detach()` did not; it only happened if the conn-close path happened to drive `Close()` via `EvError`.

### Approach

- Set raw mode / read window size through `SyscallConn().Control` (new `withRawFd` helper + `ttyGetsize`) instead of `os.File.Fd()`, so the descriptor **stays registered with the runtime poller**. That is what makes the deadline-based unblock work. `term.MakeRaw`/`term.Restore` are still used inside the callback, so cross-platform termios handling (linux/darwin/freebsd release targets) is unchanged.
- Add a single, idempotent `restoreParentTerminal()` (guarded by `sync.Once`) that runs on **every** teardown path — detach, process exit, peer hangup, ctx-cancel (SIGINT/SIGTERM). It: cancels the ctx (unblocks the socket side via `CopierManager`), sets a past read deadline on stdin (unblocks the parked stdin read), waits for both copier goroutines to exit (new `dualcopier.Copier.Wait()`), restores termios raw→cooked, then `syscall.SetNonblock(fd, false)` returns the descriptor to blocking mode.
- Wire `restoreParentTerminal()` into both `Close()` and `Detach()`.

### Non-obvious calls (for reviewer scrutiny)

- I deliberately **avoid `os.File.Fd()`** during raw-mode setup; an empirical check (pty slave) confirmed `Fd()` makes the read a truly-blocking syscall that `SetReadDeadline` cannot interrupt, whereas the `SyscallConn().Control` path keeps it pollable and the deadline unblocks it.
- `restoreParentTerminal` cancels the ctx on the **detach** path too (previously detach left the ctx live until `Close`). The detach RPC has already returned by then, and the controller's ctx is a parent of the runner's ctx, so this only unblocks the runner's own copiers; `Close` still runs afterward (idempotent no-op for restore).
- The regression test is `restore_linux_test.go` (Linux suffix) because it introspects termios flags via `TCGETS`; CI runs on Linux. The cross-platform production code is not gated.

## Test plan

- [x] `make sbsh-sb` (canonical build) + `file ./sbsh` is ELF
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — all pass except a **pre-existing, unrelated flake** in `cmd/sb/stop` (`Test_StopViaRPC_StaleSocket_RespectsTimeout`); reproduced on clean `origin/main` under `-count=5`, and `cmd/sb/stop` has no dependency on the changed packages (verified via `go list -deps`). Distinct from the stop-flake already in flight in #366.
- [x] `go test -race ./internal/client/clientrunner/` (new restore tests + existing)
- [x] `pre-commit run --files <changed>` (go-fmt, go-mod-tidy, golangci-lint, addlicense)
- [x] New regression tests assert the parent tty is left blocking (`O_NONBLOCK` cleared) with cooked line discipline (`ECHO`/`ICANON` set) after a simulated process exit (`Close`) and a simulated detach (`Detach`), plus an idempotency test for the detach-then-close sequence. Verified they fail against the pre-fix behavior.
- [ ] Manual `stty -a` before/after on a real interactive tty — not run (no interactive terminal in the dev sandbox); covered by the pty-backed regression tests above.

Closes #364

### DIFF
--- a/internal/client/clientrunner/client_runner_exec.go
+++ b/internal/client/clientrunner/client_runner_exec.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"sync"
 
+	"github.com/eminwux/sbsh/internal/dualcopier"
 	"github.com/eminwux/sbsh/pkg/api"
 	"github.com/eminwux/sbsh/pkg/rpcclient/terminal"
 	"golang.org/x/term"
@@ -39,6 +40,14 @@ type Exec struct {
 
 	uiMode        UIMode
 	lastTermState *term.State
+
+	// copier drives the stdin<->socket I/O for the attached session; it is
+	// retained so restoreParentTerminal can wait for both copier goroutines to
+	// exit before returning the stdin descriptor to blocking mode.
+	copier *dualcopier.Copier
+	// restoreOnce guarantees the parent-terminal restore runs exactly once
+	// across the detach, process-exit, peer-hangup, and ctx-cancel paths.
+	restoreOnce sync.Once
 
 	// stdin/stdout/stderr are the user-facing terminal handles. They
 	// default to os.Stdin/Stdout/Stderr but can be overridden by
--- a/internal/client/clientrunner/io.go
+++ b/internal/client/clientrunner/io.go
@@ -27,7 +27,6 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/creack/pty"
 	"github.com/eminwux/sbsh/internal/dualcopier"
 	"github.com/eminwux/sbsh/internal/filter"
 	"github.com/eminwux/sbsh/pkg/api"
@@ -75,6 +74,7 @@ func (sr *Exec) startConnectionManager() error {
 	}
 
 	dc := dualcopier.NewCopier(sr.ctx, sr.logger)
+	sr.copier = dc
 
 	// Only create escape filter if detach keystroke is enabled
 	var escapeFilter filter.Filter
@@ -174,7 +174,7 @@ func (sr *Exec) attach() error {
 
 func (sr *Exec) forwardResize() error {
 	// Send initial size once (use the client's TTY: sr.stdin)
-	if rows, cols, errSize := pty.Getsize(sr.stdin); errSize == nil {
+	if rows, cols, errSize := ttyGetsize(sr.stdin); errSize == nil {
 		const resizeTimeout = 100 * time.Millisecond
 		ctx, cancel := context.WithTimeout(sr.ctx, resizeTimeout)
 		defer cancel()
@@ -200,7 +200,7 @@ func (sr *Exec) forwardResize() error {
 				return
 			case <-ch:
 				// Query current terminal size again on every WINCH
-				rows, cols, err := pty.Getsize(sr.stdin)
+				rows, cols, err := ttyGetsize(sr.stdin)
 				if err != nil {
 					sr.logger.WarnContext(sr.ctx, "forwardResize: Getsize failed", "error", err)
 					continue
--- a/internal/client/clientrunner/lifecycle.go
+++ b/internal/client/clientrunner/lifecycle.go
@@ -189,7 +189,7 @@ func (sr *Exec) Close(_ error) error {
 		}
 	}
 
-	_ = sr.toExitShell()
+	sr.restoreParentTerminal()
 	sr.logger.Debug("Close: cleanup complete")
 
 	sr.metadataMu.Lock()
@@ -216,9 +216,14 @@ func (sr *Exec) Detach() error {
 	}
 
 	sr.logger.Info("Client detached", "client_id", sr.id)
-	if _, err := sr.stdout.WriteString("\x1b[93m\r\nDetached\x1b[0m\r\n"); err != nil {
-		sr.logger.Warn("Failed to write detach message to stdout", "error", err)
-		return err
+	// Write the confirmation while still in raw mode (explicit \r\n), then
+	// restore the parent terminal unconditionally — the detach path must not
+	// rely on Close() being driven later via the conn-close EvError. See #364.
+	_, werr := sr.stdout.WriteString("\x1b[93m\r\nDetached\x1b[0m\r\n")
+	sr.restoreParentTerminal()
+	if werr != nil {
+		sr.logger.Warn("Failed to write detach message to stdout", "error", werr)
+		return werr
 	}
 
 	return nil
--- a/internal/client/clientrunner/restore_linux_test.go
+++ b/internal/client/clientrunner/restore_linux_test.go
@@ -1,0 +1,183 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package clientrunner
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"sync"
+	"syscall"
+	"testing"
+
+	"github.com/eminwux/sbsh/pkg/api"
+	"golang.org/x/sys/unix"
+)
+
+// ttyState reports the descriptor's blocking flag and the ECHO/ICANON termios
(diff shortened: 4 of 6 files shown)


eminwux:
## CI diagnosis + fix (E2E job)

The E2E job failed: `TestSbsh_Detach` hung ~9m49s and blew the 10m suite timeout, with `TestLifecycle_SpawnAttachWriteReadDetachStop`, `TestSb_WriteRead_Roundtrip`, `TestSbsh_Default`, and `TestSbsh_CustomId` all timing out on teardown ("timeout waiting for … to exit"). The local test plan **excluded `/e2e`**, which is why this didn't surface before push.

### Root cause

`restoreParentTerminal`'s newly-added **unbounded `sr.copier.Wait()`** deadlocks on the real binary. The premise that setting raw mode via `SyscallConn().Control` keeps stdin pollable so `SetReadDeadline` can unblock the parked read **does not hold for `os.Stdin`**: a descriptor's poller registration is fixed when the `*os.File` is constructed, and the Go runtime constructs the standard streams **non-pollable** — independent of how the fd is later accessed. Verified directly under a real tty:

```
os.Stdin.SetReadDeadline(now)            -> "file type does not support deadline"
pty.Open() slave (*os.File).SetReadDeadline(now) -> nil
```

So in the attach path (`sr.stdin == os.Stdin`) `SetReadDeadline` is a silent no-op (logged at debug only), the stdin→socket copier stays parked in `Read()` forever, and `copier.Wait()` blocks forever → the process never exits → e2e teardown timeouts.

The unit tests in `restore_linux_test.go` pass because `openTTY` builds a **pollable** `*os.File` via `pty.Open()` (the helper even asserts `O_NONBLOCK` is set post-attach) — so the deadline works there and `Wait()` returns. That gap is exactly why the green unit tests didn't catch the e2e hang.

Confirmed as a regression introduced by this PR, not pre-existing: `TestLifecycle_SpawnAttachWriteReadDetachStop` passes on clean `origin/main` and hangs on the branch tip.

### Fix (`b05550d`)

Bound `copier.Wait()` with a 500 ms ceiling. The socket→stdout copier still unblocks promptly via the ctx-cancel deadline on the `UnixConn`; the un-interruptible `os.Stdin` reader (which `main` never waited on either) is reaped at process exit, which happens immediately after restore. Misleading comments asserting `os.Stdin` is pollable are corrected. `term.Restore` + `SetNonblock(false)` still run on every teardown path, so the #364 fix is unchanged.

### Verification (local, on the fixed branch)

- [x] Full `make test` — green, e2e included (e2e 10.2 s vs the 10 m CI timeout)
- [x] The four formerly-hanging tests + `TestSbsh_Detach` pass
- [x] `go build ./...`, `go vet ./...`, `pre-commit` (incl. golangci-lint), `go test -race ./internal/client/clientrunner/ ./internal/dualcopier/`

eminwux:
**PR #367 Review — `fix: restore parent terminal on client detach and process exit`**

LGTM — routes raw-mode/resize/restore through `SyscallConn().Control` (keeping stdin pollable) and runs a single `sync.Once`-guarded `restoreParentTerminal()` on every teardown path, fixing the broken parent tty from #364. Verified: the `errgroup.Group.Wait()` concurrent-call claim is sound, ctx-cancel propagates to unblock the socket copier, the pollable-vs-`os.Stdin` distinction is handled by the bounded `copierWaitTimeout`, the leaked-reader window is reaped by the post-detach `EvError → Close → Run`-exit, all new symbols are referenced, and the pty regression tests assert the actual termios + `O_NONBLOCK` end state.

One deferred verification for you to run pre-merge: the manual `stty -a` before/after on a real interactive tty (the last test-plan box) can't run in the agent sandbox — non-blocking, the pty-backed tests cover it.

PR #369, opened 2026-05-30
fix: restore parent terminal output-side DEC mode state on teardown

### LINKED ISSUE #365: client: restore output-side terminal state (cursor, alt-screen) on teardown
## Context

After detaching from or exiting an attached session, the parent terminal's **cursor is gone** (no visible/blinking cursor), and if a full-screen program was running it can be left in the alternate-screen buffer. Input still works (this is distinct from #364, which is the input-side stdin/termios breakage) — the problem is purely *output-side* terminal state that the client never normalizes on teardown.

Root cause: the client passes remote PTY output through verbatim and emits **no compensating reset** when the session ends. The reader copier writes raw remote bytes straight to the parent terminal (`internal/client/clientrunner/io.go:116`, filter `nil`), and the only other client stdout writes are the attach help line (`io.go:108`) and the `Detached` message (`lifecycle.go:219`). There is no `\x1b[?25h` (show cursor), no `\x1b[?1049l` (leave alt-screen), and no soft reset anywhere on the detach or exit path.

Two ways the parent terminal ends up in a bad output state:

1. **The remote program/shell sets it.** A TUI (vim, htop, less, fzf) or a prompt emits `\x1b[?25l` (hide cursor) or a DECSCUSR cursor-style change (`\x1b[… q`); those bytes stream to the parent terminal and are never undone on teardown, so the cursor stays hidden.
2. **The attach repaint sets it.** The server-side repaint emits `escEnterAltScreen` (`\x1b[?1049h`) when alt-screen is active and `escHideCursor` (`\x1b[?25l`) when the live cursor is hidden (`internal/terminal/terminalrunner/terminal_screen.go:176-191`). These modes are entered on attach with no symmetric exit in the client teardown.

This is a different fix surface from #364: #364 restores the input fd (stdin blocking mode + raw→cooked termios); that does not bring the cursor back. This issue needs an output-side normalization.

## Proposal

On every session-end path (user detach, remote process exit, peer hangup, ctx cancel), have the client write a small fixed "normalize the parent terminal" sequence before it exits — symmetric with the DEC modes the attach repaint can set. At minimum:

- show the cursor (`\x1b[?25h`),
- reset cursor style to default (DECSCUSR `\x1b[0 q`),
- leave the alternate-screen buffer (`\x1b[?1049l`).

## Acceptance criteria

- [ ] After detaching from a session whose remote side had hidden the cursor, the parent terminal shows a normal (blinking) cursor with no manual `reset`.
- [ ] After the remote process exits and the client exits, the parent terminal shows a normal cursor.
- [ ] If a full-screen (alt-screen) program was active at detach time, the parent terminal is returned to the main screen buffer.
- [ ] Output-side normalization runs on all teardown paths: detach keystroke, remote process exit, peer hangup, and ctx-cancel (SIGINT/SIGTERM).
- [ ] Restoration is unconditional — it does not depend on the remote side emitting a matching reset.

## Notes

- Distinct from #364 (input-side stdin non-blocking leak + termios restore). The two share the theme "client doesn't clean up the parent terminal on teardown" but have separate fixes; both should land for a session to leave the terminal fully usable.
- Key code: `internal/client/clientrunner/io.go` (reader copier 116, help line 108), `internal/client/clientrunner/lifecycle.go` (`Detach` 213-225, `Close` 162-203), and the modes the repaint can set in `internal/terminal/terminalrunner/terminal_screen.go:176-191` (`escEnterAltScreen`, `escHideCursor`/`escShowCursor`).
- Manual confirmation: attach while a TUI is running (or with the cursor hidden), detach, and observe the missing cursor / leftover alt-screen in the parent shell.

eminwux:
## Summary

Fixes the missing cursor and leftover alt-screen buffer left in the parent terminal after every attached-session teardown (issue #365): when a TUI or shell on the remote side hides the cursor (`\x1b[?25l`) or enters the alternate screen (`\x1b[?1049h`) — or when the server's own attach repaint emits those modes — none of it was being undone on the client side, and the parent shell was handed back a terminal with a hidden cursor / wrong screen buffer until a manual `reset`.

Distinct from #364: that fix was input-side (stdin nonblocking leak + termios raw→cooked). #364's `restoreParentTerminal()` already runs on every teardown path (detach keystroke, remote process exit, peer hangup, SIGINT/SIGTERM); this PR adds the output-side normalization to the same helper, so all four teardown paths benefit from the existing sync.Once discipline.

### Approach

- Define `escRestoreParentTerm = "\x1b[?1049l" + "\x1b[?25h" + "\x1b[0 q"` — leave the alternate-screen buffer, show the cursor, reset cursor style (DECSCUSR) to terminal default. Symmetric with the modes the server-side attach repaint can set (`escEnterAltScreen` / `escHideCursor` in `internal/terminal/terminalrunner/terminal_screen.go`).
- Write `escRestoreParentTerm` to `sr.stdout` inside `restoreParentTerminal`, **after the copier wait** (so we do not race the socket→stdout copier still draining remote bytes) and **before the termios restore** (escape bytes are interpreted by the parent terminal emulator, not by the line discipline, so termios mode doesn't affect them).
- Emit unconditionally per the AC — the client cannot rely on the remote side sending the matching reset on its own.

### Non-obvious calls (for reviewer scrutiny)

- Order is `?1049l` (leave alt-screen) → `?25h` (show cursor) → `0 q` (reset cursor style). Leaving alt-screen first means the show-cursor and DECSCUSR land on the main screen buffer the user is about to see, not on the alt-screen that's going away.
- The escape is inside the `sync.Once` body so it runs exactly once per session — same idempotency as the rest of `restoreParentTerminal`, matching the existing detach-then-close test (`TestRestore_Idempotent`).
- I deliberately did **not** issue a soft reset (RIS / `\x1bc`) — soft reset wipes the scrollback and resets colors, which would be visually destructive on every detach. The three-escape minimum named in #365 is the smallest sequence that solves the reported user-visible problems (missing cursor, stuck alt-screen).
- Substitutes `sr.stdout` with an `os.Pipe()` in the new tests rather than reading the master side of the existing pty pair, because `newAttachedExec` wires the socket→stdout copier with the original stdout — and the test cares about what `restoreParentTerminal` writes, not what the copier wrote earlier.

## Test plan

- [x] `make sbsh-sb` (canonical build) + ELF magic verified
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — all pass
- [x] `go test -race -count=3 -timeout=2m ./internal/client/clientrunner/...` — clean (covers the new tests under repeat + race)
- [x] `pre-commit run --files <changed>` (go-fmt, go-mod-tidy, golangci-lint, addlicense) — passes
- [x] New regression tests `TestRestore_EmitsOutputNormalize_OnProcessExit` and `TestRestore_EmitsOutputNormalize_OnDetach` substitute `sr.stdout` with an `os.Pipe`, drive each teardown path, and assert the three normalization escapes (`\x1b[?1049l`, `\x1b[?25h`, `\x1b[0 q`) appear in the captured output. Verified they fail against the pre-fix code (no escape sequence is written).
- [ ] Manual: attach to a session whose remote side hides the cursor or enters alt-screen, detach, and confirm the parent shell shows a normal blinking cursor on the main screen with no `reset` — not run (no interactive terminal in the dev sandbox); covered by the pipe-backed regression tests above.

Closes #365

### DIFF
--- a/internal/client/clientrunner/restore_linux_test.go
+++ b/internal/client/clientrunner/restore_linux_test.go
@@ -18,8 +18,10 @@ package clientrunner
 
 import (
 	"context"
+	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"syscall"
 	"testing"
@@ -158,6 +160,72 @@ func TestRestore_OnDetach(t *testing.T) {
 		t.Fatalf("Detach: %v", err)
 	}
 	assertRestored(t, tty)
+}
+
+// assertNormalizeEmitted reads everything still in r and checks that the
+// output-side normalization escape (leave alt-screen, show cursor, reset
+// cursor style) was written by restoreParentTerminal. The caller closes the
+// pipe writer before calling so ReadAll terminates on EOF.
+func assertNormalizeEmitted(t *testing.T, r *os.File) {
+	t.Helper()
+	buf, err := io.ReadAll(r)
+	if err != nil {
+		t.Fatalf("ReadAll: %v", err)
+	}
+	out := string(buf)
+	for _, want := range []string{"\x1b[?1049l", "\x1b[?25h", "\x1b[0 q"} {
+		if !strings.Contains(out, want) {
+			t.Errorf("missing normalization escape %q in stdout %q", want, out)
+		}
+	}
+}
+
+// TestRestore_EmitsOutputNormalize_OnProcessExit asserts that the
+// spawned-process-exit teardown path writes the output-side normalization
+// escape to the parent terminal: leave alt-screen, show cursor, reset cursor
+// style. Regression for issue #365.
+func TestRestore_EmitsOutputNormalize_OnProcessExit(t *testing.T) {
+	sr, _ := newAttachedExec(t)
+
+	r, w, errPipe := os.Pipe()
+	if errPipe != nil {
+		t.Fatalf("Pipe: %v", errPipe)
+	}
+	t.Cleanup(func() { _ = r.Close() })
+	sr.stdout = w
+
+	sockPath := filepath.Join(t.TempDir(), "ctrl.sock")
+	if errSeed := os.WriteFile(sockPath, []byte{}, 0o600); errSeed != nil {
+		t.Fatalf("seed socket file: %v", errSeed)
+	}
+	sr.metadata.Spec.SockerCtrl = sockPath
+
+	if errClose := sr.Close(nil); errClose != nil {
+		t.Fatalf("Close: %v", errClose)
+	}
+	_ = w.Close()
+	assertNormalizeEmitted(t, r)
+}
+
+// TestRestore_EmitsOutputNormalize_OnDetach asserts the user-detach teardown
+// path writes the output-side normalization escape, independently of the
+// Close()/EvError path. Regression for issue #365.
+func TestRestore_EmitsOutputNormalize_OnDetach(t *testing.T) {
+	sr, _ := newAttachedExec(t)
+	sr.terminalClient = &mockTerminalClient{}
+
+	r, w, errPipe := os.Pipe()
+	if errPipe != nil {
+		t.Fatalf("Pipe: %v", errPipe)
+	}
+	t.Cleanup(func() { _ = r.Close() })
+	sr.stdout = w
+
+	if errDetach := sr.Detach(); errDetach != nil {
+		t.Fatalf("Detach: %v", errDetach)
+	}
+	_ = w.Close()
+	assertNormalizeEmitted(t, r)
 }
 
 // TestRestore_Idempotent asserts the restore runs exactly once even when both
--- a/internal/client/clientrunner/terminal.go
+++ b/internal/client/clientrunner/terminal.go
@@ -71,6 +71,18 @@ func (sr *Exec) toExitShell() error {
 // uninterruptible stdin read (see restoreParentTerminal) cannot wedge teardown.
 const copierWaitTimeout = 500 * time.Millisecond
 
+// escRestoreParentTerm normalizes the parent terminal's output-side DEC private
+// mode state on session teardown: leave the alternate-screen buffer, show the
+// cursor, and reset the cursor style to the terminal default. Symmetric with
+// the modes the server-side attach repaint may set (escEnterAltScreen /
+// escHideCursor in internal/terminal/terminalrunner/terminal_screen.go) and
+// with anything a TUI/shell on the remote side wrote through to the parent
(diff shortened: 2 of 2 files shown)


eminwux:
## Review

**Verdict: LGTM — no blocking items.**

### Correctness

**Escape sequence content and order** ✓  
`\x1b[?1049l` (leave alt-screen) → `\x1b[?25h` (show cursor) → `\x1b[0 q` (DECSCUSR reset) is the right order: leaving alt-screen first ensures the show-cursor and DECSCUSR writes land on the main screen buffer the user is about to see, not the discarded alt-screen. All three sequences are correct for their stated purpose.

**Placement inside `restoreParentTerminal`** ✓  
After the copier wait (no race with socket→stdout draining remote bytes) and before `toExitShell` (escape bytes are interpreted by the terminal emulator, not the line discipline, so termios mode is irrelevant). Consistent with the existing ordering rationale in the surrounding code.

**`sr.stdout.WriteString`** ✓  
`sr.stdout` is `*os.File`, which implements `io.StringWriter`. The `nil` guard matches the existing `sr.stdin` pattern in the same function.

**Unconditional emission** ✓  
`?1049l` on a terminal that never entered alt-screen is a no-op; `?25h` and `0 q` are idempotent resets. Emitting unconditionally is both safe and the correct policy: the client cannot depend on the remote side having sent a matching reset.

**`sync.Once` integration** ✓  
The write is inside the existing `restoreOnce.Do` body, so it runs exactly once per session across all four teardown paths — matching `TestRestore_Idempotent`'s existing coverage.

### Tests

**Pipe-substitution approach** ✓  
`newAttachedExec` wires the original stdout to the socket→stdout copier; assigning `sr.stdout = w` after construction redirects what `restoreParentTerminal` writes without disturbing the copier setup. The pipe writer is closed before `assertNormalizeEmitted` drains it, so `io.ReadAll` terminates on EOF rather than blocking. Both the process-exit (`Close`) and user-detach (`Detach`) paths are covered independently.

**`assertNormalizeEmitted` helper** ✓  
Checks all three sequences individually; a missing sequence is called out with the captured output in the error message. Sound.

### Cross-reference check

`escRestoreParentTerm` used at `terminal.go:142`. `assertNormalizeEmitted` called from both new test functions. Referenced file `internal/terminal/terminalrunner/terminal_screen.go` exists on this branch and contains both `escEnterAltScreen` and `escHideCursor` as described in the constant's comment.

### Per-commit signature check

Single commit `51761e8` — GPG-verified ✓

---

No nits.

PR #371, opened 2026-05-30
fix: decouple flaky test from absolute wall-clock bound

### LINKED ISSUE #368: cmd/sb/stop: flaky Test_StopViaRPC_StaleSocket_RespectsTimeout under parallel load
## Repro

On clean `origin/main` (worktree at the current tip), under repeated/parallel load:

```bash
go test -count=5 ./cmd/sb/stop/...
```

Fails intermittently (also surfaces when the full suite runs the package alongside others). `go test ./cmd/sb/stop/...` alone usually passes — the flake needs scheduling contention.

## Observed

```
WARN dial failed attempt=0 error="dial unix /tmp/.../controller.sock: connect: connection refused"
WARN context done before retry attempt=1 error="context deadline exceeded"
FAIL
FAIL	github.com/eminwux/sbsh/cmd/sb/stop	9.006s
```

The failing assertion is the wall-clock bound in `Test_StopViaRPC_StaleSocket_RespectsTimeout` (`cmd/sb/stop/terminals_test.go:257`):

```go
timeout := 100 * time.Millisecond
...
if elapsed > 500*time.Millisecond {
    t.Fatalf("stopViaRPC took %v with timeout=%v; expected <500ms ...", elapsed, timeout)
}
```

`stopViaRPC` *does* honor the bounded ctx (the "context done before retry" log confirms the 100ms deadline fired), but the measured `elapsed` exceeds the 500ms ceiling under host scheduling load — scheduler latency between the deadline firing and the goroutine resuming inflates the wall-clock past the assertion bound.

## Root cause

Not a logic regression — `stopViaRPC`'s timeout behavior is correct. The test couples a correctness assertion (bounded retry) to an absolute wall-clock budget (500ms) that is sensitive to CI/host load. With `timeout=100ms` and ~600ms of retry-delay budget, a 500ms wall-clock cap leaves too little slack once the runner is contended (parallel packages, `-count=N`).

## Suggested fix

Decouple the assertion from absolute wall-clock. Options, roughly in order of robustness:
- Assert the *relative* bound — `elapsed` should be within a generous multiple of `timeout` (e.g. `< 10*timeout`) rather than a fixed 500ms — and/or raise the ceiling well above the ~600ms retry-delay budget.
- Better: assert on the *behavior* (an error is returned and it is the deadline/ECONNREFUSED class) and drop the wall-clock comparison, or gate the timing check behind a non-`-short`, non-parallel run.

Distinct from #366 (`Test_RunStopTerminals_AllAggregation`, a trap-install race) — same package, different test and different cause; surfaced alongside #366's flakiness during validation for #364.

eminwux:
## Summary

Fixes flaky test `Test_StopViaRPC_StaleSocket_RespectsTimeout` by replacing the absolute 500ms wall-clock ceiling with a relative bound (10x the configured timeout).

The original test used a fixed 500ms ceiling that was too strict under CI/host scheduling load. With a 100ms timeout and ~600ms of retry-delay budget, the 500ms wall-clock cap left too little slack when the runner is contended (parallel packages, `-count=N`).

The fix uses `elapsed > 10*timeout` (1s with the 100ms test timeout) which:
- Gives generous slack for scheduler variance under load
- Remains well below the pre-fix ceiling (~9.6s)
- Catches unbounded retry regressions
- Decouples the assertion from absolute wall-clock time

Closes #368

## Test plan

- [x] `go test -count=5 ./cmd/sb/stop/...` passes (repro from the issue)
- [x] `make sbsh-sb` builds successfully
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./cmd/sb/stop/...` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/sb/stop/terminals_test.go
+++ b/cmd/sb/stop/terminals_test.go
@@ -252,10 +252,12 @@ func Test_StopViaRPC_StaleSocket_RespectsTimeout(t *testing.T) {
 	if rpcErr == nil {
 		t.Fatal("stopViaRPC against a stale socket: expected error, got nil")
 	}
-	// Generous slack for CI variance but well below the retry-delay budget
-	// (~600ms) and orders of magnitude below the pre-fix 9.6s ceiling.
-	if elapsed > 500*time.Millisecond {
-		t.Fatalf("stopViaRPC took %v with timeout=%v; expected <500ms (pre-fix: 0.6s-9.6s)", elapsed, timeout)
+	// Relative bound decouples the assertion from absolute wall-clock under
+	// host/CI load; 10x timeout (1s with 100ms timeout) gives generous slack
+	// while remaining well below the pre-fix ceiling (~9.6s) and catching
+	// unbounded retry regressions.
+	if elapsed > 10*timeout {
+		t.Fatalf("stopViaRPC took %v with timeout=%v; expected <10x timeout (pre-fix: ~9.6s)", elapsed, timeout)
 	}
 }
 


eminwux:
## Review

**Verdict: LGTM**

### Summary

Clean, targeted fix for `Test_StopViaRPC_StaleSocket_RespectsTimeout`. The relative bound `10*timeout` (1 s with the 100 ms test timeout) is well-reasoned:

- `timeout` is defined in the same function scope, so the expression is correctly typed and compiles without new imports.
- Doubles the effective ceiling (500 ms → 1 s) while remaining well below the pre-fix worst-case (~9.6 s), giving enough scheduler slack under CI load without masking genuine retry-budget regressions.
- The removed `500*time.Millisecond` literal had no other references in the package.

### Checklist

- [x] Correctness — relative bound correctly scoped; `t.Fatalf` message updated to match
- [x] Sensitive data — test file only
- [x] Dead code — none; `time.Millisecond` import still used elsewhere in the file
- [x] Conventions — build/test conventions followed; commit signed and verified
- [x] Test plan — `-count=5` is the right repro for a flakiness fix; `make sbsh-sb` + `go vet` also confirmed

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

Issue #345, opened 2026-05-22
terminalrunner: close-time metadata write fails permission-denied when create/close run under different uids


eminwux:
## Context

When sbsh is embedded as an in-container terminal manager (here: kukeon's `kuketty` wrapper, cell `sbsh-pm-0`), the close-time metadata update fails with permission denied:

```
WARN  "failed to update metadata on close" id=58ac0c22 err=write /run/kukeon/tty/terminals/58ac0c22/metadata.json: open /run/kukeon/tty/terminals/58ac0c22/.meta-3372523117.tmp: permission denied
ERROR "failed to update terminal state" id=58ac0c22 err=...permission denied
... (repeats on client cleanup)
```

Code path:
- `internal/terminal/terminalrunner/metadata.go:38` creates the terminal dir with `os.MkdirAll(dir, 0o700)` — single-owner, no group write.
- `updateMetadata()` (`metadata.go:80`) → `shared.WriteMetadata(ctx, metadataCopy, dir)` writes a `.meta-*.tmp` temp file in `dir` then renames.
- `updateTerminalState()` / `updateTerminalAttachers()` (`metadata.go:94,117`) invoke `updateMetadata` on close and log `"failed to update metadata on close"`.

Because the dir is created `0o700`, only the creating uid can write temp files in it. When the terminal is created under one uid and the close path runs under another (the privilege transition common in an embedded/sandboxed runtime — e.g. setup as root, in-cell process as a non-root container uid), the close-time write is denied. The error also repeats (close, client cleanup), spamming logs.

## Proposal

Make the close-time metadata write robust to this embedding, e.g.:
- create the terminal dir with a mode/group that allows the runtime's effective uid to write across create+close (group-writable under a known gid), and/or
- detect the un-writable case and surface a single actionable error instead of repeated WARN/ERROR on every close/cleanup hook, and/or
- document the uid/ownership contract embedders must satisfy for `RunPath/terminals/<id>`.

## Acceptance criteria
- [ ] Terminal close updates metadata successfully when create and close run under different uids (or fails once with a clear, actionable error rather than repeated permission-denied spam).
- [ ] The ownership/mode contract for `RunPath/terminals/<id>` is documented for embedders.
- [ ] Test covering the close path when the terminal dir is not writable by the closing uid.

## Notes
- Observed in sbsh v0.11.1, embedded by kukeon.
- Downstream/paired issue (kukeon-side tty dir ownership setup): eminwux/kukeon#672
[...]

PR #374, opened 2026-06-03
fix: never strand terminal state in-memory on metadata write failure

### LINKED ISSUE #373: terminalrunner: failed metadata write in PostAttachShell strands terminal in non-Ready state forever (no recovery after ENOSPC clears)
## Summary

A failed metadata write during `PostAttachShell` permanently strands a terminal in a non-`Ready` state. Once it happens, **every subsequent attach hangs until the client's context deadline** (`context done before ready: context deadline exceeded`) — even though the underlying PTY is alive and serving, and even after the original failure condition (e.g. a full disk) has cleared. Only restarting the kuketty/terminal process recovers it.

Observed in `sbsh v0.12.1` embedded as `kuketty` in kukeon. Trigger in the wild: the terminal dir (`/run/kukeon/tty`, bind-mounted from an ext4 volume) filled up (`ENOSPC`), an attach landed at that moment, and the cell's attach plane never recovered after disk space was freed.

## Root cause

`PostAttachShell` (`internal/terminal/terminalrunner/lifecycle.go:515`), run on every attach via `Controller.Attach` (`internal/terminal/controller.go:322`):

```go
for sr.metadata.Status.State != api.Ready {        // (1) spin until Ready
    select {
    case <-sr.ctx.Done(): return sr.ctx.Err()
    case <-time.After(waitReadyPeriod):
    }
}
if err := sr.updateTerminalState(api.PostAttach); err != nil {   // (2)
    return fmt.Errorf("failed to update terminal state on post attach: %w", err)
}
...
if err := sr.updateTerminalState(api.Ready); err != nil { ... }  // (3) restores Ready
```

`updateTerminalState` (`internal/terminal/terminalrunner/metadata.go:161`) mutates **in-memory state first, then persists, with no rollback on failure**:

```go
sr.metadataMu.Lock()
sr.metadata.Status.State = status   // in-memory: now PostAttach
sr.metadataMu.Unlock()
if err := sr.updateMetadata(); err != nil {   // disk write fails (ENOSPC)
    sr.noteMetadataWriteErr("updateTerminalState", err)
    return err                                // returns WITHOUT reverting State
}
```

Failure sequence:

1. Attach: loop `(1)` exits because `State == Ready`.
2. `updateTerminalState(api.PostAttach)` flips **in-memory** `Status.State = PostAttach`, then the disk write hits `ENOSPC` and returns the error. Step `(3)` — which would restore `Ready` — never runs.
3. `Status.State` is now **permanently `PostAttach`**. Every future attach re-enters the `for State != Ready` loop, which can never exit, and spins until the client's context deadline → `context done before ready: context deadline exceeded`.

The disk-full is incidental — it's just the trigger. **Any** error returned from `updateTerminalState(api.PostAttach)` strands the state machine the same way. There is no retry and no path back to `Ready` short of recreating the runner.

## Observed log (real incident)

```
INFO  PostAttachShell: waiting for terminal to be ready id=d6bb3fa8
INFO  PostAttachShell: updating terminal state to post attach id=d6bb3fa8
WARN  failed to update metadata on close ... where=updateTerminalState err=... no space left on device
ERROR failed to update terminal state on post attach id=d6bb3fa8 err=... no space left on device
```

After disk space was freed, the PTY (with a live shell session) kept running, but every new attach failed with:

```
attach ping timed out: kuketty control socket bound but not yet serving:
error attaching client: context done before ready: context deadline exceeded
```

## Minimal repro

1. Bring up a terminal so its state reaches `Ready`.
2. Force `updateMetadata()` to fail during `PostAttachShell` (fill the terminal dir / make it read-only / inject a write error).
3. Attach — it fails (expected).
4. Restore writes (free space / restore perms).
5. Attach again → hangs until ctx deadline; terminal never serves again without a process restart.

A unit test can drive this directly: set `State = Ready`, stub `updateMetadata` to error once on the `PostAttach` write, call `PostAttachShell`, then assert that a second `PostAttachShell` with a healthy writer still reaches `Ready` (currently it never does, because `State` is stuck at `PostAttach`).

## Proposed fixes (either resolves the non-recovery; ideally both)

1. **Make the attach-time metadata write best-effort.** `metadata.json` is an observability/reconnect artifact; failing to *persist* it should not abort an attach to a PTY that is alive in memory. This mirrors the existing `noteMetadataWriteErr` / `metadataWriteDeniedOnce` log-and-continue handling already used for permission-denied (cf. #345). PostAttach/Ready transitions should log-and-continue rather than return fatally.

2. **Never strand in-memory state on a persistence failure.** `updateTerminalState` should either roll back `Status.State` to its prior value on write error, or set state only *after* a successful write — so a failed write can't leave the terminal permanently non-`Ready`.

Fix #1 keeps attach working through transient disk/perm failures; fix #2 hardens the state machine so no future caller can strand it.

## Environment

- sbsh `v0.12.1` (as `kuketty`, embedded in kukeon)
- Terminal dir on ext4, bind-mounted, hit 100% full
- Related: #345 (close-time metadata write permission-denied) — same metadata-write-robustness theme, different trigger/site

eminwux:
## Summary

- `updateTerminalState` now writes to disk *first* and commits in-memory only on success, so a transient persistence failure (ENOSPC, perm-denied) can no longer leave `Status.State` stranded at an intermediate value. Pre-fix, a failed `PostAttach` write left in-memory `State = PostAttach` forever, and `PostAttachShell`'s `for State != Ready` spin then deadlocked every future attach until the runner was recreated.
- `PostAttachShell` demotes the `updateTerminalState` returns to log-and-continue: `metadata.json` is an observability/reconnect artifact, not the source of truth for the live PTY, and aborting attach on a persistence failure leaves a healthy terminal unreachable. Same shape as the close-time `noteMetadataWriteErr` log-and-continue pattern from #345.

Both fixes apply (the issue notes "ideally both"): fix #2 — the rollback / write-first ordering — is the foundational fix that hardens the state machine against any future caller; fix #1 keeps attach working through transient disk/perm failures.

## Test plan

- [x] `go test ./internal/terminal/terminalrunner/ -run 'TestUpdateTerminalState|TestPostAttachShell|TestShellStateUpdateError|TestUpdateMetadata_PermissionDeniedLoggedOnce|TestUpdateTerminalAttachers' -v -count=1` — including the two new tests:
  - `TestUpdateTerminalState_WriteFailure_DoesNotStrandInMemory` — drives a Ready→PostAttach write under a non-writable dir, asserts in-memory state stays Ready, then restores writability and confirms the next call advances.
  - `TestPostAttachShell_RecoversAfterTransientWriteFailure` — drives the issue's reproducer: `PostAttachShell` under a non-writable dir returns nil (best-effort), state stays Ready, restored dir, second `PostAttachShell` reaches Ready.
- [x] `make sbsh-sb` — canonical build (`sbsh` + `sb` hardlink).
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./... -count=1` — one unrelated pre-existing flake in `pkg/spawn` (`TestGracefulShutdown_SIGKILLEscalation`, `text file busy` on the trap.sh exec) — re-run passes; same test category as the recent flake-fix #371; not touched by this PR (pkg/spawn unchanged).

Closes #373

### DIFF
--- a/internal/terminal/terminalrunner/capture_metadata_cluster_test.go
+++ b/internal/terminal/terminalrunner/capture_metadata_cluster_test.go
@@ -229,6 +229,54 @@ func TestUpdateTerminalState_WriteFailure(t *testing.T) {
 	}
 }
 
+// updateTerminalState's write-first-then-commit ordering must not strand
+// Status.State at an intermediate value when the disk write fails: a
+// stranded PostAttach value would deadlock PostAttachShell's
+// `for State != Ready` spin on every subsequent attach until the runner
+// is recreated. See #373.
+func TestUpdateTerminalState_WriteFailure_DoesNotStrandInMemory(t *testing.T) {
+	if os.Geteuid() == 0 {
+		t.Skip("root bypasses directory DAC, so a non-writable dir cannot deny the write")
+	}
+
+	runPath := t.TempDir()
+	id := api.ID("term-strand")
+	dir := filepath.Join(runPath, defaults.TerminalsRunPath, string(id))
+	if err := os.MkdirAll(dir, 0o700); err != nil {
+		t.Fatalf("mkdir terminal dir: %v", err)
+	}
+	t.Cleanup(func() { _ = os.Chmod(dir, 0o700) })
+
+	sr := newMetadataExec(t, runPath, "", id)
+	// Reach Ready while the dir is still writable.
+	if err := sr.updateTerminalState(api.Ready); err != nil {
+		t.Fatalf("updateTerminalState(Ready) preflight: %v", err)
+	}
+
+	// Now make the next write fail.
+	if err := os.Chmod(dir, 0o500); err != nil {
+		t.Fatalf("chmod terminal dir: %v", err)
+	}
+
+	if err := sr.updateTerminalState(api.PostAttach); err == nil {
+		t.Fatal("updateTerminalState(PostAttach): want write error on un-writable dir, got nil")
+	}
+	if doc, _ := sr.Metadata(); doc.Status.State != api.Ready {
+		t.Fatalf("in-memory state stranded at %v on write failure; want unchanged Ready", doc.Status.State)
+	}
+
+	// Restore writability — subsequent updates must still advance.
+	if err := os.Chmod(dir, 0o700); err != nil {
+		t.Fatalf("restore chmod: %v", err)
+	}
+	if err := sr.updateTerminalState(api.PostAttach); err != nil {
+		t.Fatalf("updateTerminalState(PostAttach) after restore: %v", err)
+	}
+	if doc, _ := sr.Metadata(); doc.Status.State != api.PostAttach {
+		t.Fatalf("state after successful retry = %v, want PostAttach", doc.Status.State)
+	}
+}
+
 // When create and close run under different uids the terminal dir is not
 // writable by the closing uid, so every close/cleanup hook's metadata write is
 // denied. The close path must surface this once, actionably, instead of
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -527,9 +527,15 @@ func (sr *Exec) PostAttachShell() error {
 
 	sr.logger.Info("PostAttachShell: updating terminal state to post attach", "id", sr.id)
 
+	// Attach-time metadata writes are best-effort: metadata.json is an
+	// observability/reconnect artifact, not the source of truth for the
+	// live PTY, and aborting attach on a persistence failure (ENOSPC,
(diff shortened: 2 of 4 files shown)


eminwux:
Re-running failed CI jobs once. Will revisit when they finish; if they fail again the PR will be marked `ci-stalled`.

eminwux:
**PR #374 Review — `fix: never strand terminal state in-memory on metadata write failure`**

LGTM — write-first / commit-on-success in `updateTerminalState` makes the in-memory state un-strandable for every caller (the foundational fix), `PostAttachShell`'s log-and-continue keeps attach working through transient disk/perm failures, and both new tests drive the issue-#373 reproducer end-to-end.

PR #377, opened 2026-06-03
test: write pkg/spawn sleeper once in TestMain to de-flake ETXTBSY

### LINKED ISSUE #376: test: pkg/spawn sleeperBinary races fork+exec with WriteFile under full-suite parallel load (ETXTBSY)
## Repro

`origin/main` at 01477fd (also reproduces on the branch for #372, unrelated to that diff — see #375).

```bash
go test -timeout=5m $(go list ./... | grep -v '/e2e$') 2>&1 | grep -E "FAIL|text file"
```

Two tests in `pkg/spawn` flake intermittently under the CI shape's full-suite parallelism (`go test ./...` with many packages running concurrently):

- `TestClientHandle_WaitReady_Timeout`
- `TestClientHandle_Close_NoSocketEscalates`

Both pass reliably when run in isolation (`go test ./pkg/spawn/... -run TestClientHandle_WaitReady_Timeout -count=3` is green even on a stressed host); the flake only surfaces when many sibling packages are forking children in parallel.

## Observed

```
--- FAIL: TestClientHandle_WaitReady_Timeout (0.00s)
    lifecycle_test.go:125: NewClient: start client process: fork/exec /tmp/TestClientHandle_WaitReady_Timeout1870295205/002/sleeper.sh: text file busy
--- FAIL: TestClientHandle_Close_NoSocketEscalates (0.00s)
    lifecycle_test.go:64: NewClient: start client process: fork/exec /tmp/TestClientHandle_Close_NoSocketEscalates2005805253/002/sleeper.sh: text file busy
FAIL
FAIL	github.com/eminwux/sbsh/pkg/spawn	2.164s
```

Each test uses its own `t.TempDir()`, so the `sleeper.sh` path is unique per test — this isn't a path collision between concurrent goroutines. The failure is `ETXTBSY` on `execve` of a file that the same goroutine just wrote and closed via `os.WriteFile` a few statements earlier.

## Root cause

`pkg/spawn/lifecycle_test.go:42-49`:

```go
func sleeperBinary(t *testing.T) string {
    t.Helper()
    path := filepath.Join(t.TempDir(), "sleeper.sh")
    if err := os.WriteFile(path, []byte("#!/bin/sh\nexec sleep 30\n"), 0o755); err != nil {
        t.Fatalf("write sleeper: %v", err)
    }
    return path
}
```

This is the known Linux "ETXTBSY-after-recent-write" race (cf. golang/go#22315, runc, bazel). The kernel returns `ETXTBSY` from `execve` when *any* file descriptor referencing the inode is still open for writing **anywhere in the process group**. `os.WriteFile` closes its own FD synchronously, but every parallel test in `lifecycle_test.go` (10+ tests, each calling `t.Parallel()`) is a parent goroutine inside the same `go test` process that just `exec`s a sibling fork. Under high concurrency, Go's runtime briefly retains write FDs from sibling goroutines' `os.WriteFile` calls during the fork window — when one sibling's fork sees another sibling's still-open write FD, `execve` returns `ETXTBSY`.

The reason isolation reproduces clean: with only one test running, no sibling write FD is open at the moment of fork. The reason full-suite parallelism reproduces it: dozens of `sleeperBinary` writes can be in-flight across `pkg/spawn`'s ~10 parallel tests AND `pkg/spawn/process_test.go`'s direct `/bin/sleep` execs, multiplying the fork-vs-sibling-write collision window.

## Suggested fix

The cleanest fix is `O_CLOEXEC` on the WriteFile path (Go's `os.WriteFile` already uses `O_CLOEXEC` since Go 1.20+, so this is already in place — confirming the issue is the *sibling* FD race, not the writer's own FD). The well-trodden workarounds for this exact race pattern:

1. **Switch sibling-execs to `/bin/sh` with `-c`**, passing the script body as a string argument — sidesteps the executable-mapping entirely. Simplest change.
2. **Write `sleeper.sh` once in `TestMain`** to a single package-shared temp path; every test reads its path. No per-test write means no concurrent-write window. Slightly more invasive (callers re-shape to take a path rather than a `*testing.T`).
3. **Retry-on-ETXTBSY with brief backoff** at the spawn site, similar to runc's exec retry loop. Works but masks the race rather than removing it; least preferred.

Sibling de-flake commits in `pkg/terminal/terminalrunner` (#371) and `cmd/sb/stop` (#369, #366) suggest the project is actively grooming this class of races; this `pkg/spawn` flake is on the same family.

eminwux:
## Summary

- The issue lists three fix options. I'm picking **option 2** (write once in `TestMain`) rather than the "simplest" option 1 (`/bin/sh -c`). Reason: `buildClientAttachArgs` unconditionally prepends `--run-path <X>` before `ExtraArgs`, so the `/bin/sh -c "exec sleep 30" sh` shape only works if every caller also clears `RunPath` on the doc — option 2 keeps the fix entirely inside the test helper and the test setup. Happy to swap to option 1 if you prefer.
- `TestMain` writes `sleeper.sh` once to a package-shared temp path before any test runs (no `t.Parallel()` window during the write); all five callers of `sleeperBinary(t)` now read that path with no per-test `os.WriteFile`. Eliminates the sibling-goroutine writer-FD overlap that the issue diagnoses as the ETXTBSY trigger.
- `sleeperBinary(t)`'s signature is preserved — the helper just returns the shared path — so the five call sites are unchanged.
- Option 3 (retry-on-ETXTBSY) was deliberately not picked, per the issue's own "masks the race rather than removing it" note.

## Out of scope

- `TestGracefulShutdown_SIGKILLEscalation` (`pkg/spawn/lifecycle_test.go:248`) has the same write-then-exec pattern with its own `trap.sh`. The issue does not name it as flaking — it writes once (vs. five `sleeperBinary` callers), so the collision window is much smaller — and I deliberately left it alone to keep the PR scoped to the named tests per `dev/CLAUDE.md`'s no-unrelated-cleanups rule. If it surfaces in CI later, the same TestMain-write-once pattern applies.

## Test plan

- [x] `make sbsh-sb` produces an ELF (`od -c`: `\177 E L F`)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/spawn/...` (clean)
- [x] `go test -count=20 -run 'TestClientHandle_WaitReady_Timeout|TestClientHandle_Close_NoSocketEscalates|TestTerminalHandle_WaitReady_Timeout|TestTerminalHandle_Close_NoSocketEscalates|TestClientHandle_WaitClose_ContextCanceled' ./pkg/spawn/...` (20 iterations, all green)
- [x] The issue's exact CI-shape repro: `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` run **three times back-to-back** with `-count=3` on the third run — no `FAIL`, no `text file busy` across all runs

Closes #376

### DIFF
--- a/pkg/spawn/lifecycle_test.go
+++ b/pkg/spawn/lifecycle_test.go
@@ -22,6 +22,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"log/slog"
 	"os"
@@ -35,17 +36,36 @@ import (
 	"github.com/eminwux/sbsh/pkg/api"
 )
 
-// sleeperBinary writes an executable shell script that sleeps regardless of
-// the argv the handle constructs (NewClient/NewTerminal append their own
-// attach/terminal args, so a real /bin/sleep would reject them and exit). It
-// stands in for a long-lived sb child whose control socket never appears.
+// sharedSleeperPath is the package-shared path to a sleeper.sh script that
+// sleeps regardless of the argv the handle constructs (NewClient/NewTerminal
+// append their own attach/terminal args, so a real /bin/sleep would reject
+// them and exit). Written once in TestMain before any test runs; see #376 —
+// per-test os.WriteFile of an executable then fork+exec from sibling parallel
+// goroutines races on the kernel's writer-FD check and surfaces as ETXTBSY.
+var sharedSleeperPath string
+
+func TestMain(m *testing.M) {
+	dir, err := os.MkdirTemp("", "spawn-test-sleeper-")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "spawn tests: create sleeper tempdir: %v\n", err)
+		os.Exit(1)
+	}
+	sharedSleeperPath = filepath.Join(dir, "sleeper.sh")
+	if err := os.WriteFile(sharedSleeperPath, []byte("#!/bin/sh\nexec sleep 30\n"), 0o755); err != nil {
+		fmt.Fprintf(os.Stderr, "spawn tests: write sleeper: %v\n", err)
+		_ = os.RemoveAll(dir)
+		os.Exit(1)
+	}
+	code := m.Run()
+	_ = os.RemoveAll(dir)
+	os.Exit(code)
+}
+
+// sleeperBinary returns the package-shared sleeper.sh path written once in
+// TestMain. See sharedSleeperPath's docstring for the rationale.
 func sleeperBinary(t *testing.T) string {
 	t.Helper()
-	path := filepath.Join(t.TempDir(), "sleeper.sh")
-	if err := os.WriteFile(path, []byte("#!/bin/sh\nexec sleep 30\n"), 0o755); err != nil {
-		t.Fatalf("write sleeper: %v", err)
-	}
-	return path
+	return sharedSleeperPath
 }
 
 // TestClientHandle_Close_NoSocketEscalates spawns a long-lived child whose


eminwux:
**PR #377 Review — `test: write pkg/spawn sleeper once in TestMain to de-flake ETXTBSY`**

LGTM — moves the per-test `os.WriteFile` into `TestMain` so the sleeper is on disk before any `t.Parallel()` goroutine forks, eliminating the writer-FD/exec race named in #376; `sleeperBinary(t)`'s signature is preserved across all five callers, error paths in `TestMain` clean up the tempdir, and the deferred `trap.sh` site is transparently called out as out-of-scope.

PR #381, opened 2026-06-05
fix: swallow clean session-end sentinels in sbsh runClient

### LINKED ISSUE #380: bug: sbsh prints spurious 'Error: child routine exited: …' on successful operator detach
## Summary

After a successful operator-initiated detach (`Ctrl+]` twice) from an `sbsh` session, the CLI prints the intended yellow `Detached` confirmation **followed by a cobra error line** for what is in fact a clean exit:

```
Detached
Error: child routine exited: close requested: client detached: read/write routines exited
$
```

The `sb attach` path does *not* exhibit this — it swallows the same error chain inside `cmd/sb/attach/attach.go:222–235`. Only `sbsh` (which consumes the controller directly, without going through `pkg/attach.Run`) leaks the chain to stderr and returns non-zero from cobra.

## Repro

1. Build canonical target: `make sbsh-sb`.
2. Start a new session: `./sbsh` (defaults are fine).
3. At the shell prompt, press `Ctrl+]` twice.
4. Observe stderr: the yellow `Detached` line is immediately followed by `Error: child routine exited: close requested: client detached: read/write routines exited`.

Reproduces on `main` at `f95185c`.

## Observed error chain (innermost → outermost)

1. `internal/client/clientrunner/io.go:131` — dual-copier exit emits `EvError{Err: errors.New("read/write routines exited")}` on socket close (including the clean detach path).
2. `internal/client/controller.go:403` — `handleEvent` on `EvError` calls `classifySessionEnd(ev.Err)`. Because `Detach()` set `detachRequested=true` at `controller.go:455`, the wrap becomes `errdefs.ErrClientDetached: read/write routines exited`.
3. `controller.go:417` — `onClosed` → `Close(err)` pushes the error onto `closeReqCh`.
4. `controller.go:312` — `Run`'s main loop reads `closeReqCh` and wraps with `errdefs.ErrCloseReq`: `close requested: client detached: read/write routines exited`.
5. `cmd/sbsh/sbsh.go:525–535` (`runClient`'s `case err := <-errCh` arm) — sees non-nil, non-`context.Canceled` error and wraps with `errdefs.ErrChildExit`: `child routine exited: close requested: client detached: read/write routines exited`. Returns to cobra.
6. Cobra prints `Error: …` because `RunE` returned non-nil.

## Root cause

`cmd/sbsh/sbsh.go:runClient` treats every non-`context.Canceled` controller exit as a failure. It has no special case for `errdefs.ErrClientDetached`, even though the controller stack tags the cause with that sentinel precisely so callers can distinguish an operator detach from a real failure.

Compare `pkg/attach/attach.go:194–200` (`classifyError`), which already remaps `errdefs.ErrClientDetached` → public `attach.ErrDetached`, and `cmd/sb/attach/attach.go:222`, which returns nil for the detach path ("Other errors are intentionally swallowed to avoid polluting the terminal — preserves prior `sb attach` UX"). `sbsh` bypasses both layers.

## Suggested fix

In `cmd/sbsh/sbsh.go:runClient`'s `case err := <-errCh` arm, treat `errors.Is(err, errdefs.ErrClientDetached)` as a clean exit (return nil after `WaitClose`), alongside the existing `context.Canceled` check. The likely-symmetric path `errors.Is(err, errdefs.ErrPeerClosed)` (clean shell exit) should also be considered — if a `bash` session exits with `exit 0`, the same chain is plausible and the user shouldn't see `Error: …` for a normal logout either, but that's a separate observation worth verifying as part of the fix.

## Acceptance criteria

- [ ] `sbsh` → `Ctrl+]` `Ctrl+]` exits cleanly: only the yellow `Detached` line, no trailing `Error: child routine exited: …`, exit code 0.
- [ ] `errdefs.ErrClientDetached` (and the chained `ErrCloseReq` wrapper from `Controller.Run`) is treated as a clean exit by `cmd/sbsh/sbsh.go:runClient`.
- [ ] Regression test in `cmd/sbsh/` covers `runClient` returning nil when the controller returns an error chain that wraps `errdefs.ErrClientDetached`.
- [ ] No regression in genuine failure surfacing — RPC server crash, terminal start failure, etc. still surface as cobra `Error: …`.

eminwux:
## Summary

- `cmd/sbsh/sbsh.go:runClient` now treats `errdefs.ErrClientDetached` and `errdefs.ErrPeerClosed` as clean session-end sentinels alongside `context.Canceled`. Without this, the controller's `ErrCloseReq: ErrClientDetached: read/write routines exited` chain reaches cobra and prints `Error: child routine exited: …` immediately after the yellow `Detached` line on a successful operator detach (`Ctrl+]` `Ctrl+]`). Mirrors the swallow-on-clean-exit behavior `pkg/attach.Run` already provides via `classifySessionEnd` → `attach.ErrDetached` / `attach.ErrPeerClosed` and `cmd/sb/attach/attach.go:222` consumes; `sbsh` consumes the controller directly so the swallow has to live in `runClient`.
- Genuine failures (`ErrRPCServerExited`, `ErrStartTerminal`, `ErrWaitOnReady`, `ErrChildExit`-wrapped runtime errors, …) still surface as cobra `Error: …` — the existing `TestRunTerminal_ErrChildExit`/`ErrChildExitWithWaitCloseError`/`ErrWaitOnReady` tests keep covering that arm.

### `ErrPeerClosed` inclusion

The issue's "Suggested fix" calls out `ErrPeerClosed` as a "separate observation worth verifying as part of the fix" — the controller produces the same chain on a clean `exit 0` from the shell (`Controller.classifySessionEnd` wraps with `ErrPeerClosed` when `detachRequested` is unset). I verified that path by reading `internal/client/controller.go:467-476` and included it as a lower-blast-radius default: this matches the symmetric behavior `cmd/sb/attach` and `pkg/attach.Run` already provide, and AC item 4's "RPC server crash, terminal start failure, etc." enumeration does not include either sentinel. Reviewer can push back if narrower scope (only `ErrClientDetached`) is preferred — drop the `ErrPeerClosed` clause from `isCleanSessionEnd` and the corresponding test.

## Test plan

- [x] `make sbsh-sb` — canonical build produces ELF binary + hardlink (verified via ELF magic).
- [x] `go vet ./...` — clean.
- [x] `make test` — full CI test suite (`cmd/sb...`, `cmd/sbsh...`, `internal/terminal...`, `internal/client...`, integration-tagged `cmd/sb/get/...`, `e2e`) passes locally with `EXIT=0`.
- [x] `go test -run 'TestRunTerminal_CleanDetach|TestRunTerminal_CleanPeerClosed' -v ./cmd/sbsh` — new regression tests pass.
- [x] `pre-commit run --files cmd/sbsh/sbsh.go cmd/sbsh/sbsh_test.go` — clean (go-fmt, golangci-lint, addlicense, etc.).
- [ ] Live `./sbsh` → `Ctrl+]` `Ctrl+]` interactive smoke — dev host has no interactive TTY; the unit tests construct the exact error chain `Controller.Run` produces on operator detach (`ErrCloseReq` wrapping `ErrClientDetached` wrapping `read/write routines exited`) and assert `runClient` returns nil.

## Acceptance criteria

- [ ] `sbsh` → `Ctrl+]` `Ctrl+]` exits cleanly: only the yellow `Detached` line, no trailing `Error: child routine exited: …`, exit code 0. — *Verified via the regression test that drives the documented error chain through `runClient`; the live TTY-interactive smoke is left unticked because the dev host has no interactive TTY (see "Test plan" above).*
- [x] `errdefs.ErrClientDetached` (and the chained `ErrCloseReq` wrapper from `Controller.Run`) is treated as a clean exit by `cmd/sbsh/sbsh.go:runClient` — implemented via `isCleanSessionEnd` in `cmd/sbsh/sbsh.go`.
- [x] Regression test in `cmd/sbsh/` covers `runClient` returning nil when the controller returns an error chain that wraps `errdefs.ErrClientDetached` — `TestRunTerminal_CleanDetach` in `cmd/sbsh/sbsh_test.go` (and symmetric `TestRunTerminal_CleanPeerClosed`).
- [x] No regression in genuine failure surfacing — RPC server crash, terminal start failure, etc. still surface as cobra `Error: …`. `isCleanSessionEnd` only matches `context.Canceled`, `ErrClientDetached`, `ErrPeerClosed`; the pre-existing `TestRunTerminal_ErrChildExit`, `TestRunTerminal_ErrChildExitWithWaitCloseError`, and `TestRunTerminal_ErrWaitOnReady` tests still pass and continue to enforce the failure-surfacing path.

Closes #380

### DIFF
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -524,7 +524,7 @@ func runClient(
 
 	case err := <-errCh:
 		logger.DebugContext(ctx, "controller stopped", "error", err)
-		if err != nil && !errors.Is(err, context.Canceled) {
+		if err != nil && !isCleanSessionEnd(err) {
 			err = fmt.Errorf("%w: %w", errdefs.ErrChildExit, err)
 			if errC := ctrl.WaitClose(); errC != nil {
 				err = fmt.Errorf("%w: %w: %w", err, errdefs.ErrWaitOnClose, errC)
@@ -534,6 +534,21 @@ func runClient(
 		}
 	}
 	return nil
+}
+
+// isCleanSessionEnd reports whether err signals an end-of-session that
+// runClient must swallow rather than surface to cobra as `Error: …`:
+// context cancellation, an operator-initiated detach (errdefs.ErrClientDetached
+// — chained on the close cause by Controller.classifySessionEnd when
+// detachRequested is set), or the remote workload closing the IO
+// connection on a clean exit (errdefs.ErrPeerClosed). Mirrors pkg/attach.Run's
+// classifySessionEnd remap and cmd/sb/attach's nil-return on those
+// public sentinels; sbsh consumes the controller directly so the swallow
+// has to live here. See issue #380.
+func isCleanSessionEnd(err error) bool {
+	return errors.Is(err, context.Canceled) ||
+		errors.Is(err, errdefs.ErrClientDetached) ||
+		errors.Is(err, errdefs.ErrPeerClosed)
 }
 
 // readRootSocketGID resolves --terminal-socket-gid: explicit flag wins,
--- a/cmd/sbsh/sbsh_test.go
+++ b/cmd/sbsh/sbsh_test.go
@@ -19,6 +19,7 @@ package sbsh
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"log/slog"
 	"os"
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #381 Review — `fix: swallow clean session-end sentinels in sbsh runClient`**

LGTM. The fix correctly extends `runClient`'s clean-end set to include `errdefs.ErrClientDetached` and `errdefs.ErrPeerClosed`, which is exactly the chain `Controller.Run` produces (verified at `internal/client/controller.go:312` wrapping the classified cause from `classifySessionEnd`). The two new tests construct the documented `ErrCloseReq → ErrClientDetached/ErrPeerClosed → cause` chain and assert nil; the pre-existing `TestRunTerminal_ErrChildExit`/`ErrChildExitWithWaitCloseError`/`ErrWaitOnReady` keep covering the failure-surfacing arm via `ErrRPCServerExited`, so AC item 4 still holds. `ErrPeerClosed` inclusion is well-justified — it matches the symmetric remap that `pkg/attach.classifySessionEnd` already does. License headers, build target (`sbsh-sb`), commit signature all clean.

Pre-merge: please run the interactive smoke (`./sbsh` → `Ctrl+]` `Ctrl+]`, expect only the yellow `Detached` line and exit 0) on your local TTY before merging — AC item 1 is verifiable only on an interactive host.

PR #384, opened 2026-06-07
fix: restore mouse-tracking and bracketed-paste modes on client teardown

### LINKED ISSUE #382: client: restore output-side mouse tracking and bracketed-paste modes on teardown
## Repro

1. `sbsh terminal` to attach to a terminal session.
2. Inside the remote session, run any program that enables mouse tracking — vim, tmux, htop, fzf, or a modern shell with mouse support — or simply emit `printf '\x1b[?1000h\x1b[?1006h'` (basic + SGR mouse tracking).
3. Detach with `^]^]` (or let the remote process exit, or hang up the peer — any teardown path).
4. In the parent shell, move the mouse or click anywhere in the terminal window.

## Observed

The parent terminal prints raw CSI mouse-report sequences for every mouse movement/click — e.g. `\x1b[<35;42;15M` for motion or `\x1b[<0;42;15M` / `\x1b[<0;42;15m` for press/release under SGR encoding (`?1006`). The parent shell is left in mouse-reporting mode until the user manually runs `reset` or `printf '\x1b[?1000l\x1b[?1002l\x1b[?1003l\x1b[?1006l'`.

Same class of failure if the remote side enabled bracketed paste (`\x1b[?2004h`): the parent shell receives `\x1b[200~`…`\x1b[201~` wrappers around every paste after detach.

## Root cause

`internal/client/clientrunner/terminal.go:74-84` defines the teardown normalization sequence:

```go
const escRestoreParentTerm = "\x1b[?1049l" + // leave the alternate screen buffer
    "\x1b[?25h" +                            // make the cursor visible
    "\x1b[0 q"                               // reset cursor style (DECSCUSR) to terminal default
```

This was added by #365 (commit 84b8c44) to fix cursor/alt-screen leakage. The sequence does not include any of the mouse tracking mode disables (`?1000l`, `?1002l`, `?1003l`, `?1006l`, `?1015l`) or bracketed paste (`?2004l`).

The reader copier (`internal/client/clientrunner/io.go`) writes remote PTY bytes to the parent terminal verbatim (`filter: nil`), so any DEC private mode set by a TUI/shell on the remote side latches onto the parent terminal. The teardown writes only the partial normalization above, leaving every uncovered mode set.

This is the same class of bug as #365 — output-side DEC private modes that the client never normalizes on teardown — for a different group of modes that #365's AC didn't enumerate.

## Suggested fix

Extend `escRestoreParentTerm` in `internal/client/clientrunner/terminal.go:82-84` to also emit the mouse and bracketed-paste mode disables. Resetting a mode that wasn't set is a no-op on every terminal, so the additions are safe to send unconditionally on every teardown path (the existing single-point write at `terminal.go:141-145` already covers detach / process exit / peer hangup / ctx-cancel — no plumbing changes needed):

- `\x1b[?1000l` — disable X11 mouse reporting (basic)
- `\x1b[?1002l` — disable cell-motion tracking
- `\x1b[?1003l` — disable all-motion tracking
- `\x1b[?1006l` — disable SGR mouse encoding
- `\x1b[?1015l` — disable urxvt mouse encoding (defense in depth)
- `\x1b[?2004l` — disable bracketed paste

Update `internal/client/clientrunner/restore_linux_test.go` `assertNormalizeEmitted` substring list to assert the new escapes are emitted on both teardown paths (`TestRestore_EmitsOutputNormalize_OnProcessExit`, `TestRestore_EmitsOutputNormalize_OnDetach`), matching the #365 regression-test shape.

## Acceptance criteria

- [ ] After detaching from a session whose remote side had enabled mouse tracking (any of `?1000`/`?1002`/`?1003`/`?1006`), the parent terminal no longer emits mouse-report CSI sequences on mouse movement/click — verified without a manual `reset`.
- [ ] After the remote process exits and the client exits, the parent terminal is out of mouse-reporting mode.
- [ ] After detach from a session whose remote side enabled bracketed paste (`?2004`), pastes into the parent shell do not arrive wrapped in `\x1b[200~`…`\x1b[201~`.
- [ ] Mouse and bracketed-paste mode resets run on all teardown paths (detach keystroke, remote process exit, peer hangup, ctx-cancel), matching #365's coverage.
- [ ] Restoration is unconditional — does not depend on the remote side emitting a matching reset.

## References

- Closed by #365 for cursor / alt-screen / cursor-style: 84b8c44 (`fix: restore parent terminal output-side DEC mode state on teardown`).
- Same teardown call site: `internal/client/clientrunner/terminal.go:141-145`.
- Regression-test shape to extend: `internal/client/clientrunner/restore_linux_test.go` (`TestRestore_EmitsOutputNormalize_OnProcessExit`, `TestRestore_EmitsOutputNormalize_OnDetach`).

eminwux:
## Summary

- Extends `escRestoreParentTerm` in `internal/client/clientrunner/terminal.go` to also disable mouse tracking (basic/cell-motion/all-motion/SGR/urxvt) and bracketed paste on every teardown path, alongside the alt-screen / cursor-visible / cursor-style resets already emitted per #365.
- Resetting an unset DEC private mode is a no-op on every terminal, so the additions are safe to send unconditionally — no plumbing changes, same single-point write at `terminal.go:142` covers detach / process exit / peer hangup / ctx-cancel.
- Updates the `assertNormalizeEmitted` substring list in `restore_linux_test.go` so both `TestRestore_EmitsOutputNormalize_OnProcessExit` and `TestRestore_EmitsOutputNormalize_OnDetach` regress on the new escapes.

## Test plan

- [x] `make sbsh-sb` — canonical build produces ELF `sbsh` + `sb` hardlink.
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` — full project test target (unit + integration + e2e) passes; the new substrings are asserted in `internal/client/clientrunner`'s `TestRestore_EmitsOutputNormalize_OnProcessExit` and `TestRestore_EmitsOutputNormalize_OnDetach`.
- [ ] Manual on-device verification of the AC's behavioural smoke (parent shell no longer emits mouse-report CSI on click after detach; pastes are not wrapped in `\x1b[200~`…`\x1b[201~`) — deferred to reviewer's environment. The dev-host sandbox runs the e2e suite that asserts the bytes are *emitted by the client*, but it has no interactive terminal to confirm the parent emulator stops reporting after the bytes land.

## Acceptance criteria

- [x] After detach from a remote that enabled mouse tracking, the client emits `?1000l` / `?1002l` / `?1003l` / `?1006l` / `?1015l` on teardown — asserted by `TestRestore_EmitsOutputNormalize_OnDetach`. Behavioural confirmation that the parent emulator exits mouse-reporting mode is the on-device smoke deferred above.
- [x] After remote process exit the same sequence is emitted on the process-exit path — asserted by `TestRestore_EmitsOutputNormalize_OnProcessExit`.
- [x] After detach from a remote that enabled bracketed paste, the client emits `?2004l` on teardown — asserted by both teardown-path tests.
- [x] All teardown paths covered (detach keystroke, remote process exit, peer hangup, ctx-cancel) — the single-point write at `terminal.go:142` is downstream of `restoreOnce.Do`, which #365 already wired into every teardown path; both regression tests exercise the two distinct entry points, the peer-hangup / ctx-cancel paths re-use the same `restoreParentTerminal` body.
- [x] Restoration is unconditional — `escRestoreParentTerm` is a single literal written on every teardown, independent of any remote-side reset.

Closes #382

### DIFF
--- a/internal/client/clientrunner/restore_linux_test.go
+++ b/internal/client/clientrunner/restore_linux_test.go
@@ -173,7 +173,11 @@ func assertNormalizeEmitted(t *testing.T, r *os.File) {
 		t.Fatalf("ReadAll: %v", err)
 	}
 	out := string(buf)
-	for _, want := range []string{"\x1b[?1049l", "\x1b[?25h", "\x1b[0 q"} {
+	for _, want := range []string{
+		"\x1b[?1049l", "\x1b[?25h", "\x1b[0 q",
+		"\x1b[?1000l", "\x1b[?1002l", "\x1b[?1003l",
+		"\x1b[?1006l", "\x1b[?1015l", "\x1b[?2004l",
+	} {
 		if !strings.Contains(out, want) {
 			t.Errorf("missing normalization escape %q in stdout %q", want, out)
 		}
--- a/internal/client/clientrunner/terminal.go
+++ b/internal/client/clientrunner/terminal.go
@@ -73,15 +73,23 @@ const copierWaitTimeout = 500 * time.Millisecond
 
 // escRestoreParentTerm normalizes the parent terminal's output-side DEC private
 // mode state on session teardown: leave the alternate-screen buffer, show the
-// cursor, and reset the cursor style to the terminal default. Symmetric with
-// the modes the server-side attach repaint may set (escEnterAltScreen /
+// cursor, reset the cursor style to the terminal default, disable mouse
+// tracking under every common encoding, and disable bracketed paste. Symmetric
+// with the modes the server-side attach repaint may set (escEnterAltScreen /
 // escHideCursor in internal/terminal/terminalrunner/terminal_screen.go) and
 // with anything a TUI/shell on the remote side wrote through to the parent
 // terminal without a matching reset. Emitted unconditionally per #365 — the
-// client cannot rely on the remote side emitting a matching reset on its own.
+// client cannot rely on the remote side emitting a matching reset on its own;
+// resetting a mode that wasn't set is a no-op on every terminal.
 const escRestoreParentTerm = "\x1b[?1049l" + // leave the alternate screen buffer
 	"\x1b[?25h" + // make the cursor visible
-	"\x1b[0 q" // reset cursor style (DECSCUSR) to terminal default
+	"\x1b[0 q" + // reset cursor style (DECSCUSR) to terminal default
+	"\x1b[?1000l" + // disable X11 mouse reporting (basic)
+	"\x1b[?1002l" + // disable cell-motion mouse tracking
+	"\x1b[?1003l" + // disable all-motion mouse tracking
+	"\x1b[?1006l" + // disable SGR mouse encoding
+	"\x1b[?1015l" + // disable urxvt mouse encoding
+	"\x1b[?2004l" // disable bracketed paste
 
 // restoreParentTerminal performs a single, idempotent "unblock stdin + restore
 // tty" at the session boundary so the parent shell is always handed back a


eminwux:
**PR #384 Review — `fix: restore mouse-tracking and bracketed-paste modes on client teardown`**

LGTM — extends the existing single-point `escRestoreParentTerm` const (`terminal.go:150`, downstream of `restoreOnce.Do`, so all teardown paths are covered) with standard DEC private-mode resets; resetting an unset mode is a documented no-op, so it's safe to send unconditionally with no plumbing change. Escapes are correct, the byte-emission is regressed by `TestRestore_EmitsOutputNormalize_OnDetach`/`OnProcessExit`, no new symbols, no leaks.

One deferred item for you to run pre-merge (host-gated, not blocking): the on-device behavioural smoke — confirm the parent shell no longer emits mouse-report CSI on click after detach and pastes aren't wrapped in `\x1b[200~`…`\x1b[201~`. The sandbox has no interactive tty to verify the emulator stops reporting once the bytes land.

PR #385, opened 2026-06-08
fix: restore parent terminal on Detach RPC failure and drain before banner

### LINKED ISSUE #383: client: Exec.Detach skips parent-terminal restore on RPC failure and writes banner before copier drain
## Context

While reviewing the parent-terminal cleanup paths added by #364 / #365 (and the missing mouse-mode resets tracked in #382), two control-flow gaps surfaced in `Exec.Detach` (`internal/client/clientrunner/lifecycle.go:213-230`). Both are independent of the missing-DEC-modes gap in #382 — they are about *whether* and *when* the existing normalize sequence fires, not its contents.

Distinct from #382 (missing mouse / bracketed-paste mode constants in `escRestoreParentTerm`). The fix surface here is `lifecycle.go`, not `terminal.go`.

## Gap 1 — early-return on RPC failure skips `restoreParentTerminal`

`internal/client/clientrunner/lifecycle.go:213-230`:

```go
func (sr *Exec) Detach() error {
    if err := sr.terminalClient.Detach(sr.ctx, &sr.id); err != nil {
        return err   // ← restoreParentTerminal never called on this path
    }

    sr.logger.Info("Client detached", "client_id", sr.id)
    _, werr := sr.stdout.WriteString("\x1b[93m\r\nDetached\x1b[0m\r\n")
    sr.restoreParentTerminal()
    if werr != nil { ... }
    return nil
}
```

If `terminalClient.Detach` fails (broken control socket, server crashed mid-RPC, RPC timeout), `Exec.Detach` returns early without calling `restoreParentTerminal`. The user has already committed to leaving the session by pressing `^]^]` — they are now stuck in raw mode with broken termios + uncleaned DEC modes until the secondary `EvError → onClosed → Close → restoreParentTerminal` path fires through the IO copier.

In practice the IO socket that backs the RPC is usually dead at the same time the RPC fails, so `EvError` fires nearly simultaneously and the `sync.Once` collapses the two firings to one restore. But there is no contract guaranteeing the IO copier observes the failure on the same tick — a control-socket failure with a still-live IO socket would leave the terminal stranded in raw mode indefinitely (until something else closes the IO conn).

The local cleanup is non-negotiable once `^]^]` has been observed. The RPC failure is a server-side concern, not a reason to keep the parent terminal broken.

## Gap 2 — banner write races inbound copier bytes

`internal/client/clientrunner/lifecycle.go:222-223`:

```go
_, werr := sr.stdout.WriteString("\x1b[93m\r\nDetached\x1b[0m\r\n")
sr.restoreParentTerminal()
```

The "Detached" banner is written *before* `restoreParentTerminal`, which is intentional (raw mode is still active so the embedded `\r\n` does the right thing — see #364 context). However:

- `restoreParentTerminal`'s body cancels ctx, sets the stdin read deadline, then waits up to 500ms for the copier goroutines to drain (`terminal.go:91-145`).
- The socket→stdout copier may be mid-flush with bytes the remote already wrote *before* the detach RPC reply (e.g. a TUI's final repaint frame).
- The banner write at line 222 happens *before* this drain wait, so it can interleave with whatever bytes the copier is mid-flush — producing a "Det" + remote bytes + "ached" or worse, mid-escape splitting.

In normal practice queues are quiescent at detach time and this is invisible, but the ordering is fragile.

### For Gap 1

Restructure `Exec.Detach` so `restoreParentTerminal` runs on every exit path once the function has begun. Either a `defer sr.restoreParentTerminal()` at the top, or a single tail call after both the RPC and the banner write — pick whichever reads cleaner with the existing `\x1b[93mDetached\x1b[0m` banner suppression on the error path (the banner is meaningless if the RPC failed; the restore is not).

### For Gap 2

Move the banner write to *after* the copier drain wait but *before* the `escRestoreParentTerm` write — i.e., after `restoreParentTerminal`'s internal `ctxCancel` + copier-wait, before any termios change. Concretely: split the current monolithic `restoreParentTerminal` so the banner sits at the safe point (post-drain, pre-restore), or inline the banner into the normalize-write section via an optional pre-normalize hook.

Either approach is local to `lifecycle.go` + `terminal.go`. The simpler reading: drop the banner from `Detach` and have `restoreParentTerminal` accept an optional "post-drain message" parameter (empty string on the `Close` path).

## Acceptance criteria

- [ ] `Exec.Detach` calls `restoreParentTerminal` unconditionally once it has been entered — including the `terminalClient.Detach` RPC failure path. Verified by a new test that injects an RPC-returning-error `terminalClient` and asserts the parent terminal is restored.
- [ ] The "Detached" banner is written *after* the inbound socket→stdout copier has been drained (or after the bounded drain timeout), so it cannot interleave with mid-flush remote bytes.
- [ ] Existing `TestRestore_OnDetach` and `TestRestore_EmitsOutputNormalize_OnDetach` regressions continue to pass — the restore reaches the same final state on the happy path.
- [ ] New regression test for the RPC-failure path uses a `mockTerminalClient` whose `Detach` returns an error and verifies the parent terminal is still restored (termios cooked, normalize sequence emitted).

## Out of scope

- Missing DEC mode constants in `escRestoreParentTerm` (mouse modes, bracketed paste) — tracked separately in #382.
- The 500ms `copierWaitTimeout` itself — the bound is documented and correct for the `os.Stdin` non-pollable case (`golang/go#29137`).

## References

- Surfaced during the cleanup-completeness review that also produced #382.
- `internal/client/clientrunner/lifecycle.go:213-230` (`Exec.Detach`).
- `internal/client/clientrunner/terminal.go:91-162` (`restoreParentTerminal`, drain-wait ordering).
- Prior cleanup work: #364 (input-side termios + stdin blocking restore), #365 (output-side DEC mode normalize), #382 (missing mouse / bracketed-paste modes).
- `internal/client/clientrunner/restore_linux_test.go` — shape to extend for the new RPC-failure regression test.

eminwux:
## Summary

Closes two control-flow gaps in `Exec.Detach` (`internal/client/clientrunner/lifecycle.go`) surfaced during the #364/#365/#382 cleanup review. Both are about *whether* and *when* the existing restore sequence fires, not its contents.

- **Gap 1 — restore skipped on RPC failure.** Previously `Detach` returned early when `terminalClient.Detach` failed (broken control socket, server crash mid-RPC, RPC timeout), leaving the parent terminal stranded in raw mode until the secondary `EvError → Close` path *might* fire — with no contract that the IO copier observes the failure on the same tick. The user has already committed to leaving via `^]^]`; the local restore is non-negotiable regardless of the server-side RPC outcome. `Detach` now calls `restoreParentTerminal` on the failure path too, suppressing the (now-misleading) "Detached" banner, and still propagates the RPC error.

- **Gap 2 — banner raced inbound copier bytes.** The "Detached" banner was written *before* `restoreParentTerminal`'s copier-drain wait, so it could interleave with bytes the socket→stdout copier was mid-flush (e.g. a TUI's final repaint frame). The banner is now handed to `restoreParentTerminal` as an optional `postDrainMsg`, written after the inbound copier drains (or the bounded `copierWaitTimeout`) but before the termios raw→cooked restore — so raw mode is still active and the embedded `\r\n` behaves, and it can no longer interleave with mid-flush remote bytes. This is the issue's endorsed "simpler reading."

### Design note (reviewer please confirm)

The banner write is now a best-effort, logged operation inside `restoreParentTerminal` rather than `Detach`'s return value. Under the old code a failed banner `WriteString` (RPC succeeded) turned a successful detach into an error return; under the new contract the restore is non-negotiable and the banner is best-effort, so a banner-write failure on a successful RPC no longer fails `Detach`. `TestDetach_StdoutWriteError` was updated to encode this new contract (it now asserts `Detach` returns `nil` on banner-write failure when the RPC succeeded). This follows directly from the issue's Gap-1 framing ("the banner is meaningless if the RPC failed; the restore is not"); flagging it so the reviewer can push back if the error-propagation behavior should be preserved instead.

## Test plan

- [x] `go build ./...`
- [x] `make sbsh-sb` (canonical build; verified ELF magic `7f 45 4c 46`)
- [x] `go vet ./...`
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full non-e2e suite, all pass
- [x] `make e2e` — all pass; `TestSbsh_Detach` runtime output confirms the new ordering: banner + normalize emitted as one coherent post-drain block `\x1b[93m\r\nDetached\x1b[0m\r\n\x1b[?1049l\x1b[?25h\x1b[0 q`
- [x] `TestRestore_OnDetach` and `TestRestore_EmitsOutputNormalize_OnDetach` regressions still pass (AC item 3)

### Acceptance criteria

- [x] `Exec.Detach` calls `restoreParentTerminal` unconditionally once entered, including the RPC-failure path. New `TestRestore_OnDetach_RPCFailure` injects an RPC-returning-error `mockTerminalClient` and asserts the parent terminal is restored (cooked termios + blocking descriptor) and the normalize sequence is emitted.
- [x] The "Detached" banner is written after the inbound socket→stdout copier drains (or after the bounded drain timeout). New `TestDetach_BannerWrittenBeforeNormalize` asserts the banner is emitted from within `restoreParentTerminal` and precedes the normalize sequence; e2e `TestSbsh_Detach` output confirms it at runtime.
- [x] Existing `TestRestore_OnDetach` and `TestRestore_EmitsOutputNormalize_OnDetach` continue to pass.
- [x] New RPC-failure regression test uses a `mockTerminalClient` whose `Detach` returns an error and verifies the parent terminal is still restored (termios cooked, normalize sequence emitted, banner suppressed).

## Out of scope

- Missing DEC mode constants in `escRestoreParentTerm` (mouse modes, bracketed paste) — tracked in #382 (`terminal.go`, not `lifecycle.go`).
- The 500ms `copierWaitTimeout` bound itself — documented and correct for the `os.Stdin` non-pollable case.

Closes #383

### DIFF
--- a/internal/client/clientrunner/lifecycle.go
+++ b/internal/client/clientrunner/lifecycle.go
@@ -189,7 +189,7 @@ func (sr *Exec) Close(_ error) error {
 		}
 	}
 
-	sr.restoreParentTerminal()
+	sr.restoreParentTerminal("")
 	sr.logger.Debug("Close: cleanup complete")
 
 	sr.metadataMu.Lock()
@@ -211,20 +211,26 @@ func (sr *Exec) Resize(_ api.ResizeArgs) {
 }
 
 func (sr *Exec) Detach() error {
+	// The user has committed to leaving by pressing ^]^]; the local parent-terminal
+	// restore is non-negotiable once we are here. Restore on every exit path —
+	// including a failed Detach RPC (broken control socket, server crash mid-RPC,
+	// RPC timeout) — rather than relying on Close() being driven later via the
+	// conn-close EvError, which is not guaranteed to fire on the same tick. See
+	// issues #364 and #383.
 	if err := sr.terminalClient.Detach(sr.ctx, &sr.id); err != nil {
+		sr.logger.Warn("Detach RPC failed; restoring parent terminal anyway", "client_id", sr.id, "error", err)
+		// Suppress the "Detached" banner — it would be misleading when the RPC
+		// failed — but the restore still runs. See #383.
+		sr.restoreParentTerminal("")
 		return err
 	}
 
 	sr.logger.Info("Client detached", "client_id", sr.id)
-	// Write the confirmation while still in raw mode (explicit \r\n), then
-	// restore the parent terminal unconditionally — the detach path must not
-	// rely on Close() being driven later via the conn-close EvError. See #364.
-	_, werr := sr.stdout.WriteString("\x1b[93m\r\nDetached\x1b[0m\r\n")
-	sr.restoreParentTerminal()
-	if werr != nil {
-		sr.logger.Warn("Failed to write detach message to stdout", "error", werr)
-		return werr
-	}
+	// Hand the confirmation banner to restoreParentTerminal as a post-drain
+	// message so it is written after the inbound copier has drained (no
+	// interleave with mid-flush remote bytes) but while raw mode is still active
+	// so the embedded \r\n behaves. See issues #364 and #383.
+	sr.restoreParentTerminal("\x1b[93m\r\nDetached\x1b[0m\r\n")
 
 	return nil
 }
--- a/internal/client/clientrunner/lifecycle_test.go
+++ b/internal/client/clientrunner/lifecycle_test.go
@@ -196,7 +196,10 @@ func TestDetach_RPCError(t *testing.T) {
 }
 
 // TestDetach_StdoutWriteError covers the branch where the detach succeeds at
-// the RPC layer but writing the confirmation to stdout fails.
+// the RPC layer but writing the confirmation banner to stdout fails. As of #383
+// the banner is a best-effort post-drain message inside restoreParentTerminal,
+// not Detach's return value — the parent-terminal restore is non-negotiable and
+// a failed banner write must not turn a successful detach into an error.
 func TestDetach_StdoutWriteError(t *testing.T) {
 	sr, _ := newLifecycleExec(t)
 	r, w, err := os.Pipe()
@@ -204,12 +207,12 @@ func TestDetach_StdoutWriteError(t *testing.T) {
 		t.Fatalf("pipe: %v", err)
 	}
 	_ = r.Close()
-	_ = w.Close() // closed write end -> WriteString fails
+	_ = w.Close() // closed write end -> banner WriteString fails (best-effort, logged)
 	sr.stdout = w
 	sr.terminalClient = &mockTerminalClient{} // detach succeeds by default
 
-	if errDetach := sr.Detach(); errDetach == nil {
-		t.Fatal("Detach returned nil error when stdout write failed")
+	if errDetach := sr.Detach(); errDetach != nil {
+		t.Fatalf("Detach returned %v on banner-write failure; want nil (RPC succeeded, banner is best-effort)", errDetach)
 	}
 }
 
--- a/internal/client/clientrunner/restore_linux_test.go
+++ b/internal/client/clientrunner/restore_linux_test.go
@@ -18,6 +18,7 @@ package clientrunner
 
 import (
 	"context"
+	"errors"
 	"io"
 	"os"
 	"path/filepath"
@@ -29,6 +30,10 @@ import (
 	"github.com/eminwux/sbsh/pkg/api"
 	"golang.org/x/sys/unix"
 )
+
+// errMockDetach is the sentinel a Detach-RPC-failure mock returns so a test can
+// assert Exec.Detach propagates it while still restoring the parent terminal.
+var errMockDetach = errors.New("detach rpc failed")
 
 // ttyState reports the descriptor's blocking flag and the ECHO/ICANON termios
 // line-discipline flags. It reads them through SyscallConn().Control so the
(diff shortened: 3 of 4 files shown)


eminwux:
**PR #385 Review — `fix: restore parent terminal on Detach RPC failure and drain before banner`**

LGTM — both gaps are correctly closed: the RPC-failure path now calls `restoreParentTerminal("")` (restore runs, banner suppressed, error still propagated), and the success-path banner is handed through as `postDrainMsg` so it lands post-drain / pre-normalize while raw mode is still active. New `TestRestore_OnDetach_RPCFailure` and `TestDetach_BannerWrittenBeforeNormalize` match the code; tests cleanly merge with #384's `escRestoreParentTerm` additions, and the banner still precedes the (now-extended) normalize sequence.

On the design note — **confirmed, keep the change.** Returning `nil` on a banner-write failure when the RPC succeeded is the correct contract: the only real callers (`controller.go:406` log, `controller.go:457` `ErrDetachTerminal` wrap) would otherwise surface a spurious detach failure for a fully-successful detach where only the cosmetic banner couldn't reach a closed stdout. `TestDetach_StdoutWriteError` correctly encodes the new contract.

PR #402, opened 2026-06-09
fix: refuse duplicate --id start that clobbers a live terminal's metadata

### LINKED ISSUE #386: Terminal - duplicate --id start clobbers a live terminal's metadata; prune then deletes the running terminal's run dir
## Summary

Starting a terminal with an `--id` that is already in use by a **live** terminal corrupts the running terminal's on-disk state. The collision is only detected by a *name* check (which a different `--name` sails past) and by a socket probe that runs **after** `metadata.json` has already been overwritten. The result: the live terminal's metadata is clobbered with the dead duplicate's pid/name/state, the live terminal becomes invisible and unaddressable, and `sb prune terminals` then `os.RemoveAll`s the still-running terminal's entire run directory — orphaning the live process with no socket and no metadata.

## Repro

```bash
# 1. Start terminal A (alive)
sbsh terminal --id aaaa1111 --name terma -- /bin/sleep 600 &
#    A is Ready, pid e.g. 22452, serving its socket

# 2. Start B with the SAME --id but a different --name
sbsh terminal --id aaaa1111 --name termb -- /bin/sleep 600
#    -> exits 1: "waiting for readiness has failed: context canceled"

# 3. Inspect state
sb get terminals            # -> "no active terminals found"
sb get terminals -a -o wide # -> aaaa1111 shown as name=termb, STATUS=Exited
sb stop terminal terma      # -> Error: terminal not found
sb prune terminals          # -> "Pruned terminal aaaa1111 (termb)"
```

## Observed (all reproduced on HEAD 263a54b via `make sbsh-sb`)

- After B's failed start, A's `metadata.json` reads `name: termb, pid: <B's dead pid>, state: 0 (Initializing)` while A (pid 22452) is still alive and running `/bin/sleep 600`.
- `sb get terminals` reports no terminals; `-a` shows the entry under the wrong name as `Exited`. The live terminal is invisible and cannot be addressed by its real name.
- `sb prune terminals` deletes A's **entire** run directory (socket, metadata, capture, log) while pid 22452 is alive:
  ```
  === A pid 22452 still alive? YES
  === dir after prune: (empty)
  === ps -p 22452: ./sbsh terminal --id aaaa1111 --name terma -- /bin/sleep 600
  ```
- A is now permanently orphaned: no socket, no metadata — un-attachable and un-stoppable via any `sb` command.

## Root cause

- `internal/terminal/controller.go:127` calls `c.sr.CreateMetadata()` **before** `c.sr.OpenSocketCtrl()` at `:137`.
- `internal/terminal/terminalrunner/metadata.go:54-108` — `CreateMetadata()` unconditionally writes `metadata.json` at the ID-derived path `runPath/terminals/<id>/metadata.json`, overwriting whatever is there.
- `internal/terminal/terminalrunner/sockets.go:198-206` — the only live-peer/ID collision guard (the socket probe) runs *inside* `OpenSocketCtrl`, i.e. after the overwrite.
- `internal/discovery/terminals.go:291-315` — `VerifyTerminalNameAvailable` (the CLI pre-launch guard at `cmd/sbsh/terminal/terminal.go:396`) checks **name only**, not **ID**, so a same-ID/different-name start is not rejected.
- `internal/discovery/terminals.go:113` — `PruneTerminal` `os.RemoveAll`s `Status.TerminalRunPath` for any entry reconciled to `Exited`, which now includes the clobbered (but live) terminal.

## Suggested fix

Detect an existing live owner of `runPath/terminals/<id>/` **before** `CreateMetadata` overwrites the shared ID-derived path (move the socket/live-peer probe ahead of the metadata write), and/or extend `VerifyTerminalNameAvailable` to also reject an in-use **ID**, not just a name. The metadata write must not clobber a path whose current owner is still live.

## Acceptance criteria

- [ ] Starting a terminal with an `--id` already owned by a live terminal fails *without* modifying the live terminal's `metadata.json`.
- [ ] The live terminal remains visible in `sb get terminals` and addressable by its real name after the rejected duplicate start.
- [ ] `sb prune terminals` never removes the run directory of a process that is still alive.
- [ ] A regression test covers the duplicate-ID/different-name start path end to end (metadata intact, prune safe).

## References

- `internal/terminal/controller.go:127,137`
- `internal/terminal/terminalrunner/metadata.go:54-108`
- `internal/terminal/terminalrunner/sockets.go:198-206`
- `internal/discovery/terminals.go:113,291-315`
- `cmd/sbsh/terminal/terminal.go:396`

eminwux:
## Summary

Starting a terminal with an `--id` already owned by a **live** terminal corrupted the running terminal's on-disk state: the only ID-collision guard (the socket live-peer probe in `OpenSocketCtrl`) ran *after* `CreateMetadata` had already overwritten `metadata.json` at the shared ID-derived path, and the CLI pre-launch guard checked the **name** only (so a different `--name` sailed past). The live terminal's metadata got clobbered with the dead duplicate's pid/name/state, making it invisible/unaddressable, and `sb prune terminals` then `os.RemoveAll`'d the still-running terminal's run dir.

This adds detection of a live owner *before* the metadata write, at two layers:

- **Runner backstop (load-bearing):** `CreateMetadata` now refuses to overwrite `metadata.json` when its current owner is an active terminal — `State != Exited` **and** `IsInstanceAlive(pid, pidStart)`, the same liveness notion reconcile/prune use. This covers embedders that drive the controller directly, not just the CLI. Returns `errdefs.ErrTerminalIDInUse` without touching the live owner's metadata.
- **CLI fail-fast:** new `discovery.VerifyTerminalIDAvailable`, mirroring `VerifyTerminalNameAvailable`, wired alongside the existing name check in `cmd/sbsh/terminal`. Gives a clear pre-launch error instead of the confusing `context canceled` from the repro.

To keep "is this terminal alive" defined in one place, the existing unexported `isInstanceAlive` is exposed as `discovery.IsInstanceAlive` and reused by the runner (no terminalrunner→discovery import cycle — discovery imports neither package).

### Notes for the reviewer (non-obvious calls)

- **Gate vs. flip / blast radius:** the issue's suggested fix was "move the socket probe ahead of the write, **and/or** extend the name guard to reject in-use ID." I implemented **both** as defense-in-depth — the runner backstop is mandatory (it's the only layer covering embedders), and the CLI check is a small fail-fast UX win mirroring the existing name guard exactly. Both are purely additive: no existing valid flow changes.
- **Owner liveness uses `State != Exited && alive`, not pid-liveness alone.** A pid-only check would wrongly block an embedder that reuses one process+id after a clean close (prior `State==Exited` but the hosting process is still alive). Keying on the reconcile/prune notion of "active" makes ID-recycling semantics identical to name-recycling: an exited-or-dead prior owner frees the ID.

## Test plan
- [x] `make sbsh-sb` — canonical build; `./sbsh` is ELF (`7f 45 4c 46`)
- [x] `go build ./...`
- [x] `go vet ./internal/... ./cmd/sbsh/...`
- [x] `make test` (full CI suite incl. e2e) — exit 0
- [x] New `TestCreateMetadata_RefusesLiveOwnerDuplicateID` — end-to-end repro: duplicate-id/different-name start refused, live owner's metadata intact, and a reconcile pass confirms the entry is not flipped to Exited (so prune leaves the running dir alone)
- [x] New `TestCreateMetadata_RecyclesExitedOwnerID` — exited-owner ID is free to recycle (guard doesn't over-block)
- [x] New `TestVerifyTerminalIDAvailable` — CLI guard: active id blocks reuse with a different name, exited/stale ids recycle

## Acceptance criteria
- [x] Starting a terminal with an `--id` already owned by a live terminal fails *without* modifying the live terminal's `metadata.json`
- [x] The live terminal remains visible and addressable by its real name after the rejected duplicate start (metadata untouched)
- [x] `sb prune terminals` never removes the run directory of a still-alive process (intact `State != Exited` + live pid is not reconciled to Exited; prune acts only on Exited)
- [x] A regression test covers the duplicate-ID/different-name start path end to end (metadata intact, prune-safe)

Closes #386

### DIFF
--- a/cmd/sbsh/terminal/terminal.go
+++ b/cmd/sbsh/terminal/terminal.go
@@ -402,6 +402,15 @@ that have exited can be reused.
 				return errAvail
 			}
 
+			if errIDAvail := discovery.VerifyTerminalIDAvailable(
+				cmd.Context(),
+				logger,
+				terminalSpec.RunPath,
+				string(terminalSpec.ID),
+			); errIDAvail != nil {
+				return errIDAvail
+			}
+
 			logger.Debug("Built terminal spec", "terminalSpec", fmt.Sprintf("%+v", terminalSpec))
 
 			if logger.Enabled(context.Background(), slog.LevelDebug) {
--- a/internal/discovery/reconcile.go
+++ b/internal/discovery/reconcile.go
@@ -57,6 +57,15 @@ func isProcessAlive(pid int) bool {
 	return false
 }
 
+// IsInstanceAlive reports whether the process instance identified by
+// (pid, pidStart) is still running, using the same liveness+PID-reuse rules
+// reconcile and prune rely on. Exported so the terminal runner can refuse to
+// clobber a live owner's metadata before it overwrites the ID-derived path,
+// keeping the runner's notion of "live" identical to reconcile's. See #386.
+func IsInstanceAlive(pid int, pidStart uint64) bool {
+	return isInstanceAlive(pid, pidStart)
+}
+
 // isInstanceAlive extends isProcessAlive with PID-reuse rejection: if a
 // start-time token was recorded, the live PID must still belong to the same
 // process instance that produced it. A zero token degrades to liveness-only,
--- a/internal/discovery/terminals.go
+++ b/internal/discovery/terminals.go
@@ -314,6 +314,40 @@ func VerifyTerminalNameAvailable(
 	return nil
 }
 
+// VerifyTerminalIDAvailable returns errdefs.ErrTerminalIDInUse if an active
+// (non-Exited) terminal under runPath already carries id. It mirrors
+// VerifyTerminalNameAvailable: a different --name sails past the name check, so
+// the ID needs its own pre-launch guard to fail fast before the runner would
+// otherwise clobber the live owner's metadata (the runner-level backstop lives
+// in CreateMetadata). Stale metadata whose owner is gone is reconciled to
+// Exited and does not block reuse, so an exited terminal's ID can be recycled.
+// An empty id or runPath is treated as unconditionally available. See #386.
+func VerifyTerminalIDAvailable(
+	ctx context.Context,
+	logger *slog.Logger,
+	runPath string,
+	id string,
+) error {
+	if id == "" || runPath == "" {
+		return nil
+	}
+	terminals, err := ScanTerminals(ctx, logger, runPath)
+	if err != nil {
+		return err
+	}
+	ReconcileTerminals(ctx, logger, runPath, terminals)
+	for _, t := range terminals {
+		if t.Status.State == api.Exited {
+			continue
+		}
+		if pkgdiscovery.TerminalID(t) != id {
+			continue
+		}
+		return fmt.Errorf("%w: %q (name %s)", errdefs.ErrTerminalIDInUse, id, pkgdiscovery.TerminalName(t))
+	}
+	return nil
+}
+
 func PrintTerminalSpec(s *api.TerminalSpec, logger *slog.Logger) error {
 	if s == nil {
 		logger.Info("nil terminal spec")
--- a/internal/discovery/terminals_test.go
+++ b/internal/discovery/terminals_test.go
@@ -146,3 +146,75 @@ func TestVerifyTerminalNameAvailable(t *testing.T) {
 		}
 	})
 }
+
+// TestVerifyTerminalIDAvailable mirrors the name-availability checks for the
+// ID pre-launch guard added in #386: a same-ID start (which a different --name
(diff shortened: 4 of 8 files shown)


eminwux:
**PR #402 Review — `fix: refuse duplicate --id start that clobbers a live terminal's metadata`**

LGTM — adds a live-owner guard before the ID-derived `metadata.json` write at both the runner backstop (`CreateMetadata`, covering embedders via `pkg/terminal/server`) and a CLI fail-fast (`VerifyTerminalIDAvailable`, mirroring the name guard exactly); keyed on the same `State != Exited && IsInstanceAlive` notion reconcile/prune use, no import cycle, errors `%w`-wrapped so `errors.Is` holds across both layers, and the end-to-end runner test asserts the live owner's metadata stays intact and prune-safe.

PR #403, opened 2026-06-10
fix: bound Detach and readiness-wait RPCs so a wedged daemon can't trap the client in raw mode

### LINKED ISSUE #389: Client - Detach RPC has no deadline: a wedged terminal daemon leaves the client permanently stuck in raw mode
## Summary

The Detach RPC the client issues on `^]^]` carries no timeout, unlike every sibling RPC on the same path. If the terminal daemon stops responding while still holding its listening socket (so the connect + request-write succeed but no reply ever comes), the detach call blocks forever **inside the controller event loop**. Because `restoreParentTerminal` only runs *after* the Detach RPC returns, the user's terminal is left permanently in raw mode (ISIG off, so even `^C` cannot generate SIGINT) — unrecoverable from the keyboard.

## Repro

```bash
sb attach <terminal>          # attach normally
kill -STOP <terminal-daemon-pid>   # daemon wedged but socket still in kernel backlog
# in the attached client, press ^] ^]
```

The client freezes in raw mode. Only an external `kill -TERM <client-pid>` unwedges it.

## Root cause

- `internal/client/controller.go:404-409` — `EvDetach` calls `s.Detach()` **synchronously inside the controller event loop** (`handleEvent`), so a blocked detach also prevents any subsequent `EvError` from being processed.
- `internal/client/clientrunner/lifecycle.go:220` — `sr.terminalClient.Detach(sr.ctx, …)`.
- `pkg/rpcclient/terminal/rpc_session.go:104-141` — `callWithCodec` dials, writes the request, then blocks in `rpcc.Call` with **no deadline**; the only escapes are a reply or `sr.ctx` cancellation.
- `restoreParentTerminal` runs only after Detach returns (`lifecycle.go:224/233`); the RPC never returns and `sr.ctx` is cancelled only by `Close()`/restore itself, so restore never runs.

Contrast: every other RPC on this path has a deadline — Ping 3s (`io.go:49`), initial Resize 100ms (`io.go:178-186`), per-WINCH Resize 100ms (`io.go:208`). #385 fixed the Detach-RPC-*fails* case; the Detach-RPC-*never-returns* case is still unhandled.

Related: `waitReady`'s 2s timeout (`io.go:226-249`) does not bound the `State` RPC itself (`getTerminalState` uses `sr.ctx`, `metadata.go:118`), and the second `waitReady(api.Ready)` at `lifecycle.go:151` runs *after* raw mode is set — so a server that wedges mid-attach also hangs the client in raw mode with no 2s escape.

## Suggested fix

Give the Detach RPC a bounded deadline (e.g. a `context.WithTimeout` derived from `sr.ctx`, matching the Ping/Resize pattern). On timeout, fall through to `restoreParentTerminal` + `Close` exactly as the Detach-*failure* path from #385 does, so the terminal is always restored even when the daemon never replies. Consider also bounding the post-raw-mode `State` RPC so a mid-attach wedge can't trap the client in raw mode.

## Acceptance criteria

- [ ] The Detach RPC uses a bounded context; a non-responding daemon does not block the controller event loop indefinitely.
- [ ] On Detach RPC timeout, the parent terminal is restored (raw mode, mouse-tracking, bracketed-paste cleared) before the client exits.
- [ ] The post-raw-mode readiness wait cannot trap the client in raw mode against a wedged server.

## References

- `internal/client/controller.go:404-409`
- `internal/client/clientrunner/lifecycle.go:220,224,233`
- `pkg/rpcclient/terminal/rpc_session.go:104-141`
- `internal/client/clientrunner/io.go:49,178-186,208,226-249`
- prior art: #385 (Detach RPC failure restore), #384 (mode restore)

eminwux:
## Summary

A wedged terminal daemon (one that accepts the connection and request-write but never replies — e.g. `kill -STOP` on the daemon while the socket stays in the kernel backlog) could leave the client permanently stuck in raw mode, unrecoverable from the keyboard. Two RPC paths on the client lacked a deadline:

- **Detach RPC** (`^]^]`): issued synchronously inside the controller event loop with `sr.ctx` (only cancelled by `Close()`/restore itself), so a never-returning Detach blocked forever and `restoreParentTerminal` — which runs only *after* Detach returns — never ran.
- **Post-raw-mode readiness wait**: `waitReady`'s 2s loop budget did not bound the `State` RPC it calls each tick (`getTerminalState` used `sr.ctx`), so a `State` RPC that wedged blocked the loop *inside* `getTerminalState`, and the 2s `ctx.Done()` escape never got a chance to fire.

### Changes

- `lifecycle.go` — `Detach()` now derives a `context.WithTimeout(sr.ctx, detachTimeout)` (3s, matching the Ping deadline) for the Detach RPC. On timeout the RPC returns via `callWithCodec`'s ctx-cancel branch and falls through to the existing #385 restore-on-failure path (`restoreParentTerminal("")`), so the parent terminal is always restored.
- `metadata.go` — `getTerminalState` now takes a `ctx context.Context` instead of using `sr.ctx`, so the `State` RPC honors the caller's deadline.
- `io.go` — `waitReady` passes its bounded 2s `ctx` into `getTerminalState`, so a wedged `State` RPC is cancelled at the readiness deadline and `waitReady` returns an error. The caller (`controller.go:251`) then runs `Close` → `restoreParentTerminal`, releasing raw mode.

### Notes for the reviewer

- The Detach timeout reuses the existing #385 failure path rather than adding a new branch — a timeout is just another RPC error, and that path already clears raw mode, mouse-tracking, and bracketed-paste via `restoreParentTerminal` (no banner on the error path, per #383). Verified `controller.go:249-257` calls `s.Close(errAttach)` → `sr.Close` → `restoreParentTerminal` on a post-raw-mode `waitReady` error, so AC#3's restore is covered without new wiring.
- `getTerminalState`'s only non-test caller is `waitReady`; signature change is contained (two test call sites updated).

## Test plan

- [x] `make sbsh-sb` — canonical build; binary is ELF (`7f 45 4c 46`)
- [x] `go build ./...`
- [x] `go vet ./internal/client/...`
- [x] `make test` — full CI suite incl. e2e, exit 0
- [x] New `TestDetach_RPCBounded` — asserts the Detach RPC ctx carries a deadline and a blocking RPC returns within the bound
- [x] New `Test_WaitReady_WedgedStateRPC` — asserts a wedged `State` RPC no longer hangs `waitReady`

## Acceptance criteria

- [x] The Detach RPC uses a bounded context; a non-responding daemon does not block the controller event loop indefinitely.
- [x] On Detach RPC timeout, the parent terminal is restored (raw mode, mouse-tracking, bracketed-paste cleared) before the client exits.
- [x] The post-raw-mode readiness wait cannot trap the client in raw mode against a wedged server.

Closes #389

### DIFF
--- a/internal/client/clientrunner/io.go
+++ b/internal/client/clientrunner/io.go
@@ -248,8 +248,9 @@ func (sr *Exec) waitReady(states ...api.TerminalStatusMode) error {
 			sr.logger.ErrorContext(sr.ctx, "waitReady: context done before ready", "error", ctx.Err())
 			return fmt.Errorf("context done before ready: %w", ctx.Err())
 		case <-ticker.C:
-			// refresh curState
-			curState, err := sr.getTerminalState()
+			// refresh curState — pass the bounded ctx so a wedged State RPC
+			// cannot outlive the readiness deadline (issue #389).
+			curState, err := sr.getTerminalState(ctx)
 			if err != nil {
 				sr.logger.ErrorContext(sr.ctx, "waitReady: getTerminalState failed during wait", "error", err)
 				return fmt.Errorf("get terminal state failed during wait: %w", err)
--- a/internal/client/clientrunner/io_test.go
+++ b/internal/client/clientrunner/io_test.go
@@ -329,6 +329,60 @@ func Test_WaitReady_Timeout(t *testing.T) {
 	}
 }
 
+// Test_WaitReady_WedgedStateRPC covers issue #389: a State RPC that never
+// replies (daemon wedged with the socket still in the kernel backlog) must not
+// trap the readiness loop. Because waitReady passes its bounded context to the
+// State RPC, a blocking RPC is cancelled at the readiness deadline and waitReady
+// returns an error instead of hanging the caller in raw mode forever.
+func Test_WaitReady_WedgedStateRPC(t *testing.T) {
+	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelError}))
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	sr := &Exec{
+		ctx:    ctx,
+		logger: logger,
+		terminal: &api.AttachedTerminal{
+			Spec: &api.TerminalSpec{
+				ID:   api.ID("test-terminal"),
+				Name: "test",
+			},
+		},
+		metadata: api.ClientDoc{
+			APIVersion: api.APIVersionV1Beta1,
+			Kind:       api.KindClient,
+			Metadata: api.ClientMetadata{
+				Name:        "test-client",
+				Labels:      make(map[string]string),
+				Annotations: make(map[string]string),
+			},
+			Spec: api.ClientSpec{},
+		},
+	}
+
+	// Wedged daemon: the State RPC blocks until its context is cancelled.
+	sr.terminalClient = &mockTerminalClient{
+		stateFunc: func(rpcCtx context.Context, _ *api.TerminalStatusMode) error {
+			<-rpcCtx.Done()
+			return rpcCtx.Err()
+		},
+	}
+
+	done := make(chan error, 1)
+	go func() {
+		done <- sr.waitReady(api.Ready)
+	}()
+
+	select {
+	case err := <-done:
+		if err == nil {
+			t.Fatal("waitReady returned nil against a wedged State RPC; want an error so the caller can restore the terminal")
+		}
+	case <-time.After(waitReadyTimeoutSeconds*time.Second + 3*time.Second):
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #403 Review — `fix: bound Detach and readiness-wait RPCs so a wedged daemon can't trap the client in raw mode`**

LGTM — bounds the Detach RPC (3s, matching the Ping deadline) and threads `waitReady`'s 2s ctx into the `State` RPC, so a wedged daemon can no longer block the controller loop or trap the client in raw mode; both timeouts fall through to the existing #385 restore path (`controller.go:251` → `Close` → `restoreParentTerminal`), the signature change is contained to `waitReady` + two test sites, and the two new tests exercise each wedged-RPC path.

PR #404, opened 2026-06-10
fix: sanitize control/escape bytes in sb get human-readable listings

### LINKED ISSUE #390: CLI - terminal/profile/client names are printed unescaped in 'sb get', allowing ANSI/OSC escape injection into the operator's terminal
## Summary

`sb get terminals` / `get profiles` / `get clients` write the name and command fields to the operator's terminal via `fmt.Fprintf("%s", …)` with no escaping. Names originate from attacker-controllable inputs (`--name`, profile `metadata.name`, or any writer of `metadata.json` under the run path), so a name containing raw ANSI/OSC bytes is rendered verbatim when an operator lists terminals — classic terminal-output injection (spoof rows, hide entries, move the cursor, set the title bar, clear the screen).

## Repro

```bash
sbsh terminal --name $'evil\033[31mRED\033[0m' -- /bin/bash -i &
sb get terminals | od -c | grep -A1 'e   v   i   l'
```

## Observed

```
0000060   e   v   i   l 033   [   3   1   m   R   E   D 033   [   0   m
```

The raw `033 [ 3 1 m … 033 [ 0 m` bytes are emitted to stdout unmodified — in both the compact and `-o wide` views. An OSC title-set (`\033]0;...\007`) or `\033[2J` clear in a name would execute in the terminal of whoever runs `sb get`.

## Root cause

- `internal/discovery/terminals.go:166-184` (`printTerminals`) — name/cmd written via `fmt.Fprintf(tw, "%s\t…")` with no sanitization.
- Same pattern in `internal/discovery/profiles.go:125-132` (profile name + cmd) and `internal/discovery/clients.go:165-172` (client names).

## Suggested fix

Sanitize control/escape bytes out of all externally-sourced fields before printing them to a terminal (strip or `\xNN`-escape C0/C1 control characters and ESC/CSI/OSC sequences). A small shared helper applied at the print sites in `internal/discovery/{terminals,profiles,clients}.go` covers all three. Machine-readable `-o json`/`-o yaml` output should keep the raw value (those are not terminal-rendered).

## Acceptance criteria

- [ ] Listing a terminal/profile/client whose name contains ESC/CSI/OSC bytes does not emit those bytes verbatim to the terminal in the compact or wide views.
- [ ] `-o json` / `-o yaml` still carry the exact original name.
- [ ] A test asserts control bytes are neutralized in the human-readable listing.

## References

- `internal/discovery/terminals.go:166-184`
- `internal/discovery/profiles.go:125-132`
- `internal/discovery/clients.go:165-172`

eminwux:
## Summary

- `sb get terminals` / `get profiles` / `get clients` wrote name and command fields straight to the operator's terminal via `fmt.Fprintf("%s", …)`. Names originate from attacker-controllable inputs (`--name`, profile `metadata.name`, any writer of `metadata.json` under the run path), so a name carrying raw ANSI/OSC bytes was rendered verbatim — classic terminal-output injection (spoof rows, hide entries, move the cursor, set the title bar, clear the screen) against whoever runs `sb get`.
- New shared helper `sanitizeField` (`internal/discovery/sanitize.go`) replaces every C0 (0x00–0x1F), DEL (0x7F), and C1 (0x80–0x9F) control rune with a printable `\xNN` escape. Escaping the ESC introducer alone defangs every CSI/OSC sequence; tab/newline are escaped too since they are the tabwriter column/row delimiters and a raw one would corrupt the table.
- Applied at the human-table print sites in `internal/discovery/{terminals,profiles,clients}.go`, to **all** externally-sourced free-form fields rendered there (name, profile, command, pty, attachers, labels, id, run-target) — not only the two the AC names. The root cause is shared by every metadata-sourced string in the row, so a name-only fix would leave the same injection reachable through, e.g., a profile name or a label value. Enum-derived (`STATUS`) and numeric (`CREATED`/`ATTACHED`/`vars`) columns are left untouched.
- `-o json` / `-o yaml` paths are deliberately **not** sanitized — they are not terminal-interpreted and consumers rely on the exact original bytes.

### Scope note for the reviewer

The AC enumerates name + cmd; the diff sanitizes the full set of externally-sourced fields in the same rows (lower-blast-radius would be name-only, but that leaves the identical injection reachable via sibling fields, so the security-correct reading is to neutralize every metadata-sourced field at the print boundary). Push back if you'd prefer the narrower name+cmd-only scope.

## Test plan

- [x] `make test` — full CI suite green (cmd/sb, cmd/sbsh, internal/terminal, internal/client, `-tags=integration` cmd/sb/get, e2e)
- [x] `go build ./...`
- [x] `go vet ./internal/discovery/...`
- [x] `make sbsh-sb` — produces an ELF executable (verified `7f 45 4c 46` magic; `sbsh version` runs)
- [x] New `internal/discovery/sanitize_test.go` — unit-tests the helper (ESC/OSC/CSI/NUL/DEL/tab/newline/C1 all escaped; clean ASCII + Unicode untouched) and asserts no raw control rune survives.
- [x] New `internal/discovery/injection_test.go` — listing-level tests that an evil name (`\x1b]0;…\x07\x1b[31m\x1b[2J`) emits no raw control byte in the compact or wide views for terminals/profiles/clients, and that `-o json` / `-o yaml` preserve the exact original name through a decode round-trip.

## Acceptance criteria

- [x] Listing a terminal/profile/client whose name contains ESC/CSI/OSC bytes does not emit those bytes verbatim to the terminal in the compact or wide views.
- [x] `-o json` / `-o yaml` still carry the exact original name.
- [x] A test asserts control bytes are neutralized in the human-readable listing.

Closes #390

### DIFF
--- a/internal/discovery/clients.go
+++ b/internal/discovery/clients.go
@@ -163,14 +163,14 @@ func printClients(w io.Writer, clients []api.ClientDoc, printAll, wide bool) err
 		}
 		if wide {
 			fmt.Fprintf(tw, "%s\t%s\t%s\t%s\n",
-				pkgdiscovery.ClientID(s),
-				pkgdiscovery.ClientName(s),
+				sanitizeField(pkgdiscovery.ClientID(s)),
+				sanitizeField(pkgdiscovery.ClientName(s)),
 				s.Status.State.String(),
-				joinLabels(clientLabels(s)),
+				sanitizeField(joinLabels(clientLabels(s))),
 			)
 		} else {
 			fmt.Fprintf(tw, "%s\t%s\n",
-				pkgdiscovery.ClientName(s),
+				sanitizeField(pkgdiscovery.ClientName(s)),
 				s.Status.State.String(),
 			)
 		}
--- a/internal/discovery/injection_test.go
+++ b/internal/discovery/injection_test.go
@@ -1,0 +1,147 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package discovery_test
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/eminwux/sbsh/internal/discovery"
+	"github.com/eminwux/sbsh/pkg/api"
+	"go.yaml.in/yaml/v3"
+)
+
+// evilName carries an OSC title-set, a colour CSI, and a clear-screen CSI — the
+// classic terminal-output-injection payloads an attacker can smuggle through a
+// terminal/profile/client name (#390).
+const evilName = "evil\x1b]0;pwned\x07\x1b[31m\x1b[2J"
+
+// assertNoRawEscapes fails if out carries any byte a terminal would interpret as
+// a control/escape command, and confirms the row was actually rendered (the
+// visible "evil" prefix survives, escaped ESC bytes are present) rather than
+// silently dropped.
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #404 Review — `fix: sanitize control/escape bytes in sb get human-readable listings`**

LGTM — `sanitizeField` escapes every C0/DEL/C1 control rune at the human-table print boundary while leaving `-o json`/`-o yaml` byte-exact (round-trip test proves it); the broader "sanitize all externally-sourced fields, not just name+cmd" scope is the security-correct call, and lone non-UTF-8 C1 bytes are neutralized to U+FFFD by the range loop before reaching output. Unit + listing-level tests cover terminals/profiles/clients in both views, CI is green, and the single commit is verified.

PR #405, opened 2026-06-10
fix: route terminal metadata accesses in SetupShell/PostAttachShell through metadataMu

### LINKED ISSUE #388: Terminal - data races on sr.metadata: SetupShell writes Spec.Prompt and PostAttachShell spin-reads Status.State without metadataMu
## Summary

Several accesses to `sr.metadata` in the terminal runner happen with **no lock held**, while other code paths read/write the same struct under `metadataMu`. These are genuine data races (undefined behavior under the Go memory model). The most easily triggered one occurs in the **normal CLI flow**: an attaching `sb` client polls the `State` RPC every ~50ms (which copies the whole `sr.metadata` struct under `metadataMu.RLock`) while the controller goroutine runs `SetupShell`, which writes `sr.metadata.Spec.Prompt` with no lock at all. An `RLock`'d reader does not exclude a lock-free writer.

## Affected accesses

Unlocked (racy):
- `internal/terminal/terminalrunner/lifecycle.go:471-475` — reads `sr.metadata.Spec.SetPrompt` / `sr.metadata.Spec.Prompt` and **writes** `sr.metadata.Spec.Prompt`, no lock:
  ```go
  if sr.metadata.Spec.SetPrompt {
      if sr.metadata.Spec.Prompt == "" {
          sr.metadata.Spec.Prompt = "(sbsh-" + string(sr.metadata.Spec.ID) + ")"  // unlocked write
      }
      ...
  ```
- `internal/terminal/terminalrunner/lifecycle.go:518` — `for sr.metadata.Status.State != api.Ready {` spin-reads `Status.State` unlocked (also no acquire semantics, so observing `Ready` isn't guaranteed).

Racing locked accesses (same fields, under `metadataMu`):
- `internal/terminal/terminalrunner/metadata.go:181-183` — `Status.State` written under `metadataMu.Lock`.
- `internal/terminal/terminalrunner/metadata.go:210-216` — `Metadata()` copies the whole struct (`metadataCopy := sr.metadata`) under `metadataMu.RLock`.

## Trigger paths

- **Pair A (concurrent attach):** two clients attach concurrently → two `Controller.Attach` → `PostAttachShell` calls run on separate `ServeCodec` goroutines; one runs `updateTerminalState` writing `Status.State` under the lock while the other spin-reads it unlocked at `lifecycle.go:518`. Same shape for `Close`'s `updateTerminalState(api.Exited)` vs an in-flight attach.
- **Pair B (normal single-client flow):** `sb` clients poll the `State` RPC every 50ms while waiting for readiness (`internal/client/clientrunner/io.go:242-257` → `Controller.State` → `sr.Metadata()` → `metadataCopy := sr.metadata` under `RLock`), while the controller goroutine writes `sr.metadata.Spec.Prompt` lock-free in `SetupShell` (`lifecycle.go:473`).

## Impact

Genuine data race / undefined behavior. A `-race` build of the daemon fails as soon as a client polls `State` during shell setup or two attaches overlap. The existing `make test-race` suite passes only because no current test polls `Metadata()` concurrently with `SetupShell` (or overlaps two attaches through `PostAttachShell`).

## Suggested fix

Route all `sr.metadata` reads/writes through `metadataMu` (or the existing accessor helpers). In `SetupShell`, take `metadataMu.Lock` around the `Spec.Prompt` read-modify-write; in `PostAttachShell`, read `Status.State` via a locked accessor (e.g. reuse `Metadata()` or add a small `state()` getter) rather than touching the field directly.

## Acceptance criteria

- [ ] No unlocked reads or writes of `sr.metadata` remain in `lifecycle.go` (`SetupShell`, `PostAttachShell`).
- [ ] A `-race` test that polls the `State` RPC concurrently with terminal setup, and one that overlaps two attaches, both pass clean.
- [ ] The new test would fail on the current code (demonstrates the race is real).

## References

- `internal/terminal/terminalrunner/lifecycle.go:471-475,518`
- `internal/terminal/terminalrunner/metadata.go:181-183,210-216`
- `internal/client/clientrunner/io.go:242-257`

eminwux:
## Summary

- `SetupShell` read-modify-wrote `sr.metadata.Spec.Prompt` with **no lock held** (`lifecycle.go:471-475`), while an attaching client polls the `State` RPC every ~50ms — which copies the whole `sr.metadata` struct under `metadataMu.RLock` via `Metadata()`. An `RLock`'d reader does not exclude a lock-free writer, so this is a genuine data race in the normal single-client flow (issue Pair B).
- `PostAttachShell` spin-read `sr.metadata.Status.State` unlocked (`lifecycle.go:518`) while `updateTerminalState` writes the same field under `metadataMu.Lock` — racy under two overlapping attaches, and with no acquire semantics observing `Ready` wasn't guaranteed (issue Pair A).
- Fix: take `metadataMu.Lock` around the `Spec.Prompt` read-modify-write in `SetupShell` (snapshot the resolved prompt, write to the PTY outside the critical section); add a small `currentState()` read-locked accessor and use it for `PostAttachShell`'s readiness spin. Also snapshot `Spec.Stages.PostAttach` under `RLock` before `writeStage`, since `SetupShell`'s `Spec.Prompt` write means `Spec` is not immutable — so AC #1's "no unlocked reads or writes" holds for the whole of both functions.

Note: the accessor is named `currentState()` rather than the issue's suggested `state()` because `Exec` already has a `state api.TerminalState` field (a different type from `Status.State`'s `api.TerminalStatusMode`); a method named `state` collides with the field.

## Test plan

- [x] `go vet ./...` — clean
- [x] New `-race` tests in `metadata_lock_race_test.go`:
  - `TestSetupShell_StatePollRace` — polls `Metadata()` concurrently with `SetupShell` (Pair B)
  - `TestPostAttachShell_OverlappingAttachesRace` — two goroutines drive interleaved `PostAttachShell` readiness spins and `updateTerminalState` writes (Pair A)
- [x] **Demonstrated the tests fail on pre-fix code** — stashed the production fix and re-ran under `-race`: both fail with the exact races the issue named (`lifecycle.go:473` write vs `metadata.go:271` `Metadata()` read; `lifecycle.go:518` read vs `metadata.go:239` `updateTerminalState` write).
- [x] `make test` — pass (incl. e2e)
- [x] `make test-race` — pass (all packages clean)

## Acceptance criteria

- [x] No unlocked reads or writes of `sr.metadata` remain in `lifecycle.go` (`SetupShell`, `PostAttachShell`).
- [x] A `-race` test that polls the `State` RPC concurrently with terminal setup, and one that overlaps two attaches, both pass clean.
- [x] The new tests fail on the current code (demonstrated via stash + `-race`).

Closes #388

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -468,11 +468,21 @@ func (sr *Exec) SetupShell() error {
 	// 	sr.logger.Error("failed to send CTRL-L", "id", sr.id, "err", err)
 	// 	return fmt.Errorf("failed to send CTRL-L: %w", err)
 	// }
-	if sr.metadata.Spec.SetPrompt {
-		if sr.metadata.Spec.Prompt == "" {
-			sr.metadata.Spec.Prompt = "(sbsh-" + string(sr.metadata.Spec.ID) + ")"
-		}
-		promptCmd := `export PS1=` + sr.metadata.Spec.Prompt + "\n"
+	// Take metadataMu around the Spec.Prompt read-modify-write: an attaching
+	// client polls the State RPC (which copies sr.metadata under metadataMu)
+	// every ~50ms during setup, and an RLock'd reader does not exclude a
+	// lock-free writer. Snapshot the resolved prompt under the lock and write
+	// it to the PTY outside the critical section. See #388.
+	sr.metadataMu.Lock()
+	setPrompt := sr.metadata.Spec.SetPrompt
+	if setPrompt && sr.metadata.Spec.Prompt == "" {
+		sr.metadata.Spec.Prompt = "(sbsh-" + string(sr.metadata.Spec.ID) + ")"
+	}
+	prompt := sr.metadata.Spec.Prompt
+	sr.metadataMu.Unlock()
+
+	if setPrompt {
+		promptCmd := `export PS1=` + prompt + "\n"
 
 		sr.logger.Debug("setupShell: setting prompt", "cmd", promptCmd)
 		if _, err := sr.Write([]byte(promptCmd)); err != nil {
@@ -515,7 +525,7 @@ func (sr *Exec) writeStage(stage *[]api.ExecStep) error {
 func (sr *Exec) PostAttachShell() error {
 	sr.logger.Info("PostAttachShell: waiting for terminal to be ready", "id", sr.id)
 
-	for sr.metadata.Status.State != api.Ready {
+	for sr.currentState() != api.Ready {
 		select {
 		case <-sr.ctx.Done():
 			sr.logger.Error("PostAttachShell: context done", "id", sr.id)
@@ -538,7 +548,13 @@ func (sr *Exec) PostAttachShell() error {
 		sr.logger.Warn("PostAttachShell: state write failed, continuing", "id", sr.id, "err", err)
 	}
 
-	if err := sr.writeStage(&sr.metadata.Spec.Stages.PostAttach); err != nil {
+	// Snapshot the PostAttach stage slice under the read lock: SetupShell
+	// writes Spec.Prompt under metadataMu, so Spec is not immutable and a
+	// lock-free read here would not be excluded by that writer. See #388.
+	sr.metadataMu.RLock()
+	postAttachStages := sr.metadata.Spec.Stages.PostAttach
+	sr.metadataMu.RUnlock()
+	if err := sr.writeStage(&postAttachStages); err != nil {
 		return err
 	}
 
--- a/internal/terminal/terminalrunner/metadata.go
+++ b/internal/terminal/terminalrunner/metadata.go
@@ -264,6 +264,16 @@ func (sr *Exec) updateTerminalAttachers() error {
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


eminwux:
**PR #405 Review — `fix: route terminal metadata accesses in SetupShell/PostAttachShell through metadataMu`**

Correct, well-scoped data-race fix. `SetupShell` now holds `metadataMu` around the `Spec.Prompt` RMW and snapshots the resolved prompt before the lock-free PTY write; `PostAttachShell` uses the `currentState()` RLock'd accessor for the readiness spin and snapshots `Spec.Stages.PostAttach` under RLock. `writeStage` is read-only over its arg, so the snapshot is behaviorally identical to the old in-place pointer. Lock discipline is uniform across both functions, so AC #1's "no unlocked reads or writes" invariant holds. Both `-race` tests were demonstrated to fail pre-fix and pass post-fix, and `make test-race` is green. No leaks, no dead code, all commits verified.

One non-blocking nit:

- **Stale comment** `internal/terminal/terminalrunner/metadata_lock_race_test.go:88`: doc comment says "the fix reads via the `state()` accessor", but the accessor was renamed to `currentState()` (per the PR body's rename note). Cosmetic — not a bug.

LGTM.

PR #406, opened 2026-06-10
fix: synchronize terminal startup fields against Stop-during-startup Close and close orphaned-child hole

### LINKED ISSUE #396: Terminal - Stop RPC during startup races StartTerminal: unsynchronized sr.cmd/ptmx/capture (-race confirmed) plus orphaned-child window via no-op cmd.Cancel
## Summary

A `Stop` RPC during terminal startup runs `Exec.Close` concurrently with `StartTerminal` with **no synchronization at all** on the runtime fields they share. The RPC server is up and answering before `StartTerminal` runs (`internal/terminal/controller.go:148-155`), and `Stop` spawns `go c.Close(...)` (`controller.go:352-363`), so `Close` can interleave anywhere inside `prepareTerminalCommand`/`startPty`. `go test -race` flags the access pairs immediately.

Beyond the races, the interleaving has a concrete liveness hole: if `Close` runs `shutdownChild` while `cmd.Process` is still nil (`lifecycle.go:173` returns early), and `startPty`'s `cmd.Start()` then wins the race against `ctxCancel` (`os/exec.Start` only refuses if the ctx is already done at entry), nothing ever signals the child: `cmd.Cancel` is deliberately a no-op (`terminal.go:96`), and `Close` has already passed its only kill site. The child shell survives the daemon's shutdown path; it is only mopped up if the PTY-master close (SIGHUP) happens to reach it and it doesn't ignore SIGHUP — a child that ignores SIGHUP (daemons, `nohup`-style workloads) is orphaned indefinitely.

## Repro

Throwaway in-package test interleaving the two calls on independent goroutines (the exact shape the Stop RPC produces), scanning the startup window with staggered delays:

```go
go func() { _ = sr.StartTerminal(evCh) }()
go func() { time.Sleep(stagger); _ = sr.Close(nil) }()
```

```
AUDIT_C2=1 go test -race -run TestAuditC2 -count=1 ./internal/terminal/terminalrunner/
```

## Observed

8 distinct `WARNING: DATA RACE` reports in one run, including:

```
WARNING: DATA RACE
Read at 0x00c000246300 by goroutine 12:
  (*Exec).shutdownChild()  internal/terminal/terminalrunner/lifecycle.go:173
  (*Exec).Close()          internal/terminal/terminalrunner/lifecycle.go:223
Previous write at 0x00c000246300 by goroutine 11:
  (*Exec).prepareTerminalCommand() internal/terminal/terminalrunner/terminal.go:110
  (*Exec).StartTerminal()          internal/terminal/terminalrunner/lifecycle.go:59
```

plus the same pattern for `cmd`'s internals (`os/exec.Command` write vs `shutdownChild` read at `lifecycle.go:183`), `sr.capture` (`startPty` write `terminal.go:220` vs `Close` read `lifecycle.go:301-303`), and `internal/poll.FD` close races on the PTY descriptors.

## Root Cause

No mutex covers the runtime fields written by the startup path and read by `Close`:

- writes: `sr.cmd` (`terminal.go:110`), `sr.ptmx`/`sr.pts` (`terminal.go:142-143`), `sr.stopSignalForwarder` (`terminal.go:182`), `sr.capture` (`terminal.go:220`)
- reads (and a write, for `stopSignalForwarder`): `lifecycle.go:173,183,223,230-233,287,301`

The orphan hole is the combination of `shutdownChild`'s early return on `cmd.Process == nil` (`lifecycle.go:173-175`), the no-op `cmd.Cancel` (`terminal.go:96`), and `Close` never re-checking the child after `ctxCancel`.

## Suggested Fix

- Guard the runtime fields with a mutex (or publish them once via the existing `ptyPipesMu`-style pattern) so `Close` reads a consistent snapshot.
- Close the orphan hole by making the shutdown ordering explicit: either serialize `Close` against startup (don't run `shutdownChild` until `StartTerminal` has passed `cmd.Start`, e.g. a "started" latch), or re-arm `cmd.Cancel` so ctx-cancel kills a child that was forked after `shutdownChild`'s early return.

## Acceptance Criteria

- [ ] `go test -race` with a Stop-during-startup interleaving test reports no races on `sr.cmd`/`sr.ptmx`/`sr.capture`/`sr.stopSignalForwarder`.
- [ ] A child forked concurrently with `Close` is always terminated (no surviving process after `Close` returns, even for a SIGHUP/SIGTERM-ignoring child).

## References

- Entry: `internal/terminal/controller.go:148-155` (server-before-StartTerminal), `controller.go:352-363` (`Stop` → `go c.Close`)
- Same window, same subsystem: #388 (sr.metadata races), and the Attach-before-startPty panic issue filed alongside this audit.
- `os/exec` semantics: `Start()` refuses only when ctx is already done at entry; after that, cancellation goes through `cmd.Cancel`, which `terminal.go:96` disables by design.

eminwux:
## Summary

A `Stop` RPC during terminal startup runs `Exec.Close` on its own goroutine (`controller.go:352-363`'s `Stop` → `go c.Close`) while `StartTerminal` is still inside `prepareTerminalCommand`/`startPty`. The RPC server is answering before `StartTerminal` runs (`controller.go:148-155`), so `Close` can interleave anywhere in the startup path. Two defects fall out of that, both fixed here (#396):

1. **Unsynchronized runtime fields.** `Close` read `sr.cmd` / `sr.ptmx` / `sr.capture` / `sr.stopSignalForwarder` while the startup path wrote them, with no lock — `go test -race` flagged the write/read pairs.
2. **Orphaned-child hole.** If `shutdownChild` ran while `cmd.Process` was still nil it returned early, `Close` then called `ctxCancel` and passed its only kill site, and a `cmd.Start` that won the race forked a child nothing would ever signal (`cmd.Cancel` is a deliberate no-op). A SIGHUP/SIGTERM-ignoring child was then orphaned indefinitely.

### Fix

- Added `runtimeMu sync.Mutex` guarding the runtime fields published during startup (`cmd`, `ptmx`, `pts`, `capture`, `stopSignalForwarder`). The startup writes and `Close`'s reads now both take it, so `Close` reads a consistent snapshot. Reads on the startup goroutine itself, and reads in goroutines started *after* a field's publish (`watchChildExit`, `terminalManager*`), need no lock — the `go` statement is already a happens-before edge, so only the cross-goroutine `Close` reads were racing.
- Closed the orphan hole with a `closing` latch (chosen over re-arming `cmd.Cancel`, which would have SIGKILLed a SIGTERM-respecting child on any parent-ctx cancellation and regressed graceful shutdown). `Close` sets `closing` under `runtimeMu` before its graceful sweep; `startPty` checks it under the same lock in the *same critical section* that runs `cmd.Start`. Either `startPty` observes `closing` and aborts before the fork, or it has already published `cmd.Process` for `Close`'s `shutdownChild` to SIGTERM→SIGKILL — there is no interleaving where a child exists but neither path sees it. The pre-existing graceful path is unchanged: `shutdownChild` completes fully before `ctxCancel`.

## Test plan

- [x] `make test-race` — green (AC item 3); the named `internal/terminal/...` package included.
- [x] `make test` — full CI suite green, including `./e2e`.
- [x] `go build ./...`, `go vet ./...`, `make sbsh-sb` (verified ELF) — all clean.
- [x] New `lifecycle_startup_close_race_test.go`, run under `-race`:
  - `TestStartTerminal_StopDuringStartup_NoRace` — sweeps the close timing across the startup window with a real PTY + `/bin/cat` child; race detector stays silent (AC item 1).
  - `TestStartPty_AbortsWhenClosing_NoOrphanFork` — deterministic proof the `closing` latch makes `startPty` abort before forking (AC item 2).
  - `TestStartTerminal_StopDuringStartup_SignalIgnoringChildAlwaysDies` — swept interleaving with a `trap '' TERM HUP INT` child; asserts no process in the child's group survives `Close` (AC item 2, uncatchable-signal case).

These are real instrumented tests exercising the runtime path (live goroutines, fork, PTY) under `-race`, not pure-function unit tests. The pre-fix `-race` failure is documented in the issue (8 DATA RACE reports); these tests are the positive confirmation the path is now clean.

## Acceptance criteria

- [x] `go test -race` with a Stop-during-startup interleaving test reports no races on `sr.cmd`/`sr.ptmx`/`sr.capture`/`sr.stopSignalForwarder`.
- [x] A child forked concurrently with `Close` is always terminated (no surviving process after `Close` returns, even for a SIGHUP/SIGTERM-ignoring child).

## Notes for the reviewer

- The suggested-fix menu offered two options for the orphan hole; I took the `closing`-latch route (option a) rather than re-arming `cmd.Cancel` (option b). Re-arming `cmd.Cancel` to kill the process group fires the os/exec watchdog on *any* `sr.ctx` cancellation — including a parent-context cancellation during normal daemon shutdown — which would SIGKILL a SIGTERM-respecting child and bypass the graceful `shutdownChild` sequence. The latch is surgical to the startup race and leaves graceful shutdown intact.

Closes #396

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -170,7 +170,13 @@ func (sr *Exec) watchChildExit() {
 // Grace <= 0 falls back to the package default. Must only be called after
 // startPty has run (cmd and cmd.Process are set).
 func (sr *Exec) shutdownChild(grace time.Duration) {
-	if sr.cmd == nil || sr.cmd.Process == nil {
+	// Snapshot cmd under runtimeMu: a concurrent startPty publishes it (and
+	// its Process) under the same lock, so this read is otherwise a data
+	// race against a Stop-during-startup Close. See #396.
+	sr.runtimeMu.Lock()
+	cmd := sr.cmd
+	sr.runtimeMu.Unlock()
+	if cmd == nil || cmd.Process == nil {
 		return
 	}
 	if sr.childDone() {
@@ -180,7 +186,7 @@ func (sr *Exec) shutdownChild(grace time.Duration) {
 		grace = profile.DefaultShutdownGrace
 	}
 
-	pid := sr.cmd.Process.Pid
+	pid := cmd.Process.Pid
 	// Setsid: true ensures pgid == pid for the child.
 	if err := syscall.Kill(-pid, syscall.SIGTERM); err != nil && !errors.Is(err, syscall.ESRCH) {
 		sr.logger.Warn("could not SIGTERM child process group", "id", sr.id, "pid", pid, "err", err)
@@ -216,6 +222,16 @@ func (sr *Exec) Close(reason error) error {
 		}
 	}
 
+	// Latch `closing` before the graceful child shutdown so a startPty
+	// racing this Close (Stop RPC during startup) observes it under
+	// runtimeMu and aborts before cmd.Start. This closes the orphaned-child
+	// hole: a child forked after shutdownChild's early return (cmd.Process
+	// still nil) would never be signalled, since cmd.Cancel is a deliberate
+	// no-op and Close has already passed its only kill site. See #396.
+	sr.runtimeMu.Lock()
+	sr.closing = true
+	sr.runtimeMu.Unlock()
+
 	// Graceful shutdown of the child before tearing anything else down:
 	// SIGTERM -> wait up to ShutdownGrace -> SIGKILL. This runs regardless
 	// of PID-1 status; removing the always-hard-kill default is correct in
@@ -226,10 +242,14 @@ func (sr *Exec) Close(reason error) error {
 
 	sr.logger.Debug("terminal closed", "id", sr.id)
 
-	// Tear down PID-1 helpers (no-ops outside init mode).
-	if sr.stopSignalForwarder != nil {
-		sr.stopSignalForwarder()
-		sr.stopSignalForwarder = nil
+	// Tear down PID-1 helpers (no-ops outside init mode). Read/clear under
+	// runtimeMu: startPty publishes stopSignalForwarder concurrently. See #396.
+	sr.runtimeMu.Lock()
+	stopForwarder := sr.stopSignalForwarder
+	sr.stopSignalForwarder = nil
+	sr.runtimeMu.Unlock()
+	if stopForwarder != nil {
+		stopForwarder()
 	}
 	if sr.reaper != nil {
 		sr.reaper.Stop()
@@ -284,10 +304,16 @@ func (sr *Exec) Close(reason error) error {
 	// close all output subscribers
 	sr.closeAllSubscribers()
 
-	if sr.ptmx != nil {
+	// Snapshot ptmx under runtimeMu: startPty publishes it concurrently. The
+	// closePTY sync.Once still serializes the actual close against the
+	// terminalManager goroutines' own ctx-cancel close. See #396.
+	sr.runtimeMu.Lock()
+	ptmx := sr.ptmx
+	sr.runtimeMu.Unlock()
+	if ptmx != nil {
 		var err error
 		sr.closePTY.Do(func() {
-			err = sr.ptmx.Close()
+			err = ptmx.Close()
 		})
 		if err != nil {
 			sr.logger.Warn("could not close pty", "id", sr.id, "err", err)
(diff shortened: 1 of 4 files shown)


eminwux:
**PR #406 Review — `fix: synchronize terminal startup fields against Stop-during-startup Close and close orphaned-child hole`**

LGTM — `runtimeMu` correctly serializes the startup field publishes against `Close`'s reads, and the `closing` latch is checked in the same critical section as `cmd.Start`, so a Stop-during-startup either forks-then-gets-reaped or aborts-before-fork — no interleaving orphans a child. Real instrumented `-race` tests cover both ACs; commits verified, CI green.

PR #407, opened 2026-06-10
fix: guard Attach against nil multiOutW before startPty publishes ptyPipes

### LINKED ISSUE #395: Terminal - Attach RPC before startPty publishes ptyPipes panics on nil multiOutW and crashes the daemon
## Summary

An `Attach` RPC that arrives in the window between the control RPC server becoming ready and `startPty` publishing `ptyPipes` makes `handleClient` call `multiOutW.Add(aw)` on a **nil** `*DynamicMultiWriter`. `Add` dereferences `dmw.logger` on its first statement, and because `handleClient` runs on a bare goroutine spawned by `CreateNewClient` (no `recover`), the nil-pointer panic kills the entire terminal daemon.

The window is real on every start: `internal/terminal/controller.go` starts the RPC server and waits for it to be ready (`controller.go:148-153`) **before** invoking `StartTerminal` (`controller.go:155`), and `pkg/terminal/server` has the same ordering (`server.go:176-186`). `ptyPipes.multiOutW` is only published at `terminal.go:248-253`, after the PTY is opened, the child is forked, and the capture writer is created.

The sibling RPC verb already guards exactly this state: `Subscribe` rejects with "terminal not running" when `multiOutW == nil` (`subscribe.go:95-102`), and both `cleanupClient` (`connections.go:50`) and `Close` (`lifecycle.go:272`) nil-check the same field. `handleClient` is the one consumer that doesn't.

In-tree clients poll `State` until `Ready` before attaching, so they don't normally hit the window — the exposed audience is any raw-RPC embedder driving the documented `pkg/terminal/server` socket directly (its RPC surface is up and answering before the PTY exists), or any client that retries `Attach` aggressively during daemon start.

## Repro

Throwaway in-package test driving the public `Attach` verb against an `Exec` in its pre-`startPty` state (exactly the state `NewTerminalRunnerExec` returns — `ptyPipes: &ptyPipes{}`):

```go
sr := /* Exec as built by NewTerminalRunnerExec, before StartTerminal */
var resp api.ResponseWithFD
_ = sr.Attach(&api.AttachRequest{ClientID: id}, &resp) // returns nil error
// handleClient goroutine panics ~immediately
```

## Observed

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x780701]

goroutine 8 [running]:
github.com/eminwux/sbsh/internal/terminal/terminalrunner.(*DynamicMultiWriter).Add(0x0, {0x96bb40, 0xc0000fea10})
	internal/terminal/terminalrunner/terminal_multiwriter.go:91 +0x41
github.com/eminwux/sbsh/internal/terminal/terminalrunner.(*Exec).handleClient(0xc0000c5208, 0xc0001883f0)
	internal/terminal/terminalrunner/connections.go:161 +0x38b
created by github.com/eminwux/sbsh/internal/terminal/terminalrunner.(*Exec).CreateNewClient in goroutine 7
	internal/terminal/terminalrunner/sockets.go:276 +0x4d3
```

Unrecovered panic on a bare goroutine — the daemon process dies, taking the live session with it.

## Root Cause

- `connections.go:104-107` snapshots `sr.ptyPipes.multiOutW` (nil before `startPty` runs) without the nil-check its siblings have.
- `connections.go:161` calls `multiOutW.Add(aw)`.
- `terminal_multiwriter.go:91` (`Add`) dereferences `dmw.logger` on a nil receiver.
- `sockets.go:276` runs `handleClient` via `go` with no recover, so the panic is fatal to the process.

## Suggested Fix

Mirror the `Subscribe` guard at the top of `handleClient` (or better, in `Attach`/`CreateNewClient` before the socketpair is created, so the client gets a clean RPC error instead of a dead FD): if `multiOutW == nil` (or `pipeInW == nil`), fail the attach with "terminal not running". Failing in `Attach` is preferable — it surfaces a deterministic error to the caller and avoids creating the socketpair at all.

## Acceptance Criteria

- [ ] An `Attach` RPC issued before `StartTerminal` completes returns a clean RPC error ("terminal not running" shape) instead of panicking the daemon.
- [ ] A regression test covers Attach-before-startPty (same shape as the existing `Subscribe` nil-`multiOutW` coverage).

## References

- Window ordering: `internal/terminal/controller.go:148-155`, `pkg/terminal/server/server.go:176-186`
- Publish point: `internal/terminal/terminalrunner/terminal.go:248-253`
- Missing guard: `internal/terminal/terminalrunner/connections.go:104-107,161`
- Existing guards on the same field: `subscribe.go:95-102`, `connections.go:50`, `lifecycle.go:265-272`
- Sibling races in the same startup window: #388 (sr.metadata), and the Stop-during-startup race issue filed alongside this audit.

eminwux:
## Summary

- An `Attach` RPC that arrives in the window between the control RPC server becoming ready and `startPty` publishing `ptyPipes` previously called `multiOutW.Add(aw)` on a **nil** `*DynamicMultiWriter`. `Add` dereferences `dmw.logger`, and because `handleClient` runs on a bare goroutine spawned by `CreateNewClient` with no `recover`, the nil-pointer panic killed the entire terminal daemon.
- This mirrors the sibling `Subscribe` guard (`subscribe.go:95-102`): `Attach` now snapshots `multiOutW` under `ptyPipesMu` and returns `"Attach: terminal not running"` when it is nil. Failing in `Attach` (the issue's preferred location) surfaces a deterministic RPC error to the caller and avoids allocating the socketpair at all — `CreateNewClient` is the sole consumer, so the bare goroutine is never spawned.
- `multiOutW` and `pipeInW` are published together under the same lock at `terminal.go:248-253`, so the single `multiOutW == nil` check is the canonical "not wired yet" gate — same field the existing `Subscribe`, `cleanupClient`, and `Close` guards already nil-check.

## Test plan

- [x] `go vet ./internal/terminal/terminalrunner/...` — clean
- [x] `make test-race` — green (exit 0)
- [x] `make sbsh-sb` — produces an ELF executable (verified magic bytes `7f 45 4c 46`)
- [x] New regression test `TestAttach_NotRunning` — drives `Attach` with `multiOutW` nilled (the pre-`startPty` state), asserts a clean error, no FDs returned, and no client registered. Same shape as the existing `TestSubscribe_NotRunning`.

## Acceptance criteria

- [x] An `Attach` RPC issued before `StartTerminal` completes returns a clean RPC error ("terminal not running" shape) instead of panicking the daemon.
- [x] A regression test covers Attach-before-startPty (same shape as the existing `Subscribe` nil-`multiOutW` coverage).

Closes #395

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -379,6 +379,22 @@ func (sr *Exec) Write(p []byte) (int, error) {
 }
 
 func (sr *Exec) Attach(req *api.AttachRequest, response *api.ResponseWithFD) error {
+	// Reject an Attach that lands in the window between the control RPC
+	// server becoming ready and startPty publishing ptyPipes. multiOutW
+	// (and pipeInW) are published together at terminal.go's
+	// ptyPipes.multiOutW assignment, so a nil multiOutW means the PTY
+	// fan-out is not wired yet. Without this guard handleClient would call
+	// multiOutW.Add on a nil *DynamicMultiWriter and panic on a bare
+	// goroutine with no recover, killing the daemon. Failing here surfaces
+	// a clean RPC error to the caller and avoids allocating the socketpair
+	// at all. Mirrors the sibling Subscribe guard (subscribe.go). See #395.
+	sr.ptyPipesMu.RLock()
+	multiOutW := sr.ptyPipes.multiOutW
+	sr.ptyPipesMu.RUnlock()
+	if multiOutW == nil {
+		return fmt.Errorf("Attach: terminal not running")
+	}
+
 	cliFD, err := sr.CreateNewClient(&req.ClientID, req.FullCapture)
 	if err != nil {
 		return err
--- a/internal/terminal/terminalrunner/socket_rpc_cluster_test.go
+++ b/internal/terminal/terminalrunner/socket_rpc_cluster_test.go
@@ -423,6 +423,35 @@ func TestSubscribe_NotRunning(t *testing.T) {
 	}
 }
 
+// Attach rejects the attach when the PTY fan-out is not wired (an Attach RPC
+// that lands before startPty publishes ptyPipes) and returns a clean
+// "terminal not running" error instead of panicking the daemon on a bare
+// handleClient goroutine. Same shape as TestSubscribe_NotRunning. See #395.
+func TestAttach_NotRunning(t *testing.T) {
+	sr := newWiredExec(t)
+	// Clear the fan-out so Attach's "terminal not running" guard fires —
+	// exactly the pre-startPty state (ptyPipes published but multiOutW nil).
+	sr.ptyPipesMu.Lock()
+	sr.ptyPipes.multiOutW = nil
+	sr.ptyPipesMu.Unlock()
+
+	var resp api.ResponseWithFD
+	err := sr.Attach(&api.AttachRequest{ClientID: api.ID("att-1")}, &resp)
+	if err == nil {
+		t.Fatal("Attach with nil multiOutW returned nil; want terminal-not-running error")
+	}
+	if len(resp.FDs) != 0 {
+		t.Errorf("Attach returned FDs on the not-running path: %v", resp.FDs)
+	}
+	// No client should have been registered (socketpair never allocated).
+	sr.clientsMu.RLock()
+	n := len(sr.clients)
+	sr.clientsMu.RUnlock()
+	if n != 0 {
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #407 Review — `fix: guard Attach against nil multiOutW before startPty publishes ptyPipes`**

LGTM — guards `Attach` against a nil `multiOutW` (pre-`startPty` window) that would otherwise panic the daemon via `handleClient`'s unguarded `multiOutW.Add` on a bare goroutine; mirrors the sibling `Subscribe` guard exactly, covers the sole `CreateNewClient` caller, and adds a regression test matching `TestSubscribe_NotRunning`.

PR #408, opened 2026-06-10
fix: guard nil logger on sb config-error path when --verbose is off

### LINKED ISSUE #387: CLI - sb crashes with a nil-pointer panic (SIGSEGV) on a malformed config when --verbose is off
## Summary

Every `sb` subcommand panics with a nil-pointer dereference (SIGSEGV, full Go stack trace, exit 2) when `$HOME/.sbsh/config.yaml` is malformed and `--verbose`/`-v` is not set. The root pre-run only assigns the logger inside the `if verbose` branch, but the config-error branch dereferences that logger unconditionally. A user-fixable config typo crashes the tool with a stack trace instead of printing the intended `config error` message.

## Repro

```bash
mkdir -p /tmp/cfgtest/.sbsh
printf 'apiVersion: [bad\n' > /tmp/cfgtest/.sbsh/config.yaml
HOME=/tmp/cfgtest ./sb get terminals
```

## Observed

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x550098]

goroutine 1 [running]:
log/slog.(*Logger).Handler(...)
log/slog.(*Logger).Enabled(...)
...
```
Exit code 2. With `-v` set, the same malformed config prints a clean error and exits non-zero without panicking.

## Root cause

`cmd/sb/sb.go:59-94` — in `PersistentPreRunE`:

```go
var logger *slog.Logger
if viper.GetBool(config.SB_ROOT_VERBOSE.ViperKey) {
    ...
    logger = slog.New(handler)   // only assigned here
    ...
}

err := LoadConfig()
if err != nil {
    logger.DebugContext(cmd.Context(), "config error", "error", err)  // line 91: nil deref when -v off
    return fmt.Errorf("%w: %w", errdefs.ErrConfig, err)
}
```

When `--verbose` is off, `logger` is nil at line 91. The `sbsh` side (`cmd/sbsh/sbsh.go:70-78`) does not have this bug — it uses `fmt.Fprintln` + `os.Exit(1)`.

## Suggested fix

Guard the `logger.DebugContext` call (`if logger != nil`), or drop the debug log on the error path and just return the wrapped error, or initialize `logger` to a no-op/`slog.Default()` handler before `LoadConfig()`.

## Acceptance criteria

- [ ] A malformed `config.yaml` with `--verbose` off prints a clean `config error` and exits non-zero (no panic, no stack trace).
- [ ] Behavior with `--verbose` on is unchanged.
- [ ] A test exercises the malformed-config + no-verbose path for `sb` and asserts no panic.

## References

- `cmd/sb/sb.go:59-94` (panic at line 91)
- `cmd/sbsh/sbsh.go:70-78` (the correct pattern on the sbsh side)

eminwux:
## Summary

- `sb`'s root `PersistentPreRunE` only assigned `logger` inside the `if verbose` branch, but the config-error branch (`cmd/sb/sb.go:91`) dereferenced it unconditionally — so a malformed `config.yaml` with `--verbose` off triggered a nil-pointer panic (SIGSEGV, full stack trace, exit 2) instead of the intended `config error` message.
- Fix: guard the `logger.DebugContext` call with `if logger != nil`. This is the lowest-blast-radius option of the three suggested in the issue — the debug log still fires when verbose is on (logger set) and is skipped when off, leaving the verbose path byte-for-byte unchanged.

## Test plan

- [x] `make test` — full CI suite (cmd/sb, cmd/sbsh, internal/terminal, internal/client, integration get, e2e) all pass.
- [x] `go vet ./cmd/...` clean.
- [x] Manual repro from the issue (`make sbsh-sb`; malformed config + `HOME` override): `--verbose` off now prints `Error: config error: ...` and exits 1, no panic; `-v` on prints the same clean error and exits 1. Both modes verified.
- [x] New `Test_PersistentPreRunE_NonVerbose_ConfigError` exercises the malformed-config + no-verbose path and asserts no panic (returns `errdefs.ErrConfig`).

## Acceptance criteria

- [x] Malformed `config.yaml` with `--verbose` off prints a clean `config error` and exits non-zero (no panic, no stack trace).
- [x] Behavior with `--verbose` on is unchanged.
- [x] A test exercises the malformed-config + no-verbose path for `sb` and asserts no panic.

Closes #387

### DIFF
--- a/cmd/sb/sb.go
+++ b/cmd/sb/sb.go
@@ -88,7 +88,9 @@ You can see available options and commands with:
 
 			err := LoadConfig()
 			if err != nil {
-				logger.DebugContext(cmd.Context(), "config error", "error", err)
+				if logger != nil {
+					logger.DebugContext(cmd.Context(), "config error", "error", err)
+				}
 				return fmt.Errorf("%w: %w", errdefs.ErrConfig, err)
 			}
 			return nil
--- a/cmd/sb/sb_coverage_test.go
+++ b/cmd/sb/sb_coverage_test.go
@@ -125,6 +125,34 @@ func Test_PersistentPreRunE_Verbose_ConfigError(t *testing.T) {
 	}
 }
 
+func Test_PersistentPreRunE_NonVerbose_ConfigError(t *testing.T) {
+	t.Cleanup(viper.Reset)
+	viper.Reset()
+
+	bad := filepath.Join(t.TempDir(), "config.yaml")
+	if err := os.WriteFile(bad, []byte("\tnot: [valid yaml"), 0o600); err != nil {
+		t.Fatalf("write bad config: %v", err)
+	}
+	t.Setenv(config.SBSH_ROOT_CONFIG_FILE.EnvVar(), bad)
+	_ = config.SBSH_ROOT_CONFIG_FILE.BindEnv()
+	viper.Set(config.SB_ROOT_VERBOSE.ViperKey, false)
+
+	cmd, err := NewSbRootCmd()
+	if err != nil {
+		t.Fatalf("NewSbRootCmd() error = %v", err)
+	}
+	cmd.SetContext(context.Background())
+
+	// With --verbose off the logger is nil; the error path must not panic.
+	err = cmd.PersistentPreRunE(cmd, nil)
+	if err == nil {
+		t.Fatal("expected error but got nil")
+	}
+	if !errors.Is(err, errdefs.ErrConfig) {
+		t.Fatalf("expected '%v'; got: '%v'", errdefs.ErrConfig, err)
+	}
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #408 Review — `fix: guard nil logger on sb config-error path when --verbose is off`**

LGTM — the `if logger != nil` guard at `cmd/sb/sb.go:91` correctly closes the nil-deref on the config-error path with verbose off, leaves the verbose path byte-for-byte unchanged, and the new `Test_PersistentPreRunE_NonVerbose_ConfigError` exercises the exact panic scenario and asserts the `ErrConfig` sentinel.

PR #409, opened 2026-06-10
fix: clean SIGTERM to attached client exits 0, no spurious context-cancelled error

### LINKED ISSUE #401: Client - clean SIGTERM always exits 1 with spurious 'Error: context has been cancelled' (runClient ctx.Done arm bypasses isCleanSessionEnd); latent %!w(<nil>) wrap in rpcDoneCh-nil branch
## Summary

A clean `SIGTERM` (or external `SIGINT`) to an attached `sbsh` client always exits **1** with a spurious `Error: context has been cancelled` line — 20/20 in a measured loop. This contradicts the project's own clean-session-end contract from #380/#381: `runClient`'s `isCleanSessionEnd` explicitly swallows `context.Canceled` (`cmd/sbsh/sbsh.go:548-552`) — but only on the `errCh` select arm. The `ctx.Done()` arm (`sbsh.go:511-523`) unconditionally returns `errdefs.ErrContextDone`, and on a signal the `ctx.Done` arm wins deterministically (it becomes ready the instant the signal lands, while `errCh` is only ready after the controller finishes teardown). So the same event — cancellation of the `signal.NotifyContext` ctx installed at `sbsh.go:233` — is treated as clean on one arm and as a fatal error on the other, and in practice the error arm is the one users see.

Secondary (latent, same region): the client controller's `rpcDoneCh` branch (`internal/client/controller.go:299-305`) wraps the received error unconditionally: `fmt.Errorf("%w: %w", errdefs.ErrRPCServerExited, errRPC)`. On ctx-cancel the accept loop delivers `nil` (`internal/client/clientrunner/control_server.go:76-83`), which would render as `rpc server exited: %!w(<nil>)`. The terminal-side controller has the graceful nil-check this one lacks (`internal/terminal/controller.go:218-226`). Today the malformed error is unobservable — in 22 logged shutdowns the controller's own select always took its `ctx.Done` arm first, and even if `rpcDoneCh` won, `runClient` never reads `errCh` on the signal path — but the branch is one select-ordering quirk away from surfacing and should get the same nil guard while this area is touched.

## Repro

```
$ mkfifo stdin.fifo; (sleep 600 > stdin.fifo &)
$ ./sbsh < stdin.fifo > out.log 2>&1 &
$ sleep 2; kill -TERM $!; wait $!; echo "exit=$?"
exit=1
$ tail -c 60 out.log
...Error: context has been cancelled
```

Looped 20×: every run exits 1 with that error line. The client log shows the clean path each time (`WARN "parent context canceled, shutting down controller"`), never an RPC failure.

## Observed

```
run 1 exit=1 lastline=Error: context has been cancelled
...
run 20 exit=1 lastline=Error: context has been cancelled
```

## Root Cause

- `cmd/sbsh/sbsh.go:511-523`: the `<-ctx.Done()` arm of `runClient`'s select returns `errdefs.ErrContextDone` unconditionally; `isCleanSessionEnd` (which would swallow `context.Canceled`) is only applied to the `errCh` arm at `:527`.
- `cmd/sbsh/sbsh.go:233`: the ctx is `signal.NotifyContext(..., os.Interrupt, syscall.SIGTERM)`, so this arm is precisely the clean-signal path.
- Secondary: `internal/client/controller.go:299-305` wraps a possibly-nil `errRPC` with `%w` (→ `%!w(<nil>)`), unlike `internal/terminal/controller.go:218-226`.

## Suggested Fix

- In `runClient`'s `ctx.Done` arm, treat signal-driven cancellation as a clean end: return `nil` (after `WaitClose`) instead of `errdefs.ErrContextDone`, or route the result through the same `isCleanSessionEnd` filter the `errCh` arm uses. (If a distinct exit code for "terminated by signal" is desired, that's a deliberate UX decision — but printing `Error: context has been cancelled` while swallowing the identical `context.Canceled` on the other arm is not it.)
- Add the terminal-controller-style `errRPC == nil` branch to `internal/client/controller.go:299-305` so a graceful RPC-server exit never renders `%!w(<nil>)`.

## Acceptance Criteria

- [ ] `kill -TERM <attached sbsh client>` ends the session with exit code 0 and no `Error:` line (terminal restore unchanged).
- [ ] The client controller's `rpcDoneCh` branch handles a nil error without producing a `%!w(<nil>)` string.
- [ ] Existing #380/#381 detach/peer-close swallow behavior unchanged.

## References

- `cmd/sbsh/sbsh.go:233,479-552`
- `internal/client/controller.go:262-314` (select), `:299-305` (nil-unsafe wrap)
- `internal/client/clientrunner/control_server.go:72-91` (`doneCh <- nil` on ctx-cancel)
- `internal/terminal/controller.go:218-232` (the nil-aware counterpart)
- Precedent: #380 / #381 (swallow clean session-end sentinels), #392 (SIGHUP/SIGQUIT teardown — distinct: that issue is about terminal-mode restore on unhandled signals, this one is about the exit contract on the handled ones)

eminwux:
## Summary

- A clean `SIGTERM`/`SIGINT` to an attached `sbsh` client deterministically exited **1** with a spurious `Error: context has been cancelled` line (20/20 in the issue's measured loop). The cause: `runClient`'s select has two arms for the same `signal.NotifyContext` cancellation, and they disagreed. The `errCh` arm runs `context.Canceled` through `isCleanSessionEnd` and swallows it (the #380/#381 contract); the `ctx.Done()` arm returned `errdefs.ErrContextDone` unconditionally. On a signal the `ctx.Done()` arm wins deterministically, so users always saw the error path.
- Fix: in the `ctx.Done()` arm, take `ctx.Err()` (== `context.Canceled` on a handled signal) and route it through the same `isCleanSessionEnd` filter — return `nil` on a clean cancellation, while a real `WaitClose` teardown failure still surfaces (wrapped in `ErrContextDone`).
- Secondary (latent): `internal/client/controller.go`'s `rpcDoneCh` branch wrapped a possibly-nil error with `%w` (→ `rpc server exited: %!w(<nil>)` when the accept loop delivers `nil` on ctx-cancel). Added the nil-aware branch already present in `internal/terminal/controller.go:218-226`.

## Test plan

- [x] `make test` — full CI suite (cmd/sb, cmd/sbsh, internal/terminal, internal/client, integration-tagged get, e2e) — all green.
- [x] `make sbsh-sb` — canonical build; binary verified ELF (magic `7f 45 4c 46`).
- [x] `go build ./...`, `go vet ./cmd/... ./internal/client/...` — clean.
- [x] Issue repro on-device: `kill -TERM <attached sbsh client>` → `exit=0`, zero `Error:` lines, terminal restore sequences (`[?1049l` …) unchanged. Looped 5× — 5/5 `exit=0 errlines=0`.

## Acceptance criteria

- [x] `kill -TERM <attached sbsh client>` ends the session with exit code 0 and no `Error:` line (terminal restore unchanged) — verified on-device, 5/5.
- [x] The client controller's `rpcDoneCh` branch handles a nil error without producing a `%!w(<nil>)` string — nil branch added, mirrors the terminal controller.
- [x] Existing #380/#381 detach/peer-close swallow behavior unchanged — `isCleanSessionEnd` reused as-is, errCh arm untouched; `internal/client` + `cmd/sbsh` tests pass.

Closes #401

### DIFF
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -510,17 +510,22 @@ func runClient(
 	logger.DebugContext(ctx, "controller ready, entering client event loop")
 	select {
 	case <-ctx.Done():
-		var err error
 		logger.DebugContext(ctx, "context canceled, waiting for controller to exit")
+		// ctx is signal.NotifyContext(...), so ctx.Err() is context.Canceled
+		// on a SIGTERM/SIGINT. Route it through the same isCleanSessionEnd
+		// filter the errCh arm uses so a handled signal exits 0 with no
+		// spurious `Error: context has been cancelled` line — matching the
+		// clean-session-end contract from #380/#381. A real WaitClose failure
+		// during teardown is not clean and still surfaces.
+		err := ctx.Err()
 		if errC := ctrl.WaitClose(); errC != nil {
 			err = fmt.Errorf("%w: %w", errdefs.ErrWaitOnClose, errC)
 		}
 		logger.DebugContext(ctx, "context canceled, controller exited")
-		if err != nil {
+		if err != nil && !isCleanSessionEnd(err) {
 			return fmt.Errorf("%w: %w", errdefs.ErrContextDone, err)
 		}
-		// return fmt.Errorf("%w: %w", errdefs.ErrContextDone, err)
-		return errdefs.ErrContextDone
+		return nil
 
 	case err := <-errCh:
 		logger.DebugContext(ctx, "controller stopped", "error", err)
--- a/internal/client/controller.go
+++ b/internal/client/controller.go
@@ -297,6 +297,18 @@ func (s *Controller) Run(doc *api.ClientDoc) error {
 			s.handleEvent(ev)
 
 		case errRPC := <-s.rpcDoneCh:
+			// The accept loop delivers nil on ctx-cancel (a graceful RPC
+			// server exit); wrapping it with %w would render
+			// `rpc server exited: %!w(<nil>)`. Mirror the nil-aware branch
+			// in internal/terminal/controller.go.
+			if errRPC == nil {
+				s.logger.Info("rpc server exited gracefully")
+				if errC := s.Close(nil); errC != nil {
+					s.logger.Error("error during Close after rpc server graceful exit", "error", errC)
+					return fmt.Errorf("%w: %w", errdefs.ErrRPCServerExited, errC)
+				}
+				return errdefs.ErrRPCServerExited
+			}
 			s.logger.Error("rpc server exited", "error", errRPC)
 			if errC := s.Close(errRPC); errC != nil {
 				s.logger.Error("error during Close after rpc server failure", "error", errC)


eminwux:
**PR #409 Review — `fix: clean SIGTERM to attached client exits 0, no spurious context-cancelled error`**

LGTM — the `ctx.Done()` arm now routes `ctx.Err()` through the same `isCleanSessionEnd` filter the `errCh` arm already uses (symmetric, swallows `context.Canceled` on a handled signal, still surfaces a real `WaitClose` teardown failure), and the client controller's nil-aware `rpcDoneCh` branch is an exact mirror of `internal/terminal/controller.go` — verified the accept loop does deliver `nil` on ctx-cancel. No new symbols, no leaks, commits verified, CI green, on-device repro 5/5.

PR #412, opened 2026-06-10
fix: wait for SIGKILLed child to be reaped before reaper teardown so PID-1 zombie can't leak

### LINKED ISSUE #398: Terminal - PID-1 reaper.Stop final drain races the kill-to-zombie window: permanent zombie plus watchChildExit parked forever
## Summary

In PID-1 init mode, `Close`'s SIGKILL escalation returns before the child is actually dead, and `reaper.Stop()` then runs its **final** `drainOnce` before the kernel has turned the child into a zombie. After that the SIGCHLD handler is unregistered and the reap loop is gone, so when the child does die:

1. the zombie is never reaped (it persists for as long as the PID-1 process lives), and
2. `trackedExitCh` is never delivered **nor closed** (`deliverTrackedExitLocked` is the only closer), so `watchChildExit` parks forever on the bare receive at `lifecycle.go:130` — which also means `markChildDone`/`closeReqCh` never fire for that runner.

For the one-shot `sbsh terminal` daemon the process exits right after `Close`, so the zombie's lifetime is short. The audience that actually pays is an embedder running `pkg/terminal/server` as a long-lived PID 1 (container init): permanent zombie + permanently parked goroutine per affected shutdown.

## Repro

The race window (SIGKILL → zombie transition vs `reaper.Stop`'s drain) is wide in practice — the throwaway repro hit it on the **first** cycle:

```go
rp := newReaper(logger); rp.Start()
cmd := exec.Command("/bin/sh", "-c", `trap "" TERM; while :; do sleep 1; done`)
cmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
cmd.Start(); rp.RegisterChild(cmd.Process.Pid)
sr := /* Exec with initMode: true, reaper: rp, ShutdownGrace: 50ms */
go sr.watchChildExit()
sr.Close(nil)
// poll /proc/<pid>/stat: state stays 'Z'; goroutine profile shows
// watchChildExit parked at lifecycle.go:130
```

```
AUDIT_C4=1 go test -run TestAuditC4 -count=1 ./internal/terminal/terminalrunner/
```

## Observed

```
BUG CONFIRMED on cycle 0: child 26559 is a permanent zombie after Close; watchChildExit parked=true
goroutine profile: total 4
1 @ ...
#	0x772537  (*Exec).watchChildExit+0x197  internal/terminal/terminalrunner/lifecycle.go:130
```

`/proc/26559/stat` still reported state `Z` 2.5 s after `Close` returned, with the reaper stopped and SIGCHLD unregistered — nothing in the process will ever reap it.

## Root Cause

- `lifecycle.go:200-203`: the SIGKILL branch of `shutdownChild` sends the signal and returns immediately (no wait on `childDoneCh`).
- `lifecycle.go:234-236`: `Close` then calls `sr.reaper.Stop()`.
- `reaper_linux.go:81-91` + `:199-201`: `Stop` closes the stop channel, unregisters SIGCHLD (`signal.Stop`), and the loop performs one final `drainOnce`. If the killed child is not yet a zombie at that instant (`Wait4(-1, WNOHANG)` returns 0), the loop exits and no further reaping ever happens.
- `reaper_linux.go:185-191`: `trackedExitCh` is delivered+closed only by `deliverTrackedExitLocked`; an unreaped child means the channel stays open forever.
- `lifecycle.go:130`: `watchChildExit` does a bare `<-sr.reaper.TrackedExitCh()` with no ctx escape → parked forever.

## Suggested Fix

Make `reaper.Stop` (or `Close`'s init-mode branch) wait for the tracked child to be reaped before tearing the loop down: e.g. after SIGKILL, block on `childDoneCh` (bounded) before `reaper.Stop`; and/or have `Stop` keep draining until `Wait4` reports ECHILD (no children left) rather than stopping on the first empty sweep; and give `watchChildExit`'s receive a `ctx.Done()` escape so the goroutine can never be permanently parked regardless.

## Acceptance Criteria

- [ ] After `Close` returns in init mode, the tracked child is reaped (no `Z` entry under `/proc`) even when it ignored SIGTERM and was SIGKILLed at the grace boundary.
- [ ] No goroutine remains parked in `watchChildExit` after `Close`.

## References

- `internal/terminal/terminalrunner/lifecycle.go:130,200-203,234-236`
- `internal/terminal/terminalrunner/reaper_linux.go:81-91,124-147,185-191,193-207`
- Same lifecycle subsystem: #388, #396, and #397 (its `waitOnTerminal` leak is downstream of the same close sequencing).

eminwux:
## Summary

In PID-1 init mode, `Close`'s SIGKILL escalation returned before the child was a zombie, and `reaper.Stop()` then ran its **final** `drainOnce` before the kernel had transitioned the killed child. The final `Wait4(-1, WNOHANG)` saw nothing, the loop exited, and SIGCHLD was unregistered — so the child (which dies a moment later) is **never reaped** (permanent zombie for the life of the PID-1 process) and `trackedExitCh` is never delivered nor closed, leaving `watchChildExit` parked forever at `lifecycle.go`'s bare `TrackedExitCh()` receive (and `markChildDone`/`closeReqCh` never fire). The audience that pays is a long-lived `pkg/terminal/server` running as container init. Fixes #398.

The fix closes the kill-to-zombie race at its sequencing root, plus a defensive escape:

- **`shutdownChild`: block (bounded) on `childDoneCh` after SIGKILL.** `reaper.Stop()` runs *after* `shutdownChild` returns and *after* the reaper loop is still live, so waiting for the child to actually be reaped keeps the reaper running until the zombie has been harvested. The wait is event-driven (the still-running reaper reaps the zombie → delivers `TrackedExitCh` → `watchChildExit` closes `childDoneCh`), no polling. Bounded by `reapAfterKillGrace` (5s) so an unkillable D-state child cannot wedge `Close`.
- **`watchChildExit`: `ctx.Done()` escape on the `TrackedExitCh()` receive.** Guarantees the goroutine can never park permanently even in the pathological case where the bounded reap wait elapses (`Close` calls `ctxCancel` right after `shutdownChild`). On the normal path the context is still live during the reap wait, so the escape never fires prematurely.

### Why not the other two suggested mechanisms
The issue suggested three options. I implemented the first ("block on `childDoneCh` before `reaper.Stop`") and third ("`ctx.Done()` escape"). I deliberately did **not** implement "have `Stop` keep draining until `Wait4` reports ECHILD": draining-until-ECHILD means "until I have no children at all", which is wrong for a real container init that legitimately has other live children — it would block `Stop` indefinitely. The targeted `childDoneCh` wait reaps the *tracked* child without that hazard.

### In-scope concurrency fix surfaced by the change (please review)
Making `watchChildExit` correctly *proceed* (AC #2) unmasked a latent data race in the close path: `watchChildExit`'s `cmd.Process.Release()` (a write to the shared `*os.Process`) raced `shutdownChild`'s unsynchronized `cmd.Process.Pid` read. Before this fix the race was hidden because `watchChildExit` parked forever and never reached `Release()`. I serialized both `*os.Process` accesses under the existing `runtimeMu` (pid read moved under the lock in `shutdownChild`; `Release()` wrapped in the lock in `watchChildExit`). This is intrinsic to the sequencing the issue is about, not an unrelated cleanup, and `make test-race` requires it to be green.

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make sbsh-sb` (canonical build; verified ELF magic `7f 45 4c 46`, `sb`↔`sbsh` hardlink)
- [x] New regression test `TestShutdownChild_InitMode_ReapsAfterSIGKILL` (Linux) — reconstructs the exact race (SIGTERM-ignored child, real reaper, real `watchChildExit` goroutine), drives the `shutdownChild → ctxCancel → reaper.Stop` close sequence, and asserts on the real runtime observables: the tracked child is reaped (`Kill(pid,0) == ESRCH`, i.e. no zombie) and `watchChildExit` is not parked (`childDoneCh` closed). Run `-count=5 -race`, green. Fails on pre-fix code (zombie persists → `Kill(0)` returns nil).
- [x] `go test ./internal/terminal/terminalrunner/` (full package, non-race)
- [x] `make test-race` (full target: `cmd/sb`, `cmd/sbsh`, `internal/terminal`, `internal/client`, `pkg/...`) — green

## Acceptance criteria
- [x] After `Close` returns in init mode, the tracked child is reaped (no `Z` under `/proc`) even when it ignored SIGTERM and was SIGKILLed at the grace boundary — verified by the new test's `Kill(pid,0) == ESRCH` assertion exercising the real reaper + kill path.
- [x] No goroutine remains parked in `watchChildExit` after `Close` — verified by the `childDoneCh`-closed assertion, plus the `ctx.Done()` escape guaranteeing termination even if the bounded reap wait elapses.

Closes #398

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -34,6 +34,17 @@ import (
 const (
 	deleteTerminalsDir bool = false
 	waitReadyPeriod         = 10 * time.Millisecond
+	// reapAfterKillGrace bounds how long shutdownChild waits for the tracked
+	// child to actually be reaped after the SIGKILL escalation. SIGKILL only
+	// queues the signal; the kernel still has to schedule the child, turn it
+	// into a zombie, and the harvester (the PID-1 reaper in init mode, or
+	// watchChildExit's cmd.Wait otherwise) still has to reap it. Close tears
+	// the reaper down via reaper.Stop right after shutdownChild returns, so
+	// without this wait the reaper's final drain races the kill-to-zombie
+	// window: a late zombie is never reaped and watchChildExit parks forever on
+	// TrackedExitCh. Bounded so an unkillable (D-state) child cannot wedge
+	// Close. See #398.
+	reapAfterKillGrace = 5 * time.Second
 )
 
 func (sr *Exec) StartTerminal(evCh chan<- Event) error {
@@ -127,18 +138,34 @@ func (sr *Exec) watchChildExit() {
 
 	var exitErr error
 	if sr.initMode && sr.reaper != nil {
-		code, ok := <-sr.reaper.TrackedExitCh()
-		if ok {
-			exitErr = fmt.Errorf("shell process exited: code=%d", code)
-		} else {
-			exitErr = errors.New("shell process exited")
+		select {
+		case code, ok := <-sr.reaper.TrackedExitCh():
+			if ok {
+				exitErr = fmt.Errorf("shell process exited: code=%d", code)
+			} else {
+				exitErr = errors.New("shell process exited")
+			}
+		case <-sr.ctx.Done():
+			// Close cancelled the context before the reaper delivered the
+			// tracked exit — e.g. the child became a zombie only after the
+			// post-SIGKILL reap window (reapAfterKillGrace) elapsed. Escape
+			// rather than park forever on TrackedExitCh; the closeReqCh send
+			// below is itself ctx-guarded, so this goroutine always
+			// terminates regardless. See #398.
+			exitErr = errors.New("shell process exited (close cancelled tracked-exit wait)")
 		}
 		// Release the os.Process handle so Go doesn't hold on to the
 		// kernel reference now that the reaper has already reaped the
 		// PID. A subsequent cmd.Wait() would return ECHILD — we skip it.
+		// Take runtimeMu around the Release: it mutates the shared
+		// *os.Process, and shutdownChild reads cmd.Process.Pid under the
+		// same lock, so without it the two race once Close's post-SIGKILL
+		// wait lets this goroutine proceed. See #398.
+		sr.runtimeMu.Lock()
 		if sr.cmd != nil && sr.cmd.Process != nil {
 			_ = sr.cmd.Process.Release()
 		}
+		sr.runtimeMu.Unlock()
 	} else {
 		// cmd.Wait()'s *exec.ExitError carries the real exit code (and signal
 		// info on signaled deaths) — propagate it so EvCmdExited consumers can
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #412 Review — `fix: wait for SIGKILLed child to be reaped before reaper teardown so PID-1 zombie can't leak`**

LGTM — the bounded `childDoneCh` wait in `shutdownChild` keeps the reaper alive to harvest the SIGKILLed zombie before `Close`'s `reaper.Stop`, the `ctx.Done()` escape guarantees `watchChildExit` never parks, and the `runtimeMu`-serialized pid-read/`Release` closes the data race the sequencing fix unmasked; verified by a real `-race` regression test on the actual reaper + kill path.

PR #413, opened 2026-06-10
fix: persist stdin reader so sequential attach reuse keeps input

### LINKED ISSUE #399: Client - abandoned stdin copier survives pkg/attach.Run and steals input from the next sequential Run (documented contract breach)
## Summary

`pkg/attach.Run` documents that it "is safe to call multiple times sequentially from the same process" (`pkg/attach/attach.go:81-84`). It isn't: when `Stdin` is non-pollable — which the real `os.Stdin` is (a blocking fd 0 is never registered with the runtime poller, so `SetReadDeadline` returns `ErrNoDeadline`) — `restoreParentTerminal`'s bounded 500 ms copier wait gives up and **abandons the stdin→socket copier mid-read** (`internal/client/clientrunner/terminal.go:129-147`). The in-code rationale, "the leaked reader is harmless: the process exits immediately after restore", is true for the `sb`/`sbsh` CLIs and false for the library entry point.

The abandoned copier is still parked in `stdin.Read`. The next byte written to stdin is consumed by the **stale** copier and written to the dead session's socket (`dualcopier.readWriteBytes` writes before re-checking ctx — `internal/dualcopier/copier.go:49-131`), i.e. it is stolen from the next sequential `Run`. Measured: a byte written after `Run` returned was gone from the pipe (FIONREAD = 0) — input loss, deterministic in the no-concurrent-reader case, a read race once session 2's copier is also parked.

## Repro

Throwaway test against a **real** in-process `pkg/terminal/server` terminal (`/bin/cat`), with stdin a blocking (non-pollable) pipe — same pollability as `os.Stdin`:

```go
go attach.Run(ctx1, attach.Options{SocketPath: sock, Stdin: stdinR, Stdout: devnull})
time.Sleep(1500 * time.Millisecond)
cancel1(); <-runDone          // Run returns "cleanly"
// (a) goroutine profile still contains dualcopier runReadWriter frames
// (b) write 1 byte to stdinW; FIONREAD on stdinR drops to 0 — consumed by the stale copier
```

```
AUDIT_C5=1 go test -run TestAuditC5 -count=1 ./pkg/attach/
```

## Observed

```
stdin is non-pollable as intended (SetReadDeadline: file type does not support deadline)
attach.Run returned: context has been cancelled: context canceled
copier goroutines: before=0 after-Run-returned=2
bytes still pending in stdin after post-session write: 0 (1 = preserved for next Run, 0 = stolen by stale copier)
```

Leaked goroutine stacks (debug=2 profile): one parked in `os.(*File).Read` → `dualcopier.(*Copier).readWriteBytes` (`copier.go:53`) → `runReadWriter` (`copier.go:198`) — the stdin→socket copier with the escape filter; a second copier parked on a conn read also survived the teardown.

## Root Cause

- `internal/client/clientrunner/terminal.go:114-127`: `SetReadDeadline(time.Now())` is the only unblock mechanism for the stdin copier, and it is a no-op on the non-pollable `os.Stdin`.
- `terminal.go:129-147`: the copier wait is bounded at 500 ms (`copierWaitTimeout`) and then proceeds, leaving the reader parked. The "process exits immediately" assumption baked into the comment does not hold for `pkg/attach` embedders.
- `internal/dualcopier/copier.go:185-214`: `runReadWriter` only checks ctx at the top of the loop; a goroutine parked inside `Read` consumes the next byte and pushes it at the dead socket before noticing shutdown.
- `pkg/attach/attach.go:81-84`: the public contract promises sequential reuse.

## Suggested Fix

Options, not mutually exclusive:
- Stop reading `Stdin` directly: hand the copier a self-pipe/tee or wrap stdin reads in a one-byte hand-off goroutine owned by the package (one persistent reader per *process*, not per `Run`), so a session teardown can detach from it without abandoning the fd mid-read.
- At minimum, document the limitation honestly in `pkg/attach` (sequential reuse not supported with `os.Stdin`) — but that downgrades a published contract, so the structural fix is preferable.

## Acceptance Criteria

- [ ] Two sequential `attach.Run` calls in one process with a non-pollable `Stdin`: no `dualcopier` goroutine from session 1 survives into session 2, and a byte written between the sessions is delivered to session 2 (not lost).

## References

- `internal/client/clientrunner/terminal.go:68-72,114-147`
- `internal/dualcopier/copier.go:49-131,185-214`
- `pkg/attach/attach.go:71-84`
- Context for the bounded wait: #364 (e2e detach/exit timeouts the bound was added for).

eminwux:
## Summary

`pkg/attach.Run` documents that it is "safe to call multiple times sequentially from the same process," but it wasn't (#399). The stdin->socket copier read `sr.stdin` **directly**. On teardown, `restoreParentTerminal` tried to unblock the parked read with `SetReadDeadline(now)` — a no-op on the non-pollable real `os.Stdin` — then gave up after a bounded 500 ms wait and left the copier goroutine parked mid-`Read`. The leaked reader then consumed the next sequential `Run`'s first byte and wrote it at the dead session's socket (input loss / theft).

This change stops reading the descriptor inside the per-session copier. Instead:

- **`internal/client/clientrunner/stdin_pump.go`** (new) — a process-wide persistent reader keyed by the stdin `*os.File`, with one long-lived goroutine that owns the descriptor across sessions. Each session's copier reads through a context-bound `pumpReader` whose `Read` returns promptly on `ctx`-cancel regardless of pollability, so no `dualcopier` goroutine survives teardown. Any byte the pump already pulled off the descriptor is buffered (or held parked on its send) and handed to the **next** session rather than written at the dead socket. The `read` path deterministically prefers an already-cancelled context over a racing pending byte, and stashes a byte for the next consumer if cancellation and delivery race — so a tearing-down session can never forward a stale byte.
- **`io.go`** — the stdin->socket copier now reads the `pumpReader` instead of `sr.stdin`.
- **`terminal.go`** — removed the now-defunct `SetReadDeadline` unblock (the copier no longer reads the descriptor directly; keeping it would have killed the persistent pump on pollable fds). Both copiers now exit on `ctx`-cancel; the bounded wait remains a safety net against a wedged socket read. Refreshed the stale `#364` rationale comments (the "leaked reader is harmless: the process exits immediately" assumption was the documented bug for `pkg/attach` embedders) to describe the pump.

### Implementation notes for the reviewer

- **Behavioral note (documented here, not a regression vs. the bug):** the pump reads stdin in a process-scoped goroutine that outlives a single `Run`. For the `sb`/`sbsh` CLIs (process exits after attach) this is invisible. For a library embedder that does sequential `Run`s this is the fix. For an embedder that does one `Run` and then reads `os.Stdin` itself afterward, the pump may have read-ahead one chunk and hold it — this is inherent to any "persistent reader" design (the issue's recommended fix) and is strictly better than the prior byte-theft. Verified the held byte is delivered to the next consumer, not dropped.
- The `withRawFd` poller-preservation machinery (`#364`) is retained unchanged; only the SetReadDeadline-based unblock comment lineage was updated.

## Test plan

- [x] `make test-race` (the AC's named target) — green, all packages including `internal/client/...`, `pkg/...`.
- [x] `make test` — green, including the `e2e` detach/exit suite that exercises the real attach teardown paths (`#364`).
- [x] `pre-commit run --files <changed>` — all hooks pass (gofmt, golangci-lint, addlicense).
- [x] New `TestStdinPump_SequentialAttachReuse` (clientrunner) — drives two sequential `startConnectionManager` + `restoreParentTerminal` cycles over a non-pollable `os.Pipe` stdin and asserts (a) no `dualcopier` goroutine from session 1 survives teardown, (b) a byte written between sessions is delivered to session 2's socket. This is the AC at the layer the bug lives in (the `pkg/attach` fake-controller harness returns `attachErr` before reaching the copier, so the I/O path isn't reachable there).
- [x] New `TestStdinPump_ByteSurvivesConsumerCancellation` — pump-level invariant: cancel returns promptly without consuming; a byte produced while no consumer is active reaches the next consumer.

## Acceptance criteria

- [x] Two sequential `attach.Run` calls in one process with a non-pollable `Stdin`: no `dualcopier` goroutine from session 1 survives into session 2, and a byte written between the sessions is delivered to session 2 (not lost). — verified by `TestStdinPump_SequentialAttachReuse`.

Closes #399

### DIFF
--- a/internal/client/clientrunner/io.go
+++ b/internal/client/clientrunner/io.go
@@ -93,9 +93,14 @@ func (sr *Exec) startConnectionManager() error {
 		})
 	}
 
-	// WRITER: stdin -> socket
+	// WRITER: stdin -> socket. Read through the process-wide persistent pump
+	// rather than sr.stdin directly so this session's teardown (ctx-cancel)
+	// unblocks the copier without abandoning a blocking read on the descriptor;
+	// the pump survives to hand any already-read byte to the next session. See
+	// issue #399.
+	stdinSrc := &pumpReader{ctx: sr.ctx, pump: pumpForStdin(sr.stdin)}
 	readyWriter := make(chan struct{})
-	go dc.RunCopier(sr.stdin, sr.ioConn, readyWriter, func() {
+	go dc.RunCopier(stdinSrc, sr.ioConn, readyWriter, func() {
 		if uc != nil {
 			sr.logger.DebugContext(sr.ctx, "stdin->socket: closing write side of UnixConn")
 			_ = uc.CloseWrite()
--- a/internal/client/clientrunner/stdin_pump.go
+++ b/internal/client/clientrunner/stdin_pump.go
@@ -1,0 +1,174 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package clientrunner
+
+import (
+	"context"
+	"io"
+	"os"
+	"sync"
+)
+
+// stdinPump is a single persistent reader over a stdin *os.File (typically the
+// process's os.Stdin) that outlives any one attach session. Reading the
+// descriptor in one long-lived, process-scoped goroutine — rather than
+// directly inside each session's stdin->socket copier — is what makes
+// sequential pkg/attach.Run reuse safe (issue #399):
+//
+//   - On teardown a session detaches its consumer via context cancellation
+//     (see pumpReader) instead of abandoning a blocking Read on the
+//     descriptor. The dualcopier goroutine therefore exits promptly and does
+//     not survive into the next session.
+//   - Any byte the pump has already pulled off the descriptor but not yet
+//     handed to a consumer is buffered and delivered to the next consumer
+//     rather than written at the torn-down session's dead socket.
+//
+// The real os.Stdin (a blocking fd 0) is non-pollable: SetReadDeadline is a
+// no-op on it, so the pre-#399 deadline-based unblock could not interrupt the
+// parked read. The bounded copier wait then gave up and left the reader
+// parked, where it consumed and mis-delivered the next session's first byte.
+type stdinPump struct {
+	ch chan pumpResult
+
+	mu       sync.Mutex
+	leftover []byte // bytes read but not yet consumed; survives across sessions
+	readErr  error  // sticky: once the source errors, every read returns it
+}
+
+type pumpResult struct {
+	data []byte
+	err  error
+}
+
+//nolint:gochecknoglobals // process-wide registry: one persistent stdin reader per descriptor, shared across sequential attach sessions
+var (
+	stdinPumpMu       sync.Mutex
+	stdinPumpRegistry = map[*os.File]*stdinPump{}
+)
+
+// pumpForStdin returns the process-wide pump for f, creating and starting it on
+// first use. Keying on the *os.File pointer — not the int fd — avoids aliasing
+// a closed descriptor whose number was later reused by a different file.
+func pumpForStdin(f *os.File) *stdinPump {
+	stdinPumpMu.Lock()
+	defer stdinPumpMu.Unlock()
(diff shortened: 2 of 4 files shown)


eminwux:
**PR #413 Review — `fix: persist stdin reader so sequential attach reuse keeps input`**

Correct fix for the #399 byte-theft. The stated invariant — a tearing-down session never forwards a stale byte — holds at every reachable entry: the prefer-cancelled select plus the post-receive ctx re-check redirect a racing byte into `leftover`, both stash paths return an error (so the cancelled copier exits rather than re-reading), and the unbuffered channel keeps `leftover` strictly ahead of the parked byte so ordering is preserved. No dead code, no leaks, no unverified commits; tests cover both the pump-level cancellation invariant and the sequential-attach AC over a non-pollable `os.Pipe`.

One non-blocking note:

- **Defensive** `internal/client/clientrunner/stdin_pump.go:108`: the `leftover` early-return serves bytes without re-checking `ctx`. Not reachable today — the only `nil`-return path that fills `leftover` is a partial copy, which can't happen since the dualcopier reads with a 32 KiB buffer matching the pump's. A prefer-`ctx` check at the top of `read` would make the invariant hold independent of that buffer coupling.

LGTM to proceed — the note above is non-blocking.

PR #415, opened 2026-06-10
fix: give waitOnTerminal a ctx escape so a dropped closeReqCh send can't leak it

### LINKED ISSUE #397: Terminal - waitOnTerminal goroutine leaks on external Close when watchChildExit's send select takes ctx.Done (19/30 cycles measured)
## Summary

On an externally-driven `Close` that goes through SIGKILL escalation, `watchChildExit`'s send select races `ctxCancel` and — when the runtime picks the `ctx.Done()` arm — `waitOnTerminal` blocks on the unbuffered `closeReqCh` forever. One goroutine leaks per affected `Close`. In a measured run, **19 of 30** New→Start→Close cycles leaked.

The daemon process exits shortly after `Close`, so the CLI never notices — but `pkg/terminal/server` is a documented embeddable facade, and an embedder that cycles terminals in one process accumulates a parked goroutine (plus its `Exec`, never GC-able) per leaked cycle.

## Repro

`shutdownChild`'s SIGKILL path returns **without waiting** for the child to die (`lifecycle.go:195-203`), so `Close` runs `ctxCancel` (`lifecycle.go:225`) before `cmd.Wait` returns. By the time `watchChildExit` reaches its select (`lifecycle.go:157-160`), both arms are ready and the pick is a uniform coin flip. A SIGTERM-ignoring child with a short grace forces that path deterministically:

```go
spec := newLiveSpec(t, "/bin/sh", "-c", `trap "" TERM; while :; do sleep 1; done`)
spec.ShutdownGrace = 50 * time.Millisecond
// New → CreateMetadata → StartTerminal → Close, 30 cycles,
// then count goroutines parked in waitOnTerminal.
```

```
AUDIT_C3=1 go test -run TestAuditC3 -count=1 ./internal/terminal/terminalrunner/
```

## Observed

```
goroutine profile: total 21
19 @ 0x47a7ee 0x413345 0x412ed2 0x7720e5 0x4827c1
#	0x7720e4  (*Exec).waitOnTerminal+0x44  internal/terminal/terminalrunner/lifecycle.go:110
```

19 of 30 cycles left a `waitOnTerminal` goroutine parked on `<-sr.closeReqCh` after every runner had been Closed (2s settle).

## Root Cause

- `lifecycle.go:109-114`: `waitOnTerminal` is the sole receiver and does a bare `<-sr.closeReqCh` with no ctx escape.
- `lifecycle.go:157-160`: `watchChildExit`, the sole sender, selects between `closeReqCh <- exitErr` and `<-sr.ctx.Done()`.
- `terminal_runner_exec.go:190`: the channel is unbuffered.
- On external Close via SIGKILL escalation, `ctxCancel` (`lifecycle.go:225`) is already done when the sender selects → ~50% of the time the exit error is dropped and the receiver parks forever.

## Suggested Fix

Any one of:
- give `waitOnTerminal` the same escape the sender has (`select { case err := <-sr.closeReqCh: ... case <-sr.ctx.Done(): return }`), or
- make `closeReqCh` buffered(1) and have the sender do a plain send (it sends exactly once), or
- close-based signaling instead of a value send.

The first is the smallest, most symmetric change.

## Acceptance Criteria

- [ ] Repeated New→Start→Close cycles (including the SIGKILL-escalation path with a SIGTERM-ignoring child) leave zero goroutines parked in `waitOnTerminal`.

## References

- `internal/terminal/terminalrunner/lifecycle.go:109-114,157-160,195-203,225`
- `internal/terminal/terminalrunner/terminal_runner_exec.go:190`
- Same lifecycle subsystem: #388, #395, #396, and the reaper-zombie issue filed alongside this audit (its leaked `watchChildExit` parks one step earlier in the same chain).

eminwux:
## Summary

- `waitOnTerminal` (the sole receiver on `closeReqCh`) did a bare `<-sr.closeReqCh` with no ctx escape, while its sole sender `watchChildExit` already had a ctx-guarded send select. On an externally-driven `Close` that escalates to SIGKILL, `ctxCancel` is already done by the time the sender selects, so ~50% of the time the sender takes its `ctx.Done()` arm and **drops** the exit value. With the channel unbuffered (`terminal_runner_exec.go:209`), the receiver then parks on the bare receive forever — leaking one `waitOnTerminal` goroutine (plus its `Exec`, never GC-able) per affected `Close`. The issue measured 19/30 New→Start→Close cycles leaking.
- Fix per the issue's first (smallest, most symmetric) suggestion: give `waitOnTerminal` the same `select { case <-closeReqCh / case <-ctx.Done() }` escape the sender has. On a clean exit the value arrives and `EvCmdExited` is published exactly as before; on the dropped-send race the receiver escapes via `ctx.Done()` and returns instead of leaking.

## Note on issue premise (line numbers)

The issue cites `lifecycle.go:109-114,157-160,225`, but #398 (the reaper-zombie fix, now on `main`) shifted those: `waitOnTerminal`'s receive is now at `lifecycle.go:121`, the sender's send select at `185-187`, and `ctxCancel` at `295`. The structure the bug depends on is unchanged — bare receiver vs. ctx-guarded sender on an unbuffered channel — so the diagnosis and fix hold verbatim against current `main`.

## Test plan

- [x] New deterministic regression test `lifecycle_wait_on_terminal_test.go`:
  - `TestWaitOnTerminal_CtxCancel_NoLeak` — unbuffered `closeReqCh` (matching production), cancel ctx without ever sending (the dropped-send race), assert `waitOnTerminal` returns rather than parking. Fails (5s timeout) against the pre-fix bare receive.
  - `TestWaitOnTerminal_NormalExit_PublishesEvent` — clean path unchanged: value delivered → `EvCmdExited` carrying the error published.
- [x] `go build ./...`, `go vet ./internal/terminal/terminalrunner/...`
- [x] `make sbsh-sb` (ELF verified, `sb` hardlinked)
- [x] `make test` (full CI suite incl. e2e) — green
- [x] `make test-race` — green (AC item 2)

## Acceptance criteria

- [x] Repeated New→Start→Close cycles (including the SIGKILL-escalation path with a SIGTERM-ignoring child) leave zero goroutines parked in `waitOnTerminal` — the receiver now has a ctx escape, so a dropped sender send can no longer strand it. Verified deterministically by `TestWaitOnTerminal_CtxCancel_NoLeak` (which reproduces the dropped-send race directly rather than relying on the coin-flip full-cycle repro).

Closes #397

### DIFF
--- a/internal/terminal/terminalrunner/lifecycle.go
+++ b/internal/terminal/terminalrunner/lifecycle.go
@@ -118,7 +118,23 @@ func (sr *Exec) applyLogFilePerms() error {
 }
 
 func (sr *Exec) waitOnTerminal() {
-	err := <-sr.closeReqCh
+	// Mirror the ctx escape watchChildExit's send select already has
+	// (lifecycle.go's closeReqCh send below). closeReqCh is unbuffered and
+	// watchChildExit is its sole sender; on an externally-driven Close that
+	// escalates to SIGKILL, ctxCancel can already be done when the sender
+	// selects, so the sender takes its ctx.Done() arm and drops the value.
+	// Without a symmetric escape here this sole receiver parks on the bare
+	// receive forever, leaking one goroutine (plus its Exec) per affected
+	// Close. The receive and the ctx guard race like the send does; on a
+	// clean exit the value arrives and we publish EvCmdExited as before. See
+	// #397.
+	var err error
+	select {
+	case err = <-sr.closeReqCh:
+	case <-sr.ctx.Done():
+		sr.logger.Debug("waitOnTerminal: ctx cancelled before close request; exiting", "id", sr.id)
+		return
+	}
 	sr.logger.Info("terminal close requested", "id", sr.id, "err", err)
 	sr.logger.Debug("sending EvCmdExited event", "id", sr.id)
 	trySendEvent(sr.logger, sr.evCh, Event{ID: sr.id, Type: EvCmdExited, Err: err, When: time.Now()})
--- a/internal/terminal/terminalrunner/lifecycle_wait_on_terminal_test.go
+++ b/internal/terminal/terminalrunner/lifecycle_wait_on_terminal_test.go
@@ -1,0 +1,112 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package terminalrunner
+
+import (
+	"context"
+	"errors"
+	"io"
+	"log/slog"
+	"testing"
+	"time"
+)
+
+// newWaitOnTerminalTestExec builds a minimal Exec for exercising
+// waitOnTerminal's receive in isolation. closeReqCh is UNBUFFERED to match the
+// production wiring (terminal_runner_exec.go:209) — the leak only manifests on
+// an unbuffered channel where a dropped sender send leaves the receiver with
+// nothing to take.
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #415 Review — `fix: give waitOnTerminal a ctx escape so a dropped closeReqCh send can't leak it`**

LGTM — gives the sole `closeReqCh` receiver the same `ctx.Done()` escape its sole sender already has (`lifecycle.go:201-203`), so a dropped send on an unbuffered channel no longer strands `waitOnTerminal`; no event-contract change (a ctx-cancelled Close already dropped the event), deterministic regression test reproduces the race directly, single verified commit, CI incl. `make test-race` green.

PR #417, opened 2026-06-10
fix: treat os.ErrProcessDone from SIGKILL as stopped, not failed

### LINKED ISSUE #411: sb stop: --kill-after reports failure when process exits on its own between grace deadline and SIGKILL
## Repro

Surfaced as a flaky failure of `Test_StopOneTerminal_KillAfterEscalates` under full-suite parallel load (`go test ./...`). It passes 10/10 in isolation but flaked once when the box was under CPU contention from the rest of the suite — the contention is incidental; it just widens the race window.

```
--- FAIL: Test_StopOneTerminal_KillAfterEscalates (0.32s)
    stop_test.go:186: expected resultStopped, got 2
```

## Observed

```
stdout="terminal \"trap-term\" did not exit within 200ms; escalating to SIGKILL\n"
stderr="failed to SIGKILL terminal \"trap-term\": failed to signal terminal process: os: process already finished\n"
```

The escalation path fires (terminal didn't exit within the 200ms SIGTERM grace), then the SIGKILL fails with `os: process already finished` because the child exited on its own — its SIGTERM trap finally ran — in the window between the grace deadline and the SIGKILL syscall. The code treats that error as a failure and returns `resultFailed` (exit-code path → `got 2`), even though the terminal *is* stopped.

## Root cause

`cmd/sb/stop/terminals.go:316-319`:

```go
if err := sendSignal(doc.Status.Pid, doc.Status.PidStart, syscall.SIGKILL); err != nil {
    fmt.Fprintf(stderr, "failed to SIGKILL terminal %q: %v\n", name, err)
    return resultFailed
}
```

A SIGKILL that fails with `process already finished` / `ESRCH` means the process is gone — which is exactly the goal of the stop. Returning `resultFailed` here is a TOCTOU race: the process satisfied the stop between the `waitForExit` grace check (`:298`) and this kill. The companion pgroup kill at `:312-315` has the same shape and likely the same exposure.

## Suggested fix

In the escalation branch, treat a "process already finished" / `os.ErrProcessDone` / `ESRCH` error from the SIGKILL (and the pgroup kill) as success rather than failure — or re-run `waitForExit` before deciding, so a process that exited on its own at the grace boundary reports `resultStopped`. Detecting the not-found error explicitly (rather than re-checking) keeps the success signal precise. The 200ms grace value is fine; the bug is the error classification, not the timeout.

eminwux:
## Summary
- `sb stop --kill-after` misreported a stop as `resultFailed` when the child exited on its own in the TOCTOU window between the SIGTERM grace deadline (`waitForExit` at `terminals.go:298`) and the escalation SIGKILL (`terminals.go:316`). Surfaced as a flaky `Test_StopOneTerminal_KillAfterEscalates` failure under parallel-suite CPU contention.
- Root cause was error *classification*, not the 200ms grace: `sendSignal` (`terminals.go:398`) swallowed `syscall.ESRCH` from `proc.Signal`, but the pidfd-backed `os.Process.Signal` path (Go 1.23+; this repo builds on go1.24) returns `os.ErrProcessDone` ("os: process already finished") for an already-exited process — exactly the observed stderr — which the ESRCH-only check missed, so the error propagated and the escalation branch returned `resultFailed`.
- Fix extracts the "process is gone" classification into `isProcessGoneErr`, matching both `syscall.ESRCH` and `os.ErrProcessDone`, and uses it in `sendSignal`. Because all SIGKILL callers (graceful SIGTERM fallback, `--force`, `--kill-after` escalation) route through `sendSignal`, fixing it at the source covers every path. A SIGKILL that finds the process already gone is success — the goal of the stop.
- The companion pgroup kill (`sendPgroupKill`, `terminals.go:367`) was already covered: it calls `syscall.Kill` directly, which returns raw `ESRCH` (not `os.ErrProcessDone`) and already suppresses it. No change needed there; confirmed by reading the function.

## Notes on test choice
The bug is an inherently racy TOCTOU window, so the regression guard is a deterministic table test on the extracted `isProcessGoneErr` classifier (the exact fix point) rather than an attempt to reproduce the race. An end-to-end `sendSignal`-on-a-reaped-process test was considered and dropped: it would pass via the earlier `pidutil.Match` → `os.ErrNotExist` short-circuit (`terminals.go:384`), not the `proc.Signal` → `os.ErrProcessDone` branch being fixed, so it would be a misleading guard.

## Test plan
- [x] `make sbsh-sb` (ELF verified — magic `7f 45 4c 46`, `file` not installed on host)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./cmd/sb/stop/...` (incl. new `Test_IsProcessGoneErr`, and `Test_StopOneTerminal_KillAfterEscalates`)
- [x] Full unit suite: `go test -timeout=5m $(go list ./... | grep -v '/e2e$')`
- [x] Integration: `go test -tags=integration ./cmd/sb/get/...`
- [x] E2E: `E2E_BIN_DIR=$(pwd) go test ./e2e`
- [x] `pre-commit run --files` on the changed files (golangci-lint, go fmt, addlicense — all pass)

Closes #411

### DIFF
--- a/cmd/sb/stop/terminals.go
+++ b/cmd/sb/stop/terminals.go
@@ -396,13 +396,25 @@ func sendSignal(pid int, pidStart uint64, sig syscall.Signal) error {
 		return fmt.Errorf("%w: %w", errdefs.ErrSignalProcess, err)
 	}
 	if err := proc.Signal(sig); err != nil {
-		if errors.Is(err, syscall.ESRCH) {
+		if isProcessGoneErr(err) {
 			// Process is already gone — not an error for our purposes.
 			return nil
 		}
 		return fmt.Errorf("%w: %w", errdefs.ErrSignalProcess, err)
 	}
 	return nil
+}
+
+// isProcessGoneErr reports whether err from a signal delivery means the target
+// process has already exited. The raw syscall path (syscall.Kill) surfaces this
+// as ESRCH; the pidfd-backed os.Process.Signal path (Go 1.23+) surfaces it as
+// os.ErrProcessDone ("os: process already finished"). Either way the process is
+// gone, which is exactly the goal of a stop — so callers escalating to SIGKILL
+// must classify it as success, not failure. Without the os.ErrProcessDone arm a
+// child that exits on its own in the window between the grace deadline and the
+// SIGKILL syscall is misreported as resultFailed (a TOCTOU race).
+func isProcessGoneErr(err error) bool {
+	return errors.Is(err, syscall.ESRCH) || errors.Is(err, os.ErrProcessDone)
 }
 
 // processIsOurs reports whether pid is alive AND matches the recorded
--- a/cmd/sb/stop/terminals_test.go
+++ b/cmd/sb/stop/terminals_test.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"log/slog"
 	"net"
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #417 Review — `fix: treat os.ErrProcessDone from SIGKILL as stopped, not failed`**

LGTM — extracting `isProcessGoneErr` to match both `syscall.ESRCH` and `os.ErrProcessDone` correctly closes the pidfd-path TOCTOU gap at the single source (`sendSignal`) all SIGKILL callers route through; `sendPgroupKill`'s raw `syscall.Kill` only ever returns `ESRCH` so it's already covered, the deterministic classifier test guards the exact fix point, and CI is green.

PR #418, opened 2026-06-10
fix: trap SIGHUP/SIGQUIT in attached client so parent terminal modes are restored

### LINKED ISSUE #392: Client - SIGHUP/SIGQUIT terminate an attached client without restoring the parent terminal's modes
## Summary

The attached client (`sb attach` and `sbsh`) traps only `os.Interrupt` and `SIGTERM` via `signal.NotifyContext`. While attached — raw mode plus DEC private modes (mouse-tracking, bracketed-paste, alt-screen) active — a `SIGHUP` or `SIGQUIT` takes the default disposition: the process dies without ctx cancellation, so `restoreParentTerminal` never runs. Whenever the controlling terminal *survives* the signal (a manual `kill -HUP <client-pid>`, or HUP forwarded by a session manager), the parent shell is left in raw mode with mouse-tracking/bracketed-paste still on.

## Repro

```bash
sb attach <terminal>        # raw mode + DEC modes active
kill -HUP <client-pid>      # from another shell (or kill -QUIT)
stty -a                     # shows -icanon -echo: parent left in raw mode
```

## Root cause

- `cmd/sb/attach/attach.go:191` and `cmd/sbsh/sbsh.go:233` — `signal.NotifyContext(ctx, os.Interrupt, syscall.SIGTERM)` traps only INT and TERM; HUP/QUIT are not included, so they hit the default terminate disposition and never cancel the ctx that drives `Close` → `restoreParentTerminal`.
- Contrast: the server side (`internal/terminal/terminalrunner/signals_linux.go:36`) already handles SIGHUP; the client side does not.

## Suggested fix

Add `syscall.SIGHUP` (and `syscall.SIGQUIT`) to the client's `signal.NotifyContext` set so they drive the same ctx-cancel → restore path as INT/TERM. (In the common SIGHUP-with-tty-gone case restore is moot, but it's harmless; in the tty-survives case it's the fix.)

## Acceptance criteria

- [ ] `kill -HUP` / `kill -QUIT` on an attached client restores the parent terminal modes before exit when the tty survives.
- [ ] `stty -a` after the signal shows the terminal back in canonical mode with echo.

## References

- `cmd/sb/attach/attach.go:191`
- `cmd/sbsh/sbsh.go:233`
- `internal/terminal/terminalrunner/signals_linux.go:36` (server-side SIGHUP handling, for parity)
- related: #384 (mode restore on teardown)

eminwux:
## Summary

- The attached client (`sb attach` and `sbsh`) trapped only `os.Interrupt` and `SIGTERM` via `signal.NotifyContext`, so a `SIGHUP` or `SIGQUIT` took the default terminate disposition: the process died without ctx cancellation, `restoreParentTerminal` never ran, and whenever the controlling terminal survived the signal the parent shell was left in raw mode with mouse-tracking/bracketed-paste/alt-screen still active.
- Fix: add `syscall.SIGHUP` and `syscall.SIGQUIT` to the client's `signal.NotifyContext` set at both sites (`cmd/sb/attach/attach.go:191`, `cmd/sbsh/sbsh.go:233`), so HUP/QUIT drive the same ctx-cancel → `restoreParentTerminal` path as INT/TERM. This matches the server-side parity set in `internal/terminal/terminalrunner/signals_linux.go` (`SIGTERM, SIGINT, SIGHUP, SIGQUIT`).

## Verification — instrumented e2e regression test

Added `e2e/e2e_sbsh_signal_test.go` (`TestSbsh_SignalRestoresTerminal`, subtests SIGHUP and SIGQUIT). It spawns `sbsh` on a real pty, asserts the terminal is in **raw** mode while attached, delivers the signal directly to the client, and asserts the pty returns to **canonical mode with echo** (the `stty -a` observable the AC names) — read via `IoctlGetTermios(TCGETS)` on the pty master.

Per `agents/dev/CLAUDE.md` §Verifying behavioral runtime fixes, this is a real instrumented test of the runtime path (not a static trace). I confirmed it has teeth: reverting the fix and rebuilding makes both subtests **fail** (`terminal not restored to canonical mode`); with the fix they pass.

Notes on the test design (documented so the reviewer can audit the non-obvious parts):
- The signal must land in the controller's steady-state event loop, **not** during the earlier `WaitReady` phase. The IO copiers begin relaying while `Attach` is still in progress, so a visible prompt — or even a full command round-trip — is *not* proof of readiness (a signal caught mid-`WaitReady` takes an early-return path that legitimately skips restore and is out of scope for this issue). The test gates on the controller's `"controller ready, entering main event loop"` log line (written to `<run_path>/clients/<id>/log` exactly when `WaitReady` unblocks) — a race-free readiness signal rather than a blind sleep.

## Scope note

The e2e test exercises the `sbsh` client path. The `sb attach` path (`cmd/sb/attach/attach.go`) receives the line-identical edit and funnels through the same `pkg/attach` → client-runner `restoreParentTerminal` teardown driven by the same `signal.NotifyContext` ctx, so the mechanism the test verifies is shared. A dedicated `sb attach` e2e (start a server terminal, then attach) was not added to keep this PR focused; flagging it so the reviewer can request one if desired.

## Test plan

- [x] `make sbsh-sb` (canonical build; ELF verified via magic bytes — `file` not installed on host)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` (full suite, includes the new e2e test)
- [x] Regression test fails without the fix, passes with it (verified by reverting both sites, rebuilding, re-running)

## Acceptance criteria

- [x] `kill -HUP` / `kill -QUIT` on an attached client restores the parent terminal modes before exit when the tty survives — verified by `TestSbsh_SignalRestoresTerminal`.
- [x] `stty -a` after the signal shows the terminal back in canonical mode with echo — the test asserts exactly `ICANON` + `ECHO` re-set on the pty.

Closes #392

### DIFF
--- a/cmd/sb/attach/attach.go
+++ b/cmd/sb/attach/attach.go
@@ -186,9 +186,12 @@ func run(
 		return errdefs.ErrLoggerNotFound
 	}
 
-	// Top-level context also reacts to SIGINT/SIGTERM (nice UX). pkg/attach
-	// deliberately does not trap signals itself; the CLI owns that policy.
-	ctx, cancel := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGTERM)
+	// Top-level context also reacts to SIGINT/SIGTERM/SIGHUP/SIGQUIT (nice UX).
+	// pkg/attach deliberately does not trap signals itself; the CLI owns that
+	// policy. Trapping HUP/QUIT drives the same ctx-cancel → restoreParentTerminal
+	// path as INT/TERM, so a HUP/QUIT whose tty survives doesn't leave the parent
+	// shell in raw mode with DEC private modes still active.
+	ctx, cancel := signal.NotifyContext(cmd.Context(), os.Interrupt, syscall.SIGTERM, syscall.SIGHUP, syscall.SIGQUIT)
 	defer cancel()
 
 	socketPath, err := resolveTerminalSocket(cmd, logger, args)
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -230,7 +230,10 @@ You can also use sbsh with parameters. For example:
 				}
 			}
 
-			ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
+			// Trap HUP/QUIT alongside INT/TERM so they drive the same ctx-cancel →
+			// restoreParentTerminal path; otherwise a HUP/QUIT whose tty survives
+			// leaves the parent shell in raw mode with DEC private modes active.
+			ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM, syscall.SIGHUP, syscall.SIGQUIT)
 			ctrl := client.NewClientController(ctx, logger)
 
 			return runClient(ctx, cancel, logger, ctrl, doc)
--- a/e2e/e2e_sbsh_signal_test.go
+++ b/e2e/e2e_sbsh_signal_test.go
@@ -1,0 +1,191 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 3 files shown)


eminwux:
**PR #418 Review — `fix: trap SIGHUP/SIGQUIT in attached client so parent terminal modes are restored`**

LGTM — symmetric two-line addition of `SIGHUP`/`SIGQUIT` to the client `signal.NotifyContext` sets so HUP/QUIT drive the same ctx-cancel → `restoreParentTerminal` teardown as INT/TERM; matches the server-side parity set, `syscall` already imported at both sites, and the new pty-based e2e test delivers the real signals and asserts canonical-mode restore via `TCGETS` (confirmed to fail on revert). The two unchanged `NotifyContext` sites (`sb read`, `sbsh terminal` server) are a different class — no parent-terminal raw mode, no restore dependency — so the fix is complete.

PR #420, opened 2026-06-10
fix: restore parent terminal when signal cancels ctx during attach window

### LINKED ISSUE #419: Client - signal during attach/WaitReady window skips terminal restore, stranding parent in raw mode
## Summary

A signal that cancels the client's context **during the attach window** — after the client runner has put the terminal into raw mode but before `runClient`'s `ctrl.WaitReady()` returns — takes the `ErrWaitOnReady` early-return path and exits **without** running `restoreParentTerminal`. The parent shell is left stranded in raw mode (and with DEC private modes active), exactly the failure class #392 addresses for the steady-state case, but on a different, earlier code path.

This affects **all** the trapped signals (SIGINT, SIGTERM, and — after #392 — SIGHUP/SIGQUIT): the gap is in the readiness-wait early return, not in which signals are trapped. It is **not** fixed by #392 (whose fix only adds HUP/QUIT to the trapped set; its regression test deliberately gates on steady-state readiness to avoid this window).

## Repro

Minimal Go e2e against `HEAD` (worktree `main` @ 81e3539, with or without #392):

1. Spawn `sbsh` on a pty (`Setsid`/`Setctty`, pts as fds 0/1/2).
2. Wait for the shell prompt to render (the visible prompt appears ~milliseconds *before* `WaitReady` returns — the IO copiers relay output while `Attach` is still in progress).
3. Immediately deliver `SIGTERM` to the `sbsh` process (so it lands in the `WaitReady` window).
4. Read the pty master's termios after the process exits.

(Reproduced this session with a throwaway e2e test; signalling on prompt-detection reliably lands in the window. Signalling ~1s later instead lands in the steady-state loop and restore works — confirming the window is the trigger.)

## Observed

```
REPRO after SIGTERM: icanon=false echo=false   (terminal stranded in raw mode)
REPRO tail output: "...$ Error: waiting for readiness has failed: context canceled\n"
```

The `Error: waiting for readiness has failed: context canceled` line is the tell: control returned via `cmd/sbsh/sbsh.go:508-511` (the `ctrl.WaitReady()` error path), which returns immediately without the ctx.Done arm's `WaitClose`/restore teardown. `icanon=false` confirms the terminal was never restored to canonical mode.

## Root cause

`cmd/sbsh/sbsh.go:508-511`:

```go
if err := ctrl.WaitReady(); err != nil {
    logger.DebugContext(ctx, "controller not ready", "error", err)
    return fmt.Errorf("%w: %w", errdefs.ErrWaitOnReady, err)   // <- returns without restore
}
```

`restoreParentTerminal` is driven from the steady-state `select` loop's `ctx.Done` arm (`sbsh.go:514-531`) via `ctrl.WaitClose()`. When the signal cancels ctx *before* `WaitReady()` unblocks, `WaitReady()` returns `context.Canceled`, `runClient` returns the wrapped `ErrWaitOnReady`, and the steady-state arm is never reached. Raw mode was already entered inside `Attach` (the copiers were relaying), so the terminal is left raw. The controller's own teardown on ctx-cancel during attach (`internal/client/controller.go` `Close` paths) does not restore the parent terminal either.

The `sb attach` client (`pkg/attach`) should be audited for the analogous window.

## Suggested fix

Ensure the terminal-restore teardown runs on the `WaitReady`-error early-return path too, not only from the steady-state `ctx.Done` arm — e.g. route the `ErrWaitOnReady` return through the same `WaitClose`/`isCleanSessionEnd` teardown that the `ctx.Done` arm uses, or make the controller's attach-window `Close` path invoke `restoreParentTerminal` (it is `sync.Once`-guarded, so double-invocation is safe). The window is brief (~sub-second) so manual `kill` rarely hits it, but a session manager forwarding a signal at startup, or fast automated teardown, can.

## References

- Surfaced alongside #392 (steady-state SIGHUP/SIGQUIT restore) while building its e2e regression test.
- `cmd/sbsh/sbsh.go:508-511` (early return), `sbsh.go:514-531` (steady-state restore arm).
- `internal/client/controller.go:249-260` (`Attach` then `close(ctrlReadyCh)`).
- Related but distinct: #389 (detach RPC deadline — different mechanism).

eminwux:
## Summary

- A signal that cancels the client's context **during the attach window** — after the controller's `Attach` has put the parent terminal into raw mode (the IO copiers are already relaying the prompt) but before `runClient`'s `ctrl.WaitReady()` returns — took the `ErrWaitOnReady` early-return path at `cmd/sbsh/sbsh.go` and exited **without** driving the terminal-restore teardown. The parent shell was left stranded in raw mode. This is the attach-window analogue of the steady-state stranding #392 addresses; it is a different, earlier code path and is not fixed by #392.
- `WaitReady` only returns non-nil via `Controller.WaitReady`'s `ctx.Done` arm (`internal/client/controller.go:108-115`), i.e. `context.Canceled` from a signal landing in the window. The early return wrapped it in `ErrWaitOnReady` and returned immediately, never reaching the steady-state `ctx.Done` arm that drives `ctrl.WaitClose()` → controller `Close` → runner `Close` → `restoreParentTerminal`.

## Fix

Route the `WaitReady`-error path through the **same `WaitClose`/`isCleanSessionEnd` teardown the adjacent `ctx.Done` arm already uses** (the issue's first suggested option). When the controller's `Run` goroutine unblocks `Attach` on the cancelled ctx, its `Close` restores the parent terminal *before* closing the channel `WaitClose` blocks on (`controller.go:444` runs `sr.Close` ahead of `close(closedCh)` at `:450`), so on `WaitClose` return the restore is guaranteed complete. A handled signal (`context.Canceled`) then exits 0 like the steady-state arm, matching the clean-session-end contract from #380/#381; a genuine readiness failure still surfaces wrapped in `ErrWaitOnReady`.

No deadlock risk: `Attach`'s only blocking steps are `waitReady` calls that honour `sr.ctx` (bounded by `waitReadyTimeout`), so `Attach` reliably returns on ctx-cancel and `Run` reaches `Close` (via the `Attach`-error path at `controller.go:249` or, on a racing `Attach` success, the loop's `ctx.Done` arm).

### `pkg/attach` audit (the issue asked to audit it)

`pkg/attach.Run` already handles its analogous window correctly: on `WaitReady` error it calls `ctrl.Close(waitErr)` then drains `<-errCh` (`pkg/attach/attach.go:152-164`), and `Close` drives `restoreParentTerminal`. No change needed there. The gap was specific to `cmd/sbsh`, which relies on `Run` to drive `Close` rather than calling it proactively — fixed here by driving `WaitClose` on the early-return path, consistent with `cmd/sbsh`'s own steady-state arm.

## Verification basis

The fix's contract is *"the `WaitReady`-cancel path drives `WaitClose` and exits 0 instead of returning a spurious `ErrWaitOnReady`"*. The new fake-controller unit test observes exactly that contract (it asserts `WaitClose` is invoked and `runClient` returns nil). The actual raw-mode `restoreParentTerminal` it ultimately drives is exercised by the runner's existing terminal-restore tests; this PR wires `cmd/sbsh`'s early-return path into that same teardown.

The issue's repro is a pty-level e2e (signal-on-prompt-detection lands in the window). This repo's `cmd/sbsh` suite has no pty-e2e harness for that path, so the end-to-end raw-mode observation the issue describes is **not** reproduced as an automated test here — the regression is covered at the controller-contract tier instead. Flagged for transparency per the runtime-behavioral-verification bar.

## Test plan

- [x] `make sbsh-sb` — canonical build; binary verified ELF (`7f 45 4c 46`) and runnable (`sbsh version`)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — full suite, `GOTEST-EXIT=0`, zero FAIL lines
- [x] New regression test `TestRunTerminal_WaitReadyContextCanceledDrivesTeardown` passes; existing `TestRunTerminal_ErrWaitOnReady` (non-clean error path) still passes
- [ ] pty-level e2e reproducing the stranded-raw-mode observable — no pty-e2e harness exists for `cmd/sbsh`'s client path; covered at the controller-contract tier instead (see Verification basis)

Closes #419

### DIFF
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -503,8 +503,26 @@ func runClient(
 
 	logger.DebugContext(ctx, "waiting for controller to signal ready")
 	if err := ctrl.WaitReady(); err != nil {
-		logger.DebugContext(ctx, "controller not ready", "error", err)
-		return fmt.Errorf("%w: %w", errdefs.ErrWaitOnReady, err)
+		// A signal that cancels ctx during the attach window makes WaitReady
+		// return ctx.Err() before readiness. By this point the controller's
+		// Attach has already put the parent terminal into raw mode (the IO
+		// copiers are relaying the prompt), so returning straight away strands
+		// the parent shell in raw mode — the failure #392 fixes for the
+		// steady-state path, surfacing here on an earlier window. Drive the
+		// same WaitClose/isCleanSessionEnd teardown the ctx.Done arm below
+		// uses: the controller's Run goroutine runs Close once Attach unblocks
+		// on the cancelled ctx, and Close restores the parent terminal before
+		// closing the channel WaitClose blocks on. A handled signal
+		// (context.Canceled) then exits 0 like the steady-state arm; a genuine
+		// readiness failure still surfaces wrapped in ErrWaitOnReady. See #419.
+		logger.DebugContext(ctx, "controller not ready, draining teardown", "error", err)
+		if errC := ctrl.WaitClose(); errC != nil {
+			err = fmt.Errorf("%w: %w", errdefs.ErrWaitOnClose, errC)
+		}
+		if !isCleanSessionEnd(err) {
+			return fmt.Errorf("%w: %w", errdefs.ErrWaitOnReady, err)
+		}
+		return nil
 	}
 
 	logger.DebugContext(ctx, "controller ready, entering client event loop")
--- a/cmd/sbsh/sbsh_test.go
+++ b/cmd/sbsh/sbsh_test.go
@@ -168,6 +168,80 @@ func TestRunTerminal_ErrWaitOnReady(t *testing.T) {
 	}
 }
 
+// TestRunTerminal_WaitReadyContextCanceledDrivesTeardown covers issue #419:
+// when a signal cancels ctx during the attach window, WaitReady returns
+// context.Canceled before readiness. runClient must drive the WaitClose
+// teardown (which restores the parent terminal) and exit 0 rather than
+// returning the spurious ErrWaitOnReady that left the terminal in raw mode.
+func TestRunTerminal_WaitReadyContextCanceledDrivesTeardown(t *testing.T) {
+	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelDebug}))
+
+	waitCloseCalled := make(chan struct{}, 1)
+	newClientController := func(_ context.Context) api.ClientController {
+		return &client.ControllerTest{
+			RunFunc: func(_ *api.ClientDoc) error {
+				return nil
+			},
+			WaitReadyFunc: func() error {
+				// The real Controller.WaitReady only returns non-nil via its
+				// ctx.Done arm, i.e. context.Canceled on a signal landing in
+				// the attach window.
+				return context.Canceled
+			},
+			WaitCloseFunc: func() error {
+				waitCloseCalled <- struct{}{}
+				return nil
+			},
+			StartFunc: func() error {
+				return nil
+			},
+		}
+	}
+
+	ctrl := newClientController(context.Background())
+
+	t.Cleanup(func() {})
+
+	doc := &api.ClientDoc{
+		APIVersion: api.APIVersionV1Beta1,
+		Kind:       api.KindClient,
+		Metadata: api.ClientMetadata{
+			Name:        naming.RandomName(),
+			Labels:      make(map[string]string),
+			Annotations: make(map[string]string),
+		},
+		Spec: api.ClientSpec{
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #420 Review — `fix: restore parent terminal when signal cancels ctx during attach window`**

LGTM — the `WaitReady`-error path now drives the same `WaitClose`/`isCleanSessionEnd` teardown the adjacent `ctx.Done` arm uses (`cmd/sbsh/sbsh.go:518-525`), and `Controller.Close` restores the parent terminal before closing the channel `WaitClose` blocks on (`internal/client/controller.go:432-451`), so the raw-mode restore is guaranteed on the cancel-during-attach window; a genuine readiness failure still surfaces wrapped in `ErrWaitOnReady`.

PR #421, opened 2026-06-10
fix: reap orphan terminal dirs in prune and warn about them in get

### LINKED ISSUE #391: CLI - terminal directories with missing or corrupt metadata.json are invisible to 'sb get' and never reaped by 'sb prune' (permanent disk leak)
## Summary

A terminal run directory whose `metadata.json` is **missing** (terminal killed before it was written — the #374 window) or **truncated/corrupt** is silently dropped from `sb get` output (only a Warn-level log, hidden without `--verbose`) **and** is never removed by `sb prune terminals`. Such directories accumulate under `~/.sbsh/run/terminals/` permanently with no CLI-visible way to clean them.

This is distinct from #258 (which ensured one corrupt file doesn't *mask* other terminals from discovery) and #374 (which addressed not stranding *in-memory* state): the gap here is that the leftover *directory* can never be enumerated, so `prune` — which only deletes directories `ScanTerminals` returns — can never reap it.

## Repro

```bash
# corrupt case
sbsh terminal --name corrupttest -- /bin/bash -i &
sb stop terminal corrupttest
CID=$(sb get terminals -a -o wide | awk '/corrupttest/{print $1}')
META=~/.sbsh/run/terminals/$CID/metadata.json
head -c $(( $(wc -c < "$META") / 2 )) "$META" > "$META.tmp" && mv "$META.tmp" "$META"

sb get terminals -a       # corrupttest is silently absent
sb prune terminals        # does NOT remove it
ls ~/.sbsh/run/terminals/ # directory still present, leaked

# missing-metadata case: a terminals/<id>/ dir containing only a 'log' file
# (process died before writing metadata.json) — same outcome: invisible + un-prunable.
```

## Observed

- `sb get terminals -a` omits the corrupt/missing-metadata terminal with no warning (exit 0).
- `sb prune terminals` reports nothing for it and leaves the directory on disk.
- Both a truncated-JSON directory and a directory with no `metadata.json` at all survive indefinitely.

## Root cause

- `pkg/discovery/scan.go:53,70-91` — `scanMetadataFiles` globs `*/metadata.json` and skips any file that fails `os.ReadFile` or `json.Unmarshal` (logged at Warn, by #258 design). A directory with no `metadata.json` never matches the glob at all.
- `cmd/sb/prune/terminals.go` → `internal/discovery/terminals.go:84-104` — `ScanAndPruneTerminals` only iterates over the terminals `ScanTerminals` returns, and `PruneTerminal` (`terminals.go:106-127`) removes `metadata.Status.TerminalRunPath`. With no parsed metadata, prune has no entry to act on, so the directory is unreachable.

## Suggested fix

Make `prune` directory-driven rather than purely metadata-driven: in addition to pruning scanned-`Exited` terminals, enumerate `terminals/*` directories and reap any whose `metadata.json` is missing or unparseable (optionally gated so a *live* socket in the dir is never removed). Optionally surface a count of skipped/corrupt directories in `sb get` (non-verbose) so operators know cleanup is needed.

## Acceptance criteria

- [ ] `sb prune terminals` removes terminal directories with missing or unparseable `metadata.json` (and never removes one with a live socket).
- [ ] Operators have a non-`--verbose` signal that corrupt/orphan directories exist.
- [ ] A test covers both the truncated-metadata and missing-metadata directory cases through prune.

## References

- `pkg/discovery/scan.go:53,70-91`
- `internal/discovery/terminals.go:84-104,106-127`
- `cmd/sb/prune/terminals.go`
- related: #258 (don't mask other terminals), #374 (don't strand in-memory state)

eminwux:
## Summary
- Terminal run directories whose `metadata.json` is missing (killed in the #374 window) or corrupt are invisible to `ScanTerminals` by design (#258 skips decode failures; the glob never matches a missing file), so the metadata-driven `sb prune terminals` could never reap them — a permanent disk leak under `~/.sbsh/run/terminals/`.
- New `pkg/discovery.OrphanTerminalDirs` enumerates `terminals/*` directories whose `metadata.json` is missing or unparseable. `ScanAndPruneTerminals` now runs a directory-driven pass after the metadata-driven one and removes those orphans.
- **Live-socket guard:** an orphan dir hosting a unix socket that accepts a connection (same 100ms dial-probe pattern as `terminalrunner.OpenSocketCtrl`) is never removed — corrupt metadata doesn't prove the process is gone. Prune prints a `Skipped … (live socket)` line instead.
- **Startup-race guard:** directories modified within the last 30s are shielded. The runner `MkdirAll`s the dir before the first atomic metadata write lands, so reaping inside that window would race a starting terminal. (Design call documented below.)
- `sb get terminals` now prints a one-line warning to **stderr** when orphan dirs exist (`warning: N terminal directories with missing or corrupt metadata.json; run 'sb prune terminals' to clean up`) — visible without `--verbose`, and on stderr so parseable stdout (table/json/yaml) is undisturbed.

## Design calls for review
- Orphan classification is deliberately narrow: a `metadata.json` that exists but is *unreadable* (e.g. EACCES, foreign uid) is **not** an orphan — an unreadable file can't prove the terminal is gone, and the caller likely can't remove the dir anyway. Only ENOENT or read-ok-but-unparseable qualifies, matching the issue's "missing or unparseable".
- The same leak exists for `clients/*` directories (`ScanAndPruneClients` is metadata-driven too); the issue scopes terminals only, so clients parity is left to a follow-up issue (filed separately, referenced on this PR).
- The grace window lives in `pkg/discovery.OrphanTerminalDirs` so prune and the `sb get` warning stay consistent (no warning for a just-created dir either).

## Acceptance criteria
- [x] `sb prune terminals` removes terminal directories with missing or unparseable `metadata.json` (and never removes one with a live socket) — covered by `TestScanAndPruneTerminals_ReapsCorruptMetadataDir`, `_ReapsMissingMetadataDir`, `_SkipsOrphanWithLiveSocket`, plus a real-binary smoke of the issue's repro (both orphan classes reaped, dir empty after prune).
- [x] Operators have a non-`--verbose` signal that corrupt/orphan directories exist — `sb get terminals` stderr warning, covered by `TestReportOrphanTerminalDirs` and the smoke run.
- [x] A test covers both the truncated-metadata and missing-metadata directory cases through prune — `internal/discovery/orphans_test.go` (truncation helper halves the file, matching the repro).

## Test plan
- [x] `make sbsh-sb` (ELF verified)
- [x] `go build ./...` && `go vet ./...`
- [x] `make test` (full suite incl. integration + e2e) — green
- [x] Real-binary smoke: orphan-missing + orphan-corrupt dirs under a scratch `SB_RUN_PATH` → `sb get terminals -a` warns on stderr, `sb prune terminals` reaps both, `terminals/` left empty

Closes #391

### DIFF
--- a/cmd/sb/get/terminals.go
+++ b/cmd/sb/get/terminals.go
@@ -94,10 +94,11 @@ func listTerminals(cmd *cobra.Command, _ []string) error {
 		"args", cmd.Flags().Args(),
 	)
 
+	runPath := viper.GetString(config.SB_ROOT_RUN_PATH.ViperKey)
 	err := discovery.ScanAndPrintTerminals(
 		cmd.Context(),
 		logger,
-		viper.GetString(config.SB_ROOT_RUN_PATH.ViperKey),
+		runPath,
 		os.Stdout,
 		viper.GetBool(listAllInput),
 		format,
@@ -106,6 +107,12 @@ func listTerminals(cmd *cobra.Command, _ []string) error {
 		logger.Debug("error scanning and printing terminals", "error", err)
 		fmt.Fprintln(os.Stderr, "Could not scan terminals")
 		return err
+	}
+	// Orphan dirs (missing/corrupt metadata.json) are invisible to the scan
+	// above; warn on stderr so the leak is observable without --verbose and
+	// without disturbing parseable stdout output. See #391.
+	if _, errOrphans := discovery.ReportOrphanTerminalDirs(cmd.Context(), logger, runPath, os.Stderr); errOrphans != nil {
+		logger.Debug("error reporting orphan terminal directories", "error", errOrphans)
 	}
 	logger.Debug("terminals list completed successfully")
 	return nil
--- a/internal/discovery/orphans.go
+++ b/internal/discovery/orphans.go
@@ -1,0 +1,126 @@
+// Copyright 2025 Emiliano Spinella (eminwux)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package discovery
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/fs"
+	"log/slog"
+	"net"
+	"os"
+	"path/filepath"
+	"time"
+
+	pkgdiscovery "github.com/eminwux/sbsh/pkg/discovery"
+)
+
+// liveSocketProbeTimeout bounds the dial used to decide whether an orphan
+// terminal directory still hosts a live control socket. Mirrors the probe
+// in terminalrunner.OpenSocketCtrl: a successful dial means a live peer is
+// serving the path; ENOENT/ECONNREFUSED/timeout means nobody is accepting.
+const liveSocketProbeTimeout = 100 * time.Millisecond
+
+// pruneOrphanTerminalDirs removes terminal directories whose metadata.json
+// is missing or unparseable. ScanTerminals can never return such
+// directories, so the metadata-driven prune in ScanAndPruneTerminals never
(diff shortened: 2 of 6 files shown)


eminwux:
**PR #421 Review — `fix: reap orphan terminal dirs in prune and warn about them in get`**

LGTM — directory-driven orphan reaping closes the #391 disk leak with a conservative classifier (ENOENT/unparseable only, EACCES preserved), a 30s startup-race grace window, and a live-socket dial-probe guard that mirrors `OpenSocketCtrl`; the stderr warning keeps stdout parseable, and both orphan classes plus the live-socket and fresh-dir guards are covered by tests.

PR #426, opened 2026-06-11
fix: stop clearing terminal on attach/detach by default, add opt-in clear flags

### LINKED ISSUE #425: Attach/detach screen-clearing UX: don't clear by default, add opt-in clear flags, fix detach cursor jump to row 1
## Context

Two related UX problems introduced/exposed by the recent attach/detach work (#302 repaint-on-attach, #369 parent-terminal mode restore):

**1. Attach always clears the client's terminal.** The attach seed paint (`screenModel.repaint()`, `internal/terminal/terminalrunner/terminal_screen.go:179-180`) unconditionally emits `\x1b[2J` (erase screen) + `\x1b[H` (cursor home) before painting the session grid. This runs for every attach — `sb attach` *and* the bare `sbsh` invocation, since both go through the server-side `initialAttachPaint` → `repaint()` path (`internal/terminal/terminalrunner/connections.go:194-212`). For a brand-new `sbsh` session the screen model is empty, so the user's terminal is wiped for nothing: the paint is just clear + home + show-cursor.

**2. Detach jumps the prompt to row 1 without cleaning the screen.** On detach the client writes the `Detached` banner and then `escRestoreParentTerm` (`internal/client/clientrunner/terminal.go:85-93`), whose first sequence is `\x1b[?1049l` (leave alt screen), emitted unconditionally per #365 on the assumption that "resetting a mode that wasn't set is a no-op on every terminal". That assumption does not hold for `1049l`: xterm documents DECRST 1049 as "Use Normal Screen Buffer and **restore cursor as in DECRC**", and DECRC with no prior save restores the *default* cursor — home. Verified against vt10x (the same parser the project's screen model uses):

```
before ?1049l: cursor row=4 col=8     (shell printed 4 lines, sitting at a prompt)
after  ?1049l: cursor row=0 col=0    (no ?1049h was ever sent)
```

So when the session never entered the alt screen, detach yanks the cursor to row 1 col 1, the screen is *not* cleared, and the parent shell's prompt reprints at the top of the terminal overlaying stale session content — the reported bad UX. (`\x1b[?1049l` is the only cursor-affecting sequence on the detach path; the banner itself is just `\r\n`-framed.)

## Proposal

Defaults change to **never clear the user's terminal** on either attach or detach; clearing becomes opt-in per direction.

1. **Attach, default (no clear):** drop `escClearScreen`/`escCursorHome` from the seed paint for normal-buffer sessions. Paint the session's current screen starting at the client cursor's current line using relative framing (`\r\n` between rows, final cursor placement via relative moves, e.g. CUU/CUF), so prior terminal content scrolls up intact instead of being erased. Alt-screen sessions keep `escEnterAltScreen` + clear *inside* the alt buffer — that never destroys the user's normal-buffer content and absolute positioning is required there. An empty screen model paints nothing.
2. **Attach, opt-in clear:** new `--clear-screen` flag on `sb attach` and on bare `sbsh` (env keys following the existing pattern: `SB_ATTACH_CLEAR_SCREEN`, plus the `sbsh` client equivalent). Plumb like `FullCapture`: CLI flag → `ClientSpec`/`api.AttachRequest` → `initialAttachPaint`/`repaint` (`pkg/api/client.go:48`, `pkg/attach/attach.go:63`, `internal/terminal/terminalrunner/lifecycle.go:499`). Note `pkg/spawn/client.go:170` (`maxAttachArgs`) and the spawn arg builder must learn the new flag.
3. **Detach, default (no jump, no clear):** stop emitting a bare `\x1b[?1049l` when the alt screen may never have been entered. Candidate fixes (implementation's call): prefix with DECSC (`\x1b7`) so the DECRC component restores the just-saved position (xterm/VTE keep per-buffer saved cursors, so this does not clobber the normal-buffer slot when a TUI is exiting the alt screen), or track/query alt-screen state and emit `?1049l` only when it was active. After detach the parent prompt must appear on the line following the `Detached` banner, with prior screen content untouched.
4. **Detach, opt-in clear:** new flag on the attaching invocations (`sb attach` / bare `sbsh`), e.g. `--clear-screen-on-detach`, since detach is triggered by keystroke (`^]^]`) or remote `sb detach`, not by a command that could carry the flag itself. When opted in, the client emits a *full* clear — `\x1b[2J\x1b[H` — after `escRestoreParentTerm` in `restoreParentTerminal()` (`internal/client/clientrunner/terminal.go:107`), not merely a cursor-home. Applies on the user-detach path; the process-exit/`Close()` paths should keep current behavior.

## Acceptance criteria

- [ ] Bare `sbsh` (new session) no longer clears the invoking terminal; the shell prompt of the new session appears below the existing terminal content.
- [ ] `sb attach` to a running session no longer clears by default; the session's current screen paints from the current cursor line. Alt-screen (TUI) sessions still repaint correctly inside the alt buffer.
- [ ] `sb attach --clear-screen` / `sbsh --clear-screen` (names final at implementation) restore today's clear-and-repaint behavior, with matching env vars and viper keys following `cmd/config/env.go` conventions.
- [ ] Detach (via `^]^]` and via `sb detach`) leaves the cursor on the line after the `Detached` banner — no jump to row 1 — and does not erase screen content, on sessions that never entered the alt screen. TUI sessions (alt screen active) still return to the normal buffer with the pre-attach cursor restored.
- [ ] Opt-in detach clear flag performs a real screen erase (`\x1b[2J\x1b[H`), verified as bytes written to client stdout, not just a cursor-home.
- [ ] Unit coverage: repaint output with/without clear (normal and alt-screen models), and the detach restore sequence with/without the opt-in (vt10x-based assertions on cursor position fit the existing test style).
- [ ] `--full-capture` interaction defined: full-capture replay is raw history and is unaffected by the new flag (document in flag help).

## Out of scope

- Capture file formats, `sb read`, and the Screenshot RPC — untouched.
- The rest of `escRestoreParentTerm` (cursor visibility, mouse modes, bracketed paste) — still emitted unconditionally; only the `?1049l` cursor side effect changes.

## References

- `internal/terminal/terminalrunner/terminal_screen.go:153-193` — `repaint()`, unconditional `\x1b[2J\x1b[H`
- `internal/terminal/terminalrunner/connections.go:194-212` — `initialAttachPaint`, seed path shared by `sb attach` and bare `sbsh`
- `internal/client/clientrunner/terminal.go:85-93` — `escRestoreParentTerm`, unconditional `?1049l`
- `internal/client/clientrunner/lifecycle.go:247` — detach banner + restore call
- #302 (repaint-on-attach), #365/#369 (unconditional mode reset rationale that `1049l` violates), #383 (banner ordering)
- xterm ctlseqs: DECRST 1049 — "Use Normal Screen Buffer and restore cursor as in DECRC"; reproduced with vt10x (cursor 4,8 → 0,0 with no prior `?1049h`)

eminwux:
## Summary
- **Attach default (no clear):** `repaint()` no longer emits `\x1b[2J\x1b[H` for normal-buffer sessions. It paints the session screen relatively from the client cursor's current line (`\r\n` row framing, per-row erase-to-EOL, final cursor via LF/CUU + CR/CUF), so prior terminal content scrolls up intact. An empty screen model (bare `sbsh`) paints no rows at all. Alt-screen sessions keep `?1049h` + clear + absolute positioning — clearing inside the alt buffer never destroys normal-buffer content, and TUIs need absolute coordinates.
- **Attach opt-in clear:** new `--clear-screen` flag on `sb attach` (env `SB_ATTACH_CLEAR_SCREEN`) and bare `sbsh` (env `SBSH_CLIENT_CLEAR_SCREEN`) restores the legacy clear-and-repaint. Plumbed like `FullCapture`: flag → `ClientSpec.ClearScreen` → `api.AttachRequest.ClearScreen` → `CreateNewClient`/`ioClient` → `initialAttachPaint`/`repaint`. `pkg/builder` gains `WithClientClearScreen`/`WithClientClearScreenOnDetach`; `pkg/spawn`'s arg builder emits both flags and `maxAttachArgs` grew 9→11.
- **Detach default (no row-1 jump):** `escRestoreParentTerm` now prefixes `\x1b[?1049l` with DECSC (`\x1b7`). xterm documents DECRST 1049 as "restore cursor as in DECRC", and DECRC with no prior save restores the *home* cursor — the reported jump on sessions that never entered the alt screen. The DECSC pins the restore to the just-written position (the line after the `Detached` banner). xterm/VTE keep per-buffer saved cursors, so the alt-exit path still restores the pre-attach normal-buffer cursor.
- **Detach opt-in clear:** new `--clear-screen-on-detach` flag on both attaching invocations (`SB_ATTACH_CLEAR_SCREEN_ON_DETACH` / `SBSH_CLIENT_CLEAR_SCREEN_ON_DETACH`). When set, the client writes a real `\x1b[2J\x1b[H` after `escRestoreParentTerm` in `restoreParentTerminal()`. Both user-detach triggers (`^]^]` and `sb detach`) funnel through `clientrunner.Exec.Detach()`, which is the only call site passing the opt-in; the `Close()`/process-exit paths always pass false.

## Implementation calls (reviewer may push back)
- **Opt-in clear also applies on the Detach-RPC-failure branch** — the user committed to detaching either way; the banner stays suppressed per #383 but the clear runs.
- **The opt-in detach clear erases the `Detached` banner** — inherent to a full-screen erase issued after the banner write, and what the issue's spec orders (`clear after escRestoreParentTerm`). The post-clear screen is an empty prompt at home, which is the opted-into UX.
- **DECSC chosen over alt-screen state tracking** (the issue offered either): tracking would require scanning the raw socket→stdout passthrough for `1049h/l` in the copier hot path. The DECSC prefix is a constant change and is correct under xterm/VTE per-buffer saved-cursor semantics. One vt10x caveat documented in the tests: vt10x keeps a *single* saved-cursor slot and swaps buffers on a normal-buffer `?1049l` (`!set || !alt` vs st's XOR), so the TUI-case "pre-attach cursor restored" assertion can't be expressed under vt10x — the alt-screen test asserts buffer return + content integrity instead, and the comment on `escRestoreParentTerm` records the reasoning.
- **DECSC prefix applies to every teardown path** (process exit, ctx cancel), not just detach — the same bare-`?1049l` cursor jump existed there, and the constant is shared.

## Test plan
- [x] `make sbsh-sb` + ELF magic verified
- [x] pre-commit (gofmt, go-mod-tidy, golangci-lint) on changed files — clean
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — full unit suite
- [x] library-consumer example builds + vets
- [x] `make e2e` — full suite green (incl. the attach-repaint prompt detection in `e2e_lifecycle_test.go`, which still matches under the relative paint)
- [x] New unit coverage: repaint with/without clear on normal + alt-screen models, relative paint fed into a simulated client screen (content preserved, cursor lands relative), detach restore sequence with/without the opt-in (vt10x cursor assertions: stays on the line after the banner, not row 1; opt-in clear verified as `\x1b[2J\x1b[H` bytes on client stdout), RPC-failure clear leg, alt-screen detach returns to normal buffer with content intact

## AC mapping
- Bare `sbsh` no longer clears: empty-model paint emits no `2J`/`H` (TestRepaintEmptyScreen)
- `sb attach` default no clear, paints from current line: TestRepaintRelativePreservesClientContent / TestInitialAttachPaintRepaintsScreen; alt-screen repaint unchanged (TestRepaintAltScreen)
- `--clear-screen` + env vars per `cmd/config/env.go` conventions: TestRepaint*ClearScreen, TestInitialAttachPaintClearScreen
- Detach leaves cursor after banner, no erase: TestDetach_CursorStaysAfterBanner; TUI return to normal buffer: TestDetach_AltScreenSessionReturnsToNormalBuffer (pre-attach cursor restore holds under xterm/VTE per-buffer DECSC slots; not assertable under vt10x's single slot — see test comment)
- Opt-in detach clear is a real erase, verified as stdout bytes: TestDetach_OptInClearScreen
- Unit coverage with vt10x assertions: above
- `--full-capture` interaction: documented in `--clear-screen` flag help ("No effect with --full-capture, whose replay is raw history") and in the `ClientSpec.ClearScreen` godoc

Closes #425

### DIFF
--- a/cmd/config/env.go
+++ b/cmd/config/env.go
@@ -115,6 +115,13 @@ var (
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SB_ATTACH_FULL_CAPTURE = DefineKV("SB_ATTACH_FULL_CAPTURE", "sb/attach/fullCapture")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_ATTACH_CLEAR_SCREEN = DefineKV("SB_ATTACH_CLEAR_SCREEN", "sb/attach/clearScreen")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SB_ATTACH_CLEAR_SCREEN_ON_DETACH = DefineKV(
+		"SB_ATTACH_CLEAR_SCREEN_ON_DETACH",
+		"sb/attach/clearScreenOnDetach",
+	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SB_DETACH_NAME = DefineKV("SB_DETACH_NAME", "sb/detach/name")
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
 	SB_DETACH_ID = DefineKV("SB_DETACH_ID", "sb/detach/id")
@@ -196,6 +203,13 @@ var (
 	SBSH_CLIENT_DISABLE_DETACH_KEYSTROKE = DefineKV(
 		"SBSH_CLIENT_DISABLE_DETACH_KEYSTROKE",
 		"sbsh.client.disableDetachKeystroke",
+	)
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_CLIENT_CLEAR_SCREEN = DefineKV("SBSH_CLIENT_CLEAR_SCREEN", "sbsh.client.clearScreen")
+	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
+	SBSH_CLIENT_CLEAR_SCREEN_ON_DETACH = DefineKV(
+		"SBSH_CLIENT_CLEAR_SCREEN_ON_DETACH",
+		"sbsh.client.clearScreenOnDetach",
 	)
 
 	//nolint:revive,gochecknoglobals,staticcheck // ignore linter warning about this variable
--- a/cmd/sb/attach/attach.go
+++ b/cmd/sb/attach/attach.go
@@ -175,6 +175,15 @@ func setupAttachCmdFlags(attachCmd *cobra.Command) {
 	attachCmd.Flags().
 		Bool("full-capture", false, "Replay the entire raw capture buffer on attach instead of repainting the current screen")
 	_ = viper.BindPFlag(config.SB_ATTACH_FULL_CAPTURE.ViperKey, attachCmd.Flags().Lookup("full-capture"))
+	attachCmd.Flags().
+		Bool("clear-screen", false, "Clear the terminal before repainting the session screen on attach (legacy behavior); the default paints below existing content. No effect with --full-capture, whose replay is raw history")
+	_ = viper.BindPFlag(config.SB_ATTACH_CLEAR_SCREEN.ViperKey, attachCmd.Flags().Lookup("clear-screen"))
+	attachCmd.Flags().
+		Bool("clear-screen-on-detach", false, "Erase the screen after detaching (^] twice or sb detach); the default leaves screen content untouched")
+	_ = viper.BindPFlag(
+		config.SB_ATTACH_CLEAR_SCREEN_ON_DETACH.ViperKey,
+		attachCmd.Flags().Lookup("clear-screen-on-detach"),
+	)
 }
 
 func run(
@@ -201,6 +210,8 @@ func run(
 
 	disableDetach := viper.GetBool(config.SB_ATTACH_DISABLE_DETACH_KEYSTROKE.ViperKey)
 	fullCapture := viper.GetBool(config.SB_ATTACH_FULL_CAPTURE.ViperKey)
+	clearScreen := viper.GetBool(config.SB_ATTACH_CLEAR_SCREEN.ViperKey)
+	clearScreenOnDetach := viper.GetBool(config.SB_ATTACH_CLEAR_SCREEN_ON_DETACH.ViperKey)
 
 	logger.DebugContext(
 		ctx,
@@ -211,6 +222,10 @@ func run(
 		disableDetach,
 		"full_capture",
 		fullCapture,
+		"clear_screen",
+		clearScreen,
+		"clear_screen_on_detach",
+		clearScreenOnDetach,
 	)
 
 	runErr := attach.Run(ctx, attach.Options{
@@ -220,6 +235,8 @@ func run(
 		Stderr:                 os.Stderr,
 		DisableDetachKeystroke: disableDetach,
 		FullCapture:            fullCapture,
+		ClearScreen:            clearScreen,
+		ClearScreenOnDetach:    clearScreenOnDetach,
 		Logger:                 logger,
 	})
 	if runErr == nil || errors.Is(runErr, context.Canceled) || errors.Is(runErr, errdefs.ErrContextDone) {
--- a/cmd/sbsh/sbsh.go
+++ b/cmd/sbsh/sbsh.go
@@ -212,6 +212,10 @@ You can also use sbsh with parameters. For example:
 					TerminalSpec:    terminalSpec,
 					DetachKeystroke: !disableDetach, // Invert flag: when disable-detach is true, DetachKeystroke is false
 					ClientMode:      api.RunNewTerminal,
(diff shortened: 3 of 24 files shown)


eminwux:
**PR #426 Review — `fix: stop clearing terminal on attach/detach by default, add opt-in clear flags`**

LGTM — relative attach paint and the DECSC-prefixed detach restore are correctly reasoned (xterm DECRST-1049 / per-buffer-saved-cursor semantics), the opt-in clear flags are threaded through every public façade (`pkg/builder`, `pkg/attach.Options`, `pkg/spawn` arg builder), and the unit + e2e coverage exercises both modes with green CI.

Issue #56, opened 2025-11-10
Release - Generate `SHA256SUMS` for all release assets and upload with the GitHub Release


eminwux:
Add release automation to produce SHA256SUMS for every asset and publish them alongside releases, modifying Makefile and release workflows.

PR #428, opened 2026-06-11
feat: add one-line install script (curl | bash)

### LINKED ISSUE #427: release: add one-line install script (curl | bash)
## Context

Installing sbsh today is a four-step manual incantation, documented in the README §Install:

```bash
curl -L -o sbsh https://github.com/eminwux/sbsh/releases/download/v0.6.0/sbsh-${OS}-${ARCH} && \
chmod +x sbsh && \
sudo install -m 0755 sbsh /usr/local/bin/sbsh && \
sudo ln -f /usr/local/bin/sbsh /usr/local/bin/sb
```

The operator has to hand-set `OS`/`ARCH`, hardcode a release tag (`v0.6.0`), and remember the `sb` hardlink. The sibling project **kukeon** ships a `scripts/install.sh` that collapses its equivalent flow into `curl -fsSL https://kukeon.io/install.sh | bash`. sbsh should offer the same one-liner.

sbsh's installer is **simpler than kukeon's**: sbsh is a self-contained binary with no host prerequisites — there is no `init` command, no daemon subcommand, and no containerd/cgroups/systemd dependency. So the prereq-check, `kuke init`, and systemd-unit machinery in kukeon's script are all **out of scope**. sbsh also targets **linux, darwin, and freebsd** (kukeon is Linux-only).

## Proposal

Add `scripts/install.sh` that runs as `curl -fsSL <url> | bash` and collapses the manual flow: detect OS/arch via `uname`, resolve the latest release tag via the GitHub API (overridable with `SBSH_VERSION`), download + `chmod +x` + `install -m 0755` + `ln -f` the `sb` hardlink, verify checksum when present (skippable via `SBSH_SKIP_CHECKSUM=1`), use `sudo` only when not already root, and print next-steps. Env overrides: `SBSH_VERSION`, `SBSH_REPO`, `SBSH_INSTALL_PREFIX`, `SBSH_SKIP_CHECKSUM`. Add the one-liner to README §Install alongside the existing manual method, and add a CI workflow exercising the install path.

## Acceptance criteria

- [ ] `scripts/install.sh` installs sbsh + the `sb` hardlink via `curl -fsSL <url> | bash`.
- [ ] OS/arch auto-detected from `uname`; unsupported combos exit non-zero.
- [ ] Latest release tag resolved via the GitHub API; `SBSH_VERSION` pins a tag.
- [ ] `sb` is a hardlink (`ln -f`), not a symlink, to preserve `argv[0]` dispatch.
- [ ] Checksum verified when present; `SBSH_SKIP_CHECKSUM=1` bypass, no silent skip.
- [ ] Env overrides honored: `SBSH_VERSION`, `SBSH_REPO`, `SBSH_INSTALL_PREFIX`, `SBSH_SKIP_CHECKSUM`.
- [ ] CI workflow exercises the positive install path.
- [ ] README §Install presents the one-liner and the manual method.

## Non-goals

- No host prerequisite checks (containerd/cgroups) — sbsh has none.
- No `init` step or systemd unit — sbsh has no daemon-bootstrap requirement.

## Notes

- Prior art: kukeon's `scripts/install.sh`.
- Checksum verification relates to #56 (SHA256SUMS); degrade gracefully until it lands.
- Docs-site install pages (getting-started, install-linux, install-macos) are tracked separately as a `documentation` follow-up.

eminwux:
## Summary

- Adds `scripts/install.sh`, a `curl -fsSL <url> | bash` one-liner that collapses the four-step manual install (download → chmod → install → hardlink) from README §Install into a single invocation.
- Auto-detects OS (`uname -s` → linux/darwin/freebsd) and arch (`uname -m` → amd64/arm64), maps to the published `sbsh-<os>-<arch>` asset name (matches `Makefile` OS/ARCHS and `release.yaml`).
- Resolves the latest release tag via the GitHub API (no hardcoded tag), `SBSH_VERSION` pins a specific tag.
- Creates `sb` as a real **hardlink** (`ln -f`), not a symlink, to preserve the `argv[0]` basename dispatch.
- Mirrors kukeon's env-override surface: `SBSH_VERSION`, `SBSH_REPO`, `SBSH_INSTALL_PREFIX` (default `/usr/local/bin`), `SBSH_SKIP_CHECKSUM`.
- Adds `.github/workflows/installer.yaml` exercising the positive install path on ubuntu + macOS hosted runners (default path, hardlink/inode assertion, version-match, and the pinned-`SBSH_VERSION` override path).
- README §Install gains the one-liner above the existing manual steps (folded into a `<details>` block).

## Implementation notes (for reviewer push-back)

- **Checksum (AC item 5, depends on #56):** no release currently publishes a `.sha256` asset (confirmed: `release.yaml` uploads only the six binaries). The AC offers two graceful-degradation options — *warn + continue* or *fail with documented `SBSH_SKIP_CHECKSUM=1` bypass*. I chose **warn + continue** as the lower-blast-radius default: hard-failing would break the one-liner for everyone until #56 lands, defeating the purpose. The warning is loud and printed to stderr — **not a silent skip** — and names the `SBSH_SKIP_CHECKSUM=1` env to acknowledge/silence it. When a `.sha256` asset *is* present the script verifies it and hard-fails on mismatch. Once #56 lands, verification activates automatically with no script change.
- **`sudo` use (AC item: "sudo only when not already root"):** I generalized the literal reading slightly — the script uses `sudo` only when the install prefix is **not writable** by the current user (root and user-writable custom prefixes both skip it). This is a strict superset of "only when not root" (root is always writable) and lets CI / rootless custom-prefix installs run without a sudo prompt. Flagging in case the reviewer wants the literal "non-root ⇒ always sudo" behavior instead.
- **Out of scope (per issue Non-goals):** no prereq checks, no `init` step, no systemd unit — sbsh has none of kukeon's host dependencies.
- **Stable short-URL publish** (e.g. via sbsh.io) is the issue's documented optional follow-up — out of scope here; the README one-liner references the raw GitHub `main` path.

## Test plan

- [x] `pre-commit run --files scripts/install.sh .github/workflows/installer.yaml README.md` — all hooks pass (prettier, check-yaml, etc.)
- [x] `make sbsh-sb` + ELF-magic verify (`7f 45 4c 46`) — produces a real executable
- [x] `go build ./...`, `go vet ./...` — clean (no Go changed)
- [x] `go test -timeout=5m $(go list ./... | grep -v '/e2e$')` — pass
- [x] End-to-end install against real release v0.12.5 into a temp prefix: latest-tag resolution, download, loud missing-checksum warning, no-sudo writable-prefix install, `sb`/`sbsh` share one inode (hardlink), both run `version` OK.
- [x] Override paths: `SBSH_VERSION` pin, `SBSH_SKIP_CHECKSUM=1`, bad-version download failure (exit 1 + clear message), unsupported OS/arch (exit 1 + clear message).

Closes #427

### DIFF
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -74,9 +74,14 @@ jobs:
           # Resolve the latest tag once, then re-install pinned to it via
           # SBSH_VERSION to exercise the pin path. Skip-checksum is set so the
           # absent-checksum-asset case is acknowledged rather than warned.
-          tag="$(curl -fsSL -H "Authorization: Bearer $GITHUB_TOKEN" \
-            https://api.github.com/repos/eminwux/sbsh/releases/latest \
-            | grep -m1 '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')"
+          # Capture the API response in full before piping to `grep -m1`:
+          # piping curl directly into `grep -m1` makes grep close the pipe
+          # after the first match, so curl aborts mid-write with `curl: (23)`
+          # under `pipefail` (the install.sh resolver captures first for the
+          # same reason).
+          resp="$(curl -fsSL -H "Authorization: Bearer $GITHUB_TOKEN" \
+            https://api.github.com/repos/eminwux/sbsh/releases/latest)"
+          tag="$(printf '%s\n' "$resp" | grep -m1 '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')"
           echo "pinning to $tag"
           SBSH_VERSION="$tag" bash scripts/install.sh
           test -x "$SBSH_INSTALL_PREFIX/sbsh"
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -64,7 +64,7 @@ jobs:
           echo "sb version:   $v_sb"
           test "$v_sbsh" = "$v_sb"
 
-      - name: Verify env overrides (pinned version + fork repo defaults)
+      - name: Verify env overrides (pinned SBSH_VERSION + SBSH_SKIP_CHECKSUM)
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SBSH_INSTALL_PREFIX: ${{ github.workspace }}/.sbsh-pinned
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -1,0 +1,83 @@
+name: Installer
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "scripts/install.sh"
+      - ".github/workflows/installer.yaml"
+  pull_request:
+    branches:
+      - main
+    paths:
+      - "scripts/install.sh"
+      - ".github/workflows/installer.yaml"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  install:
+    name: Install (${{ matrix.os }})
+    # Exercise the positive `curl | bash` install path on GitHub-hosted
+    # runners. macOS covers the darwin/arm64 asset; ubuntu covers linux/amd64.
+    # The script resolves the latest published release, so this validates the
+    # real end-to-end flow against actual release assets.
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04, macos-latest]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v4
+
+      - name: Run the installer
+        env:
+          # Authenticated API call avoids the unauthenticated rate limit when
+          # resolving the latest release tag on shared runners.
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # Install to a workspace-local prefix so no sudo prompt is needed and
+          # the runner's system dirs stay untouched.
+          SBSH_INSTALL_PREFIX: ${{ github.workspace }}/.sbsh-bin
+        run: bash scripts/install.sh
+
+      - name: Verify the install
+        env:
+          SBSH_INSTALL_PREFIX: ${{ github.workspace }}/.sbsh-bin
+        run: |
+          set -euo pipefail
+          prefix="$SBSH_INSTALL_PREFIX"
+          # Both names exist and are executable.
+          test -x "$prefix/sbsh"
+          test -x "$prefix/sb"
+          # `sb` is a real hardlink to `sbsh` (same inode), not a symlink — the
+          # argv[0] dispatch depends on it.
+          test ! -L "$prefix/sb"
+          test "$prefix/sbsh" -ef "$prefix/sb"
+          # Both invocations run and report the same version.
+          v_sbsh="$("$prefix/sbsh" version)"
+          v_sb="$("$prefix/sb" version)"
+          echo "sbsh version: $v_sbsh"
+          echo "sb version:   $v_sb"
+          test "$v_sbsh" = "$v_sb"
+
+      - name: Verify env overrides (pinned version + fork repo defaults)
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SBSH_INSTALL_PREFIX: ${{ github.workspace }}/.sbsh-pinned
+          SBSH_SKIP_CHECKSUM: "1"
+        run: |
+          set -euo pipefail
+          # Resolve the latest tag once, then re-install pinned to it via
+          # SBSH_VERSION to exercise the pin path. Skip-checksum is set so the
+          # absent-checksum-asset case is acknowledged rather than warned.
+          tag="$(curl -fsSL -H "Authorization: Bearer $GITHUB_TOKEN" \
+            https://api.github.com/repos/eminwux/sbsh/releases/latest \
+            | grep -m1 '"tag_name"' | sed -E 's/.*"([^"]+)".*/\1/')"
+          echo "pinning to $tag"
+          SBSH_VERSION="$tag" bash scripts/install.sh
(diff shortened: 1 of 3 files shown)


eminwux:
Re-running the failed CI job once — `Install (ubuntu-22.04)` failed in the *Verify env overrides (pinned version + fork repo defaults)* step with `curl: (23) Failure writing output to destination`, which reads as a transient download/write hiccup rather than a script defect. Will revisit when it finishes; if it fails again the PR will be marked `ci-stalled` for human investigation.

eminwux:
**PR #428 Review — `feat: add one-line install script (curl | bash)`**

Clean adaptation of the kukeon installer — hardlink (not symlink) preserves the `argv[0]` dispatch, the `maybe_sudo` writability walk is a sound superset of "non-root ⇒ sudo", checksum degradation matches AC item 5 (loud stderr warning, no silent skip), and the ubuntu+macOS CI assertions (inode/hardlink, version-match) cover the positive path. No secret leaks, no dead code, all commits verified. One portability blocker on the newly-added darwin target:

- **Portability (blocking)** `scripts/install.sh:128`: the empty `auth` array expanded as `"${auth[@]}"` under `set -u` aborts with `auth[@]: unbound variable` on bash <4.4 — i.e. macOS stock `/bin/bash` (3.2.57) — whenever neither `GITHUB_TOKEN` nor `GH_TOKEN` is set, which is the default one-liner path. This is the common case on a supported OS; CI stays green only because both runners export `GITHUB_TOKEN`, so the empty-array path is never exercised. Guard with `${auth[@]+"${auth[@]}"}` or branch on `$token` (separate `curl` calls).

Non-blocking nits (no need to bounce on these alone):

- **Docs** `scripts/install.sh:7`: header example `releases/latest/download/install.sh` points to an asset `release.yaml` never uploads (404) — the README's raw `main` URL is the working one.
- **Naming** `.github/workflows/installer.yaml:67`: step name "fork repo defaults" doesn't exercise `SBSH_REPO`; it tests the pinned-version + skip-checksum path.

LGTM once the bash-3.2 empty-array guard lands — the nits are optional.

eminwux:
Addressed in 1f7c7c9:

- **Portability (blocking)** `scripts/install.sh:128` — guarded the empty-array expansion with `${auth[@]+"${auth[@]}"}` so the no-token one-liner path no longer aborts with `auth[@]: unbound variable` under `set -u` on bash <4.4 (macOS stock `/bin/bash` 3.2.57). Verified the idiom is safe on an empty array under `set -euo pipefail`; `bash -n` syntax-clean. Added an inline comment naming the bash-3.2 reason so the guard isn't "simplified" away later.
- **Docs nit** `scripts/install.sh:7` — replaced the `releases/latest/download/install.sh` header example (404 — `release.yaml` never uploads `install.sh` as an asset) with the working `raw.githubusercontent.com/.../main/scripts/install.sh` URL that README §Install already uses. Dropped the redundant `<raw-url>` placeholder line.
- **Naming nit** `.github/workflows/installer.yaml:67` — renamed the step from "pinned version + fork repo defaults" to "pinned SBSH_VERSION + SBSH_SKIP_CHECKSUM" to match what it actually exercises (no `SBSH_REPO` in that step).

Test plan: `pre-commit run --files` (changed files) pass, `go build`/`go vet`/`go test` (excl e2e) clean — no Go changed.

eminwux:
**PR #428 Review — `feat: add one-line install script (curl | bash)`**

Clean, well-documented installer; asset naming matches `release.yaml` (linux/darwin/freebsd × amd64/arm64), the `sb` hardlink correctly preserves argv[0] dispatch, and the new `installer.yaml` jobs validate the real end-to-end flow on ubuntu + macOS. No secret leaks (the bearer token is only sent to api.github.com, never logged). One blocking item:

- **macOS portability** `scripts/install.sh:180`: `sha256sum` is not present on stock macOS (it ships `shasum`/`openssl`, not GNU coreutils), yet `detect_os` supports `darwin` and the matrix runs `macos-latest`. The macOS CI passed only because no `.sha256` asset exists today, so `verify_checksum` returns early at the warn path and never reaches line 180. Per the PR body, verification "activates automatically with no script change" once #56 lands — at which point this line aborts the install on macOS under `set -e`. Prefer `sha256sum`, fall back to `shasum -a 256 "$bin_path"` (or `openssl dgst -sha256`).

The checksum-degradation default (warn + continue), the generalized `sudo`-when-not-writable behavior, and out-of-scope deferrals are all reasonable and well-justified in the PR notes — no concerns there.

LGTM once the macOS hashing fallback lands — that's the only blocking item.

eminwux:
Addressed in 4d365bc:

- **macOS portability (blocking)** `scripts/install.sh:180` — replaced the unconditional `sha256sum` call with a new `sha256_of()` helper that prefers `sha256sum` (Linux/FreeBSD), falls back to `shasum -a 256`, then `openssl dgst -sha256`. Stock macOS ships `shasum`/`openssl` rather than GNU coreutils' `sha256sum`, so the prior line would have aborted the install on the `darwin` target under `set -e` once a `.sha256` asset lands (#56) and `verify_checksum` reaches the hashing path. Added an inline comment naming the macOS reason so the fallback chain isn't trimmed later.

Verification: confirmed all three tools produce an identical digest for the same input on this host (the `openssl` branch extracts the hex via `$NF` from its `SHA256(<path>)= <hex>` output); `bash -n` syntax-clean; `pre-commit run --files scripts/install.sh` passes; `go build`/`go vet`/`go test` (excl e2e) clean — no Go changed.

eminwux:
**PR #428 Review — `feat: add one-line install script (curl | bash)`**

LGTM — adds a self-contained `curl | bash` installer with correct bash-3.2 token-guard, set-e-safe error propagation, and a real `sb`/`sbsh` hardlink; the `Install (ubuntu-22.04)`/`Install (macos-latest)` CI jobs execute the install end-to-end against live release assets and assert the shared inode + version-match, so the runtime path is verified on both OSes. No leaks (token bearer-only, never logged), no dead code, checksum degrades loudly-not-silently.

PR #433, opened 2026-06-12
feat: publish installer at sbsh.io/install.sh with --check/--help and docs one-liner

eminwux:
## Summary

Completes the hosted one-liner front door for #429. **The issue's premise has partially rotted:** it was filed assuming "sbsh has no `install.sh` anywhere today", but #428 (`2af1bbb`, merged after #429 was filed) already shipped a comprehensive `scripts/install.sh` (OS/arch detection, latest-tag resolution, `SBSH_VERSION`/`SBSH_INSTALL_PREFIX` overrides, gated/skippable/graceful checksum, `sb` hardlink). The hosted-one-liner *goal* of #429 remained unmet, so this PR ships only the genuinely-missing pieces rather than re-implementing #428:

- **`--check` / `--help` flags** on `scripts/install.sh`. `--check` runs prerequisite checks only (OS/arch support, `curl`, a SHA-256 tool, install-prefix writability/sudo) and mutates nothing; `--help` prints usage and the honored env vars. Both added via a new `main()` arg parser; the bare `curl | bash` path is unchanged.
- **Published mirror at `docs/site/install.sh`** — a byte-identical copy of `scripts/install.sh`. Because `mkdocs.yml` sets `docs_dir: docs/site` and `docs/site/CNAME` is `sbsh.io`, this file is served at `https://sbsh.io/install.sh` with no extra infra (same mechanism kukeon uses).
- **Sync guard**: `make install-sh-sync` regenerates the mirror; `make install-sh-check` (CI-friendly) fails if the two have drifted. Documented inline in the Makefile.
- **Docs sweep** of the six AC-named files: `README.md`, `docs/README.md`, `docs/site/getting-started.md`, `docs/site/index.md`, `docs/site/install/install-linux.md`, `docs/site/install/install-macos.md` now feature `curl -fsSL https://sbsh.io/install.sh | bash` as the primary path, with the manual version-pinned block retained as the air-gapped / pinned fallback.

## Non-obvious calls (please push back)

- **`README.md` one-liner was flipped, not gated.** #428 had pointed the README one-liner at `raw.githubusercontent.com/.../scripts/install.sh`. The AC explicitly wants the `sbsh.io/install.sh` mirror, so all six docs (including README) now use `sbsh.io`. The `raw.githubusercontent` URL is preserved as a comment in the script header for the pre-Pages-deploy case.
[...]

PR #436, opened 2026-06-12
fix: wire make install-sh-check into installer CI to catch mirror drift

### LINKED ISSUE #435: followup-nit(ci): wire make install-sh-check into installer.yaml to catch mirror drift
From PR #433. `.github/workflows/installer.yaml:7-15` — the new `make install-sh-check` drift guard (added in #433) is not run in CI, and the workflow's path filter watches only `scripts/install.sh`, not the published mirror `docs/site/install.sh`. The two files are byte-identical today, but a future edit to `scripts/install.sh` without running `make install-sh-sync` would silently serve a stale installer at `sbsh.io/install.sh`. Fix: add an `install-sh-check` step to installer.yaml (the job already triggers on `scripts/install.sh` changes) and add `docs/site/install.sh` to both `paths:` filters so direct edits to the mirror also trip the check.

eminwux:
## Summary
- The `install-sh-check` drift guard added in #433 was never run in CI, and the `installer.yaml` `paths:` filters watched only `scripts/install.sh` — so a direct edit to the published mirror `docs/site/install.sh`, or an edit to `scripts/install.sh` without `make install-sh-sync`, could silently serve a stale installer at `sbsh.io/install.sh`.
- Adds a `make install-sh-check` step to the `install` job (placed right after checkout so it fails fast, before the heavy network install) that `cmp`s the two files.
- Adds `docs/site/install.sh` to both the `push` and `pull_request` `paths:` filters so a direct edit to the mirror also triggers the workflow and trips the check.

## Implementation note
- The AC said "add an `install-sh-check` *step* to installer.yaml (the job already triggers on `scripts/install.sh` changes)" — I followed that literal, lower-blast-radius reading and added it as a step in the existing `install` job rather than introducing a separate job. It does run once per matrix OS (ubuntu + macos), which is mildly redundant for a pure source-tree `cmp`; if the reviewer prefers a dedicated single-runner `drift-check` job, that's a trivial restructure.

## Test plan
- [x] `python3 -c "yaml.safe_load(...)"` — `installer.yaml` parses as valid YAML.
- [x] `make install-sh-check` exits 0 and prints "install.sh mirror in sync" when the two files match.
- [x] Verified drift detection: appended a line to `docs/site/install.sh`, `make install-sh-check` exited 2 with the "out of sync; run 'make install-sh-sync'" message; restored the mirror byte-identical (`cmp -s` clean) afterward.

Closes #435

### DIFF
--- a/.github/workflows/installer.yaml
+++ b/.github/workflows/installer.yaml
@@ -6,12 +6,14 @@ on:
       - main
     paths:
       - "scripts/install.sh"
+      - "docs/site/install.sh"
       - ".github/workflows/installer.yaml"
   pull_request:
     branches:
       - main
     paths:
       - "scripts/install.sh"
+      - "docs/site/install.sh"
       - ".github/workflows/installer.yaml"
   workflow_dispatch:
 
@@ -33,6 +35,13 @@ jobs:
     steps:
       - name: Check out the repo
         uses: actions/checkout@v4
+
+      # Fail fast if the published mirror (docs/site/install.sh) has drifted
+      # from the canonical scripts/install.sh — an edit to either without
+      # running `make install-sh-sync` would otherwise serve a stale installer
+      # at sbsh.io/install.sh.
+      - name: Check installer mirror is in sync
+        run: make install-sh-check
 
       - name: Run the installer
         env:


eminwux:
**PR #436 Review — `fix: wire make install-sh-check into installer CI to catch mirror drift`**

LGTM — adds the existing `make install-sh-check` drift guard as a fail-fast step after checkout and extends both `paths:` filters to `docs/site/install.sh`, so a stale mirror now trips CI; target verified present, YAML valid, mirror in sync on head.

PR #423, opened 2026-06-10
fix: reap orphan client dirs in prune and warn about them in get

eminwux:
## Summary
- `clients/<id>/` directories whose `metadata.json` is missing or unparseable were invisible to `ScanClients` (decode failures skipped by design; the glob never matches a missing file), so the metadata-driven prune in `ScanAndPruneClients` could never reap them — the clients-side twin of the terminals leak fixed in #421.
- Mirrors the #421 mechanics: `pkg/discovery.OrphanClientDirs` enumerates orphan dirs (same ENOENT-or-unparseable-only classification, same 30s startup grace window), `ScanAndPruneClients` runs a directory-driven pass after the metadata-driven one with the live-socket guard (covers `clients/<id>/client.sock`), and `sb get clients` prints a one-line stderr warning when orphan dirs exist.
- Rather than copy-pasting #421's code, the shared mechanics are factored into generic helpers — `orphanDirs[T]` in `pkg/discovery/orphans.go`, `pruneOrphanDirs`/`reportOrphanDirs` in `internal/discovery/orphans.go` — with the existing terminal functions becoming thin wrappers, matching the repo's existing `scanMetadataFiles[T]`/`scanAndPruneMetadata` dedup pattern. Terminal-side behavior, output lines, and warning text are unchanged (existing #421 tests pass unmodified).

## Test plan
- [x] `make sbsh-sb` (ELF magic verified; `file` binary unavailable on this host, checked via `od`)
- [x] `go build ./...` and `go vet ./...`
- [x] `go test ./pkg/discovery/... ./internal/discovery/... ./cmd/sb/get/...`
- [x] Full CI-equivalent suite: `go test $(go list ./... | grep -v '/e2e$')`, `go test -tags=integration ./cmd/sb/get/...`, `E2E_BIN_DIR=$(pwd) go test ./e2e` — exit 0
- [x] New tests: truncated-metadata and missing-metadata client dirs reaped through `ScanAndPruneClients`; live-socket orphan survives; fresh dir shielded by grace window; `ReportOrphanClientDirs` warns/stays silent

## Acceptance criteria
- [x] `sb prune clients` removes client directories with missing or unparseable `metadata.json` (and never removes one with a live socket)
- [x] `sb get clients` surfaces a non-`--verbose` signal when orphan client directories exist
- [x] Tests cover both the truncated-metadata and missing-metadata directory cases through prune

Closes #422

PR #437, opened 2026-06-12
docs(discovery): correct orphan client socket path to clients/<id>/socket

### LINKED ISSUE #424: followup-nit(discovery): client socket docstrings/test name client.sock, real path is clients/<id>/socket
From PR #423 (non-blocking review nit).

The orphan-client-dir live-socket docstrings and test name the client control socket `clients/<id>/client.sock — see pkg/attach`, but that path exists in neither location:
- Standalone clients create `clients/<id>/socket` in `pkg/builder/client.go:146`.
- `pkg/attach` does create `client.sock` (`pkg/attach/attach.go:121`), but inside an ephemeral `/tmp/sbsh-attach-*` dir — never under `clients/`.

Sites to correct `client.sock` → `socket` (and `pkg/attach` → `pkg/builder`):
- `internal/discovery/orphans.go:57` — `pruneOrphanClientDirs` docstring
- `internal/discovery/orphans.go:105` — `hasLiveSocket` docstring
- `internal/discovery/orphans_test.go:214` — test listens on `client.sock`; should use `socket` to match the production filename

Behavior is correct (the guard dials any unix socket directly inside the dir regardless of name); this is doc/test fidelity only, so a future debugger isn't sent to a path/package that doesn't apply.

eminwux:
## Summary
- Followup-nit from PR #423: the orphan-client live-socket docstrings and test named the client control socket `clients/<id>/client.sock — see pkg/attach`, but that path exists in neither location.
- Standalone clients create `clients/<id>/socket` (`pkg/builder/client.go:165`, docstring at `:132`); `pkg/attach` does create `client.sock` (`pkg/attach/attach.go:134`) but inside an ephemeral `/tmp/sbsh-attach-*` dir — never under `clients/`.
- Corrected `client.sock` → `socket` and `pkg/attach` → `pkg/builder` at the three sites the issue named:
  - `internal/discovery/orphans.go:57` — `pruneOrphanClientDirs` docstring
  - `internal/discovery/orphans.go:105` — `hasLiveSocket` docstring
  - `internal/discovery/orphans_test.go:214` — test now listens on `socket` to match the production filename
- Behavior is unchanged: the live-socket guard dials any unix socket directly inside the dir regardless of name, so the test still passes. Doc/test fidelity only.

Note: the issue cited `pkg/builder/client.go:146` and `pkg/attach/attach.go:121`; both line numbers have since shifted (`:165` / `:134`) but the substance is unchanged and verified against current `origin/main`.

## Test plan
- [x] `make sbsh-sb` (canonical build; binary verified ELF via magic bytes — `file` not installed)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test` (full CI suite, exit 0 — includes e2e)
- [x] `git grep -n 'client\.sock' -- internal/discovery/` returns no matches

Closes #424

### DIFF
--- a/internal/discovery/orphans.go
+++ b/internal/discovery/orphans.go
@@ -54,7 +54,7 @@ func pruneOrphanTerminalDirs(ctx context.Context, logger *slog.Logger, runPath s
 // pruneOrphanClientDirs removes client directories whose metadata.json is
 // missing or unparseable — the clients-side counterpart of
 // pruneOrphanTerminalDirs (see #422). The live-socket guard covers the
-// client control socket (clients/<id>/client.sock — see pkg/attach).
+// client control socket (clients/<id>/socket — see pkg/builder).
 func pruneOrphanClientDirs(ctx context.Context, logger *slog.Logger, runPath string, w io.Writer) error {
 	orphans, err := pkgdiscovery.OrphanClientDirs(ctx, logger, runPath)
 	if err != nil {
@@ -102,7 +102,7 @@ func pruneOrphanDirs(ctx context.Context, logger *slog.Logger, w io.Writer, orph
 // a connection. Best-effort: a terminal whose control socket was placed
 // outside its run dir (custom Spec.SocketFile) is not detected, but the
 // default layouts (runPath/terminals/<id>/socket,
-// runPath/clients/<id>/client.sock) are covered.
+// runPath/clients/<id>/socket) are covered.
 func hasLiveSocket(ctx context.Context, dir string) bool {
 	entries, err := os.ReadDir(dir)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #437 Review — `docs(discovery): correct orphan client socket path to clients/<id>/socket`**

LGTM — corrects the orphan-client docstrings/test to the real path (`clients/<id>/socket`, built at `pkg/builder/client.go:165`) and source (`pkg/builder`, not `pkg/attach` whose `client.sock` lives in an ephemeral `/tmp/sbsh-attach-*` dir); `hasLiveSocket` dials sockets by `ModeSocket` regardless of name, so the test rename is fidelity-only and behavior is unchanged.

PR #439, opened 2026-06-13
fix: prefer EvCmdExited over benign PTY-read EvError in Serve cause

### LINKED ISSUE #438: Serve returns benign PTY-master EIO EvError instead of EvCmdExited on clean workload exit (init mode)
## Summary

`server.Server.Serve` returns whichever terminating event reaches `runLoop` first, treating `EvError` and `EvCmdExited` identically. On a normal workload exit in PID-1/init mode, the **PTY-master EIO `EvError` reliably preempts the `EvCmdExited` event**, so `Serve` returns a `read /dev/ptmx: input/output error` cause instead of the `shell process exited: code=N` cause. An out-of-tree supervisor (kukeon's `kuketty`) that classifies the Serve cause to derive the workload's exit code cannot tell a clean exit from a wrapper failure, because the authoritative `EvCmdExited` cause was discarded.

## Versions

- sbsh `v0.13.0`
- Consumer: kukeon `kuketty` via `pkg/terminal/server`

## Mechanism

When the shell child exits, its PTY slave closes and the master `Read` returns EIO. Two events are then published to the single `eventsCh`:

1. `terminalManagerReader` (`internal/terminal/terminalrunner/terminal.go:362-368`) emits
   `Event{Type: EvError, Err: "terminalManagerReader pty read error: …: read /dev/ptmx: input/output error"}`
2. the reaper (`internal/terminal/terminalrunner/lifecycle.go:140,160`) emits
   `Event{Type: EvCmdExited, Err: "shell process exited: code=0"}`

`runLoop` (`pkg/terminal/server/server.go:230-235`) returns `ev.Err` for the **first** of these — and the two branches are merged:

```go
case ev := <-eventsCh:
    if ev.Type == terminalrunner.EvError || ev.Type == terminalrunner.EvCmdExited {
        if ev.Err != nil {
            return ev.Err
        }
        return errors.New("server: terminal exited")
    }
```

The PTY EIO is the immediate kernel side-effect of the child closing the tty, so it wins essentially every time for a clean PID-1/init-mode exit (this reproduces deterministically, it does not flake).

## Observed log (consumer side, clean `exit 0`)

```
ERROR "pty stdout error" err=read /dev/ptmx: input/output error          <- emitted first
INFO  "closing terminal" reason=...pty read error: ...input/output error
INFO  "reaper: tracked child exited" pid=47 exit=0                        <- reaped after
INFO  "terminal close requested" err=shell process exited: code=0
```

Serve returned the EIO `EvError` cause; the `code=0` `EvCmdExited` cause never surfaced to the caller.

## Impact

A supervisor that maps the Serve cause to an exit code mis-reports a clean `exit 0` workload as a wrapper/internal failure. In kukeon this lands the cell in `Error` with exit code 70 instead of `Exited`. Because `EvError` and `EvCmdExited` are indistinguishable in the returned cause, the workload's real exit code (clean or non-zero) is unrecoverable from the public facade.

## Suggested fixes (any one closes the gap)

1. **Suppress / downgrade the PTY-master EIO `EvError` once the child has been reaped** — a master EIO after child exit is benign, not an abnormal error.
2. **Prefer `EvCmdExited` over a benign `EvError`** in `runLoop` — e.g. on `EvError` that is a PTY read EIO, give the `EvCmdExited` event a brief grace window before returning.
3. **Expose a machine-readable exit code** (and a typed terminal-exit cause) on `EvCmdExited` / the Serve return, so a consumer can wait for and trust the authoritative event rather than string-matching the cause.

## Cross-reference

Downstream report: eminwux/kukeon#1282

eminwux:
## Summary

- `server.Server.Serve` returned whichever terminating event reached `runLoop` first, merging `EvError` and `EvCmdExited` into one branch. On a clean workload exit in PID-1/init mode the PTY-master read returns EIO — the immediate kernel side-effect of the child closing its tty — and that `EvError` reliably preempts the reaper's `EvCmdExited`, so `Serve` returned `read /dev/ptmx: input/output error` instead of `shell process exited: code=N`. A supervisor classifying the Serve cause (kukeon's `kuketty`, eminwux/kukeon#1282) could not tell a clean exit from a wrapper failure.
- Implements suggested fix #2: `runLoop` now routes terminating terminal events through a new `terminalCause` helper. When the event is a benign PTY-read `EvError` (`errors.Is(err, terminalrunner.ErrPipeRead)`), it waits up to a short grace window (`ptyExitGrace = 250ms`) for the authoritative `EvCmdExited` and returns that cause instead. If no exit follows (a genuine PTY error with the child still alive), it falls back to the original `EvError` unchanged.

## Why this approach (notes for the reviewer)

- **Server-facade-only, lowest blast radius.** The defect site the issue names is `pkg/terminal/server/server.go`'s merged branch, and kukeon consumes exactly this facade. The internal controller (`internal/terminal/controller.go:244`) already handles `EvError`/`EvCmdExited` in *separate* `switch` cases, so it is not affected and is left untouched. Suggested fix #1 (suppress the EIO in `terminalManagerReader`) and #3 (new typed exit API) both touch the shared runner event semantics other in-tree consumers rely on; #2 confines the change to the affected boundary.
- **Robust to the race rather than racing back.** The EIO fires *before* the reaper marks the child done, so suppressing at emit-time (#1) would itself race. Waiting at the consumer side for the authoritative event is deterministic.
- **Gate on `ErrPipeRead`, not narrowly on `syscall.EIO`.** The issue describes the EIO instance, but the same `terminalManagerReader` path also emits an `EvError` wrapping `ErrPipeRead` on the EOF variant of a tty-close. Gating on the `ErrPipeRead` sentinel covers both and is safe: when no `EvCmdExited` actually follows (a real PTY error, child still alive), the grace window simply expires and the original `EvError` is returned — so there are no false positives, only a bounded delay on the genuine-error path. Reviewer can push back if the narrower EIO-only gate is preferred.

## Test plan

- [x] `make test` (full CI suite incl. e2e) — pass (exit 0)
- [x] `go build ./...` — pass
- [x] `go vet ./...` — pass
- [x] `make sbsh-sb` — produces ELF executable (verified `7f 45 4c 46` magic + exec bit)
- [x] New unit tests in `adapter_internal_test.go`:
  - `TestServer_runLoop_PTYReadErrorPrefersCmdExited` — benign EIO `EvError` followed by `EvCmdExited` returns the exit cause.
  - `TestServer_runLoop_PTYReadErrorWithoutExitReturnsError` — a lone PTY read error falls back to that error after the grace window.

Closes #438

### DIFF
--- a/pkg/terminal/server/adapter_internal_test.go
+++ b/pkg/terminal/server/adapter_internal_test.go
@@ -23,7 +23,9 @@ package server
 import (
 	"context"
 	"errors"
+	"fmt"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
@@ -290,6 +292,39 @@ func TestServer_runLoop_TerminalEvents(t *testing.T) {
 	ev2 <- terminalrunner.Event{Type: terminalrunner.EvCmdExited}
 	if err := srv.runLoop(context.Background(), ev2, make(chan error)); err == nil {
 		t.Fatal("runLoop EvCmdExited: expected non-nil exit error")
+	}
+}
+
+func TestServer_runLoop_PTYReadErrorPrefersCmdExited(t *testing.T) {
+	srv := newAdapterServer(t)
+
+	// A benign PTY-master read error (the EIO the child's tty-close triggers)
+	// reliably preempts the authoritative EvCmdExited on the shared channel.
+	// runLoop must return the EvCmdExited cause, not the benign read error.
+	eioErr := fmt.Errorf("terminalManagerReader pty read error: %w: %w",
+		terminalrunner.ErrPipeRead, syscall.EIO)
+	exitErr := errors.New("shell process exited: code=0")
+
+	ev := make(chan terminalrunner.Event, 2)
+	ev <- terminalrunner.Event{Type: terminalrunner.EvError, Err: eioErr}
+	ev <- terminalrunner.Event{Type: terminalrunner.EvCmdExited, Err: exitErr}
+	if err := srv.runLoop(context.Background(), ev, make(chan error)); !errors.Is(err, exitErr) {
+		t.Fatalf("runLoop pty-EIO then EvCmdExited = %v, want %v", err, exitErr)
+	}
+}
+
+func TestServer_runLoop_PTYReadErrorWithoutExitReturnsError(t *testing.T) {
+	srv := newAdapterServer(t)
+
+	// A genuine PTY read error with no EvCmdExited following (the child is
+	// still alive) must fall back to returning that error once the grace
+	// window elapses — the preference only applies when an exit actually races.
+	eioErr := fmt.Errorf("terminalManagerReader pty read error: %w: %w",
+		terminalrunner.ErrPipeRead, syscall.EIO)
+	ev := make(chan terminalrunner.Event, 1)
+	ev <- terminalrunner.Event{Type: terminalrunner.EvError, Err: eioErr}
+	if err := srv.runLoop(context.Background(), ev, make(chan error)); !errors.Is(err, eioErr) {
+		t.Fatalf("runLoop pty-EIO alone = %v, want %v", err, eioErr)
 	}
 }
 
--- a/pkg/terminal/server/server.go
+++ b/pkg/terminal/server/server.go
@@ -34,6 +34,7 @@ import (
 	"log/slog"
 	"net"
 	"sync"
+	"time"
 
 	"github.com/eminwux/sbsh/internal/terminal/terminalrpc"
 	"github.com/eminwux/sbsh/internal/terminal/terminalrunner"
@@ -212,10 +213,7 @@ func (s *Server) runLoop(
 			return ctx.Err()
 		case ev := <-eventsCh:
 			if ev.Type == terminalrunner.EvError || ev.Type == terminalrunner.EvCmdExited {
-				if ev.Err != nil {
-					return ev.Err
-				}
-				return errors.New("server: terminal exited")
+				return s.terminalCause(ctx, ev, eventsCh)
 			}
 		case err := <-rpcDoneCh:
 			if err != nil {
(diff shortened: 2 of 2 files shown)


eminwux:
**PR #439 Review — `fix: prefer EvCmdExited over benign PTY-read EvError in Serve cause`**

LGTM — `terminalCause` gates the grace-window preference on the `ErrPipeRead` sentinel (matching what `terminalManagerReader`/`Writer` actually wrap), returns the authoritative `EvCmdExited` cause when it races, and falls back to the original error unchanged with `ctx.Done()` honored; non-benign paths return immediately, so existing behavior is preserved. Confined to the server facade, both paths unit-tested with the real error format, CI green.

eminwux:
**PR #439 Review (second reviewer) — `fix: prefer EvCmdExited over benign PTY-read EvError in Serve cause`**

Independent second pass — agree with the existing `ready-to-merge`. Verified locally on the PR head: `go vet`, `go build ./...`, and the two new `runLoop` tests all pass (the fallback test takes the expected 0.25s grace window). Commit is signature-verified, base matches `origin/main`, CI green.

The fix is correctly scoped to the server facade and is deterministic: `terminalrunner` defines only `EvError` and `EvCmdExited`, so the inner-loop `continue` can only ever swallow another benign `EvError` — exactly as the comment states — and `eventCause` preserves the prior return semantics for every non-benign path. `%w` wrapping keeps `errors.Is` working (the tests rely on it).

One non-blocking observation:

- **Edge** `pkg/terminal/server/server.go:264-269`: a *genuine* non-`ErrPipeRead` `EvError` arriving inside the grace window after the benign one is also `continue`d and then masked by the timeout returning the original benign error. Realistically unobservable given the actual event sequence, and not a regression (the old code returned whichever terminating event arrived first regardless) — just flagging the discard for the record.

LGTM — no blocking items.

PR #443, opened 2026-06-20
[codex] Improve installer temp write failures

eminwux:
## Summary

- Detect an unwritable temporary directory before downloading the release binary.
- Preflight available temporary space when `df` and `Content-Length` are available.
- Preserve curl stderr and report curl exit 23 as a local write failure instead of implying the release asset is missing.
- Keep the published `docs/site/install.sh` mirror in sync with the canonical installer.

## Root Cause

The installer downloaded the binary to a temp path before checking whether that filesystem was writable and had enough space. When curl could fetch the release asset but failed to write it locally, the script surfaced a generic download failure and told users to check release assets even though the asset was available.

## Validation

- `bash -n scripts/install.sh`
- `SBSH_INSTALL_PREFIX=/tmp/sbsh-check bash scripts/install.sh --check`
- Simulated curl exit 23 to verify the new local temp-dir guidance.
- Simulated a successful install path and verified the `sbsh`/`sb` hardlink.
- `cmp -s scripts/install.sh docs/site/install.sh`
- `git diff --check`

### DIFF
--- a/docs/site/install.sh
+++ b/docs/site/install.sh
@@ -160,6 +160,33 @@ cleanup_tmpdir() {
     fi
 }
 
+create_install_tmpdir() {
+    INSTALL_TMPDIR="$(mktemp -d -t sbsh-install.XXXXXX)"
+
+    if ! : >"${INSTALL_TMPDIR}/.write-test" 2>/dev/null; then
+        fail "temporary directory is not writable: ${INSTALL_TMPDIR}"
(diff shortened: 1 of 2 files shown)


eminwux:
**PR #443 Review — \`[codex] Improve installer temp write failures\`**

LGTM — the installer now preflights temp write/space failures before download, preserves curl stderr for local write errors, and keeps the published mirror in sync.

